PR #4, opened 2019-01-17
replace much logging with more informative comments

maplebed:
removed a bunch of the ad-hoc logging. added a bunch of logging telling you what's happening inside the system, or at least which events and spans are being accepted and sent along.


metrics/honeycomb.go
```diff
@@ -308,5 +306,5 @@ func (h *HoneycombMetrics) Histogram(name string, obs float64) {
 	histogram.lock.Lock()
 	defer histogram.lock.Unlock()
 	histogram.vals = append(histogram.vals, obs)
-	fmt.Printf("appending %f to %s\n", obs, name)
+	// fmt.Printf("appending %f to %s\n", obs, name)
```
tredman: delete?

PR #5, opened 2019-01-25
add hostname to metrics and logs for better tracking

maplebed:
Mistakenly, it's impossible to look at memory used per server because there's no field identifying the samproxy instance that is sending in the metrics and logs. This change adds one.

(along for the ride are three small log content improvements.)


maplebed:
Travis-CI is having issues.  This change does successfully build and pass tests.

PR #9, opened 2019-01-31
[route] update batch response to be consistent with API

tredman:
Currently we just return a list of integer status codes from our batch endpoint. Our API returns a list of objects like `{"status": 202, "error": ""}`. The effect right now is that our SDK cannot process responses correctly and this complicates debugging of individual events in a batch.

### DIFF
--- a/route/route.go
+++ b/route/route.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"compress/gzip"
 	"encoding/json"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"math"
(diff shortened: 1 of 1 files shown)


route/route.go
```diff
@@ -32,6 +33,11 @@ type Router struct {
 	Metrics      metrics.Metrics       `inject:""`
 }
 
+type BatchResponse struct {
+	Status int    `json:"status"`
+	Error  string `json:"error"`
```
maplebed: could you add `omitempty` - when there are no errors the API omits the `error` key from the returned JSON object.

route/route.go
```diff
@@ -232,11 +238,23 @@ func (r *Router) batch(w http.ResponseWriter, req *http.Request) {
...
+				)
+				r.Logger.Debugf("event from batch %s failed to process event: %s", reqID, err.Error())
 				continue
 			}
-			batchedResponses = append(batchedResponses, http.StatusAccepted)
+			batchedResponses = append(
+				batchedResponses,
+				&BatchResponse{
+					Status: http.StatusAccepted,
+					Error:  "",
```
maplebed: If you like, you can omit this line and Error will have the default value (which is "")

PR #13, opened 2019-02-01
swap out goroutine based trace timeouts for a channel to lighten load

maplebed:
I'm not sure this actually makes much of a difference.  It'd be interesting to try it out and see. Or at least just take a look to see if it's a useful place to think about how to lighten the load in this proxy.


maplebed:
Profiling doesn't seem to indicate that this actually makes any difference, assuming my profiling is right. It's also harder to follow the logic, so I'm disinclined to merge. Happy to write it off as an interesting exercise and close.

PR #14, opened 2019-02-01
[collect] add lock to prevent concurrent map writes in sampler

tredman:


maplebed:
Do I understand it right that if you have two traces, both for the same Dataset, both of which have not yet populated the collector with their appropriate samplers, getting sent at the same time they'll collide in accessing the map here? 👍 looks like a race to me.

tredman:
Not necessarily the same dataset. I think go frowns on writing anything to the map concurrently, regardless if the keys are the same. Similar to how it doesn't like it when you range over the map and then change its contents.

PR #16, opened 2019-02-04
change logging interface to accept structured logs

maplebed:
Swap out the logger for one that is compatible with structured logging, because unstructured logging is the pits.

image 2019-02-04 at 12 00 19 pm

### DIFF
--- a/logger/honeycomb.go
+++ b/logger/honeycomb.go
@@ -28,6 +28,11 @@ type HoneycombLoggerConfig struct {
 	level HoneycombLevel
 }
 
+type HoneycombEntry struct {
+	loggerConfig HoneycombLoggerConfig
+	builder      *libhoney.Builder
+}
+
 type HoneycombLevel int
 
 const (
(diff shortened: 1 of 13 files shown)


logger/honeycomb.go
```diff
@@ -90,6 +95,24 @@ func (h *HoneycombLogger) Stop() error {
...
+		builder:      h.builder.Clone(),
+	}
+	entry.builder.AddField(key, value)
+	return entry
+}
+
+func (h *HoneycombLogger) WithFields(fields map[string]interface{}) Entry {
+	entry := &HoneycombEntry{
+		loggerConfig: h.loggerConfig,
+		builder:      h.builder.Clone(),
```
tredman: once we have an entry, couldn't we use an `Event` rather than cloning the builder with each chain?

tredman: Just thinking about GC on all of these builders. I guess we also have this problem with events too, since we create an event even if it never will be sent based on the log level.

maplebed: done, reuse builder once it's created to an Entry.

PR #19, opened 2019-02-07
Add instrumentation, enable blockOnSend in libhoney

tredman:
Libhoney has a default send queue size of a 1000. After that, we drop things on the floor. This is might be OK when the library is being used inside an instrumented application, but since we're responsible for making sure every event we get gets handled, it is not ok for samproxy. Configuring blockOnSend in the libhoney config to true. This could lead to other issues, like OOMing if there's a large backlog of unsent traces and the goroutines trying to send them pile up. We'll have to see.

I am also adding a response queue processor in the DefaultTransmission implementation. This gives us some maybe useful logging if we get non-200 responses from the API or our peer. Also added some counters to keep track of errors.


tredman:
Oops, I did, but forgot to push :)

PR #20, opened 2019-02-22
Ben.libh client

maplebed:
Adding a libhoney client for metrics and logs. This lets us create separate configurations for each and a separate send queue for each so that a flood on one doesn't affect the primary job of the samproxy; sending spans.


go.mod
```diff
@@ -17,7 +17,7 @@ require (
 	github.com/gorilla/mux v1.6.3-0.20190108142930-08e7f807d38d
 	github.com/hashicorp/golang-lru v0.5.0
 	github.com/honeycombio/dynsampler-go v0.0.0-20171107180038-7f9929d9ca1f
-	github.com/honeycombio/libhoney-go v1.8.1
+	github.com/honeycombio/libhoney-go v1.9.0
```
tredman: can you set this to 1.9.1 now that it's out? 🙏

PR #21, opened 2019-02-22
Separate peer/upstream clients, disable gzip on peer communications

tredman:
We want to separate the transmission queue for different communication channels so that one will not affect the other. This creates separate clients for upstream and peer communications. In addition, we disable gzip compression when forwarding events to peers, to hopefully cut down on CPU usage.

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -32,13 +32,8 @@ github.com/hashicorp/golang-lru v0.5.0 h1:CL2msUPvZTLb5O648aiLNJw3hnBxN2+1Jq8rCO
 github.com/hashicorp/golang-lru v0.5.0/go.mod h1:/m3WP610KZHVQ1SGc6re/UDhFvYD7pJ4Ao+sR/qLZy8=
 github.com/honeycombio/dynsampler-go v0.0.0-20171107180038-7f9929d9ca1f h1:CavWxcRtxHoNGHiz8KvjCOaXTM0rUwh9w9qT+ytfK/E=
 github.com/honeycombio/dynsampler-go v0.0.0-20171107180038-7f9929d9ca1f/go.mod h1:BOeTUPT6fCRH5X/+QqF6Kza3IyLp9uSq/rWgEtI4aZI=
-github.com/honeycombio/libhoney-go v1.8.1 h1:ZNa8QacttKO/628Nm4Bb0H6ur7i9dy7z/BUuA36VJKk=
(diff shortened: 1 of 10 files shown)


go.sum
```diff
@@ -34,6 +34,13 @@ github.com/honeycombio/dynsampler-go v0.0.0-20171107180038-7f9929d9ca1f h1:CavWx
 github.com/honeycombio/dynsampler-go v0.0.0-20171107180038-7f9929d9ca1f/go.mod h1:BOeTUPT6fCRH5X/+QqF6Kza3IyLp9uSq/rWgEtI4aZI=
 github.com/honeycombio/libhoney-go v1.8.1 h1:ZNa8QacttKO/628Nm4Bb0H6ur7i9dy7z/BUuA36VJKk=
 github.com/honeycombio/libhoney-go v1.8.1/go.mod h1:jdLxh51fcBTy6XIpx1efuJmHePs2xUfVkw25lr+hsmg=
+github.com/honeycombio/libhoney-go v1.8.2 h1:stYQJi/21p37Qan2DlSkXDuIrup9zxOul2UyeqMAqX0=
```
maplebed: cleanup - we can remove the older checksums

main.go
```diff
@@ -73,23 +76,64 @@ func main() {
...
+	}
+
+	peerClient, err := libhoney.NewClient(libhoney.ClientConfig{
+		Transmission: &transmission.Honeycomb{
+			MaxBatchSize:         libhoney.DefaultMaxBatchSize,
+			BatchTimeout:         libhoney.DefaultBatchTimeout,
+			MaxConcurrentBatches: libhoney.DefaultMaxConcurrentBatches,
+			PendingWorkCapacity:  libhoney.DefaultPendingWorkCapacity,
+			UserAgentAddition:    userAgentAddition,
+			Transport:            upstreamTransport,
```
maplebed: peerTransport here

main.go
```diff
@@ -73,23 +76,64 @@ func main() {
...
 	}
 
 	var g inject.Graph
 	err = g.Provide(
 		&inject.Object{Value: c},
 		&inject.Object{Value: lgr},
-		&inject.Object{Value: upstreamClient, Name: "upstreamClient"},
-		&inject.Object{Value: &transmit.DefaultTransmission{}},
+		&inject.Object{Value: upstreamTransport, Name: "upstreamTransport"},
+		&inject.Object{Value: peerTransport, Name: "peerTransport"},
```
maplebed: Do we still need to inject the transports if we're injecting the clients?

tredman: Yup for now because the logger and metrics things declare their own client. Also the router declares its own HTTP client, which needs an upstream transport. This could all probably be detangled a bit more though.

PR #39, opened 2019-06-10
Adding debug logging, error handling, timeouts

maplebed:
* Updating the build to correctly set the version
* adding some output indicating the service is starting always
* adding debug logging indicating each subsystem boots successfully
* adding a timeout to the redis scan to fail when trying to scan too large redis instances
* add indication that registration failed when the initial redis connection fails

### DIFF
--- a/cmd/samproxy/main.go
+++ b/cmd/samproxy/main.go
@@ -37,6 +37,7 @@ type Options struct {
 }
 
 func main() {
+	fmt.Println("Starting samproxy")
 	var opts Options
 	flagParser := flag.NewParser(&opts, flag.Default)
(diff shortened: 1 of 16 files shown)


cmd/samproxy/main.go
```diff
@@ -37,6 +37,7 @@ type Options struct {
 }
 
 func main() {
+	fmt.Println("Starting samproxy")
```
martin308: should this be using a logger?

maplebed: There's no logger available there. I was using it for testing to make sure _something_ spun up; I don't think it's necessary anymore.  Removing it.

PR #41, opened 2019-06-14
add endpoint to allow querying the version of the binary via HTTP

maplebed:
It is useful to be able to probe the running binary for version info. This endpoint enables that.

### DIFF
--- a/route/route.go
+++ b/route/route.go
@@ -33,6 +33,10 @@ type Router struct {
 	Collector            collect.Collector     `inject:""`
 	Metrics              metrics.Metrics       `inject:""`
 
+	// version is set on startup so that the router may answer HTTP requests for
+	// the version
+	versionStr string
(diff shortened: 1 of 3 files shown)


route/route.go
```diff
@@ -120,6 +129,10 @@ func (r *Router) panic(w http.ResponseWriter, req *http.Request) {
 	panic("panic? never!")
 }
 
+func (r *Router) version(w http.ResponseWriter, req *http.Request) {
+	w.Write([]byte(fmt.Sprintf(`{"source":"samproxy","version":%s}`, r.versionStr)))
```
martin308: is `versionStr` an actual version string? Or just an id? Should it be wrapped in quotes here so it's returned as a string?

maplebed: yes! it should! thank you.

PR #46, opened 2019-07-01
add support to look for peers on different ports from those at which they listen

maplebed:
In the containerized world, it is not uncommon to have the address at which a service is available be different from the address on which it listens. This change enables manually setting the external port and recognizing that it may be different from the port on which the service is actually listening. For example, in a marathon/mesos cluster, marathon provides this port as an environment variable.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -68,6 +68,8 @@ jobs:
                 root: artifacts
                 paths:
                   - samproxies.tar
+            - store_artifacts:
+                path: artifacts/
 
   publish:
     docker:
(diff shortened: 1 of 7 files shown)


maplebed:
Pausing on this PR until someone has a chance to try it out.  I don't have a marathon/mesos cluster handy to give it a whirl. I suspect changes may be necessary to maintain self-detection because the remotely advertised IP may also be different (not just the port).

MikeGoldsmith:
We’re going to close this for now as it hasn't had any activity for a while. Feel free to re-open, if it’s still relevant.

PR #48, opened 2019-07-17
Recover the error from panic state.

praveengunupati:


maplebed:
Thank you for this PR.

I like the addition of environment to metrics, though I think it would be better suited for a separate PR. 

The `pauseAndSend` function is going away with #65 for a more efficient implementation (the scheduling overhead of maintaining one goroutine per trace is much higher than the cost of locking all traces to iterate through the list), so the recovery here would also go away. 

I'm going to close this PR, but if you submit a new one with just the addition of environment to metrics, I'd happily merge it.

PR #50, opened 2019-11-07
Bump libhoney to v1.12.1

tredman:
Fixes a panic when some events in the batch are nil


martin308:
It might be worth fixing up version `1.12.1` of libhoney first so that it sends the right version string?

PR #55, opened 2020-04-08
Add support for specifying the ClearFrequencySec attribute

subnova-etsy:
At present, the period that samproxy uses to evaluate sample rates over is fixed at 30s (the default of the dynsampler-go implementation). This change allows this to be configured.

A new configuration option is provided for the DynamicSampler: ClearFrequencySec. This defaults to 30s to retain the current behavior but can be configured to support alternative periods.


tredman:
LGTM! thanks for the pr, will cut a new release shortly

PR #57, opened 2020-04-18
close possible avenue for racing and panic'ing/sending traces multiple times

toshok:
It looks to be (theoretically, I haven't verified by testing it) possible for multiple goroutines to end up in this particular switch arm at the same time.

Use a `sync.Once` to make sure that even if that happens they can't both try to close the channel and/or send the trace.


toshok:
There's another `close()` of the cancel channel in `func (i *InMemCollector) sendAfterIdleTimeout`, but that doesn't seem to be called anywhere?  Can I nuke that function?

maplebed:
Nice use of `sync.Once` there. I like it.

PR #59, opened 2020-05-14
Add EMA Dynamic Sampler implementation

tredman:
This has been a part of refinery and honeytail for a while. Time to
bring it into samproxy


config.toml
```diff
@@ -248,6 +248,115 @@ CacheCapacity = 1000
...
+		# map size under control.
+		# Eligible for live reload
+		MaxKeys = 0
+
+		# AgeOutValue indicates the threshold for removing keys from the EMA. The EMA of any key will approach 0
+		# if it is not repeatedly observed, but will never truly reach it, so we have to decide what constitutes "zero".
+		# Keys with averages below this threshold will be removed from the EMA. Default is the same as Weight, as this prevents
+		# a key with the smallest integer value (1) from being aged out immediately. This value should generally be <= Weight,
+		# unless you have very specific reasons to set it higher.
+		# Eligible for live reload
```
maplebed, line 342: This is the only paragraph in this block that is not indented to 80char
```suggestion
		# AgeOutValue indicates the threshold for removing keys from the EMA. The EMA of any key
		# will approach 0 if it is not repeatedly observed, but will never truly reach it, so we have to
		# decide what constitutes "zero". Keys with averages below this threshold will be removed
		# from the EMA. Default is the same as Weight, as this prevents a key with the smallest
		# integer value (1) from being aged out immediately. This value should generally be <= Weight,
		# unless you have very specific reasons to set it higher.
		# Eligible for live reload
```

config.toml
```diff
@@ -248,6 +248,115 @@ CacheCapacity = 1000
...
+
+		# BurstMultiple, if set, is multiplied by the sum of the running average of counts to define
+		# the burst detection threshold. If total counts observed for a given interval exceed the threshold
+		# EMA is updated immediately, rather than waiting on the AdjustmentInterval.
+		# Defaults to 2; negative value disables. With a default of 2, if your traffic suddenly doubles,
+		# burst detection will kick in.
+		# Eligible for live reload
+		BurstMultiple = 2.0
+
+		# BurstDetectionDelay indicates the number of intervals to run after Start is called before burst detection kicks in.
```
maplebed, line 353:
```suggestion
		# BurstDetectionDelay indicates the number of intervals to run after Start is called before
		# burst detection kicks in.
```

PR #61, opened 2020-06-11
sharder: Refuse to reload empty peer list

dcarley:
The peer list can temporarily become empty when using Redis discovery
and the Redis state is lost (e.g. restarting an in-memory instance).
This causes a divide by zero panic when attempting to chose a shard and
results in dropping spans that may have already been correctly routed by
another peer.

I experienced this while running a load test to confirm the behaviour
when Redis becomes unavailable. It continues working fine while Redis is
unavailable, but it resulted in these panics when Redis came back:

    $ kc logs samproxy-96765ff85-44qg5
    time="2020-06-10T14:13:34Z" level=info msg="using identifier from interface" identifier=172.17.0.7 interface=eth0
    time="2020-06-10T14:15:52Z" level=error msg="redis scan encountered an error" err=EOF keys_returned=0 timeoutSec=5s
    time="2020-06-10T14:15:52Z" level=error msg="redis scan encountered an error" err=EOF keys_returned=0 timeoutSec=5s
    time="2020-06-10T14:15:52Z" level=error msg="handler returning error" error.err="caught panic: runtime error: integer divide by zero" error.msg="caught panic" error.status_code=500
    time="2020-06-10T14:15:53Z" level=error msg="handler returning error" error.err="caught panic: runtime error: integer divide by zero" error.msg="caught panic" error.status_code=500
    time="2020-06-10T14:15:53Z" level=error msg="handler returning error" error.err="caught panic: runtime error: integer divide by zero" error.msg="caught panic" error.status_code=500
    $ kc logs samproxy-96765ff85-2zkzw
    time="2020-06-10T14:13:32Z" level=info msg="using identifier from interface" identifier=172.17.0.10 interface=eth0
    time="2020-06-10T14:15:52Z" level=error msg="non-20x response when sending event" api_host="http://172.17.0.7:8081" dataset=beeline-go event_type=unknown status_code=500 target=unknown
    time="2020-06-10T14:15:52Z" level=error msg="non-20x response when sending event" api_host="http://172.17.0.7:8081" dataset=beeline-go event_type=unknown status_code=500 target=unknown
    time="2020-06-10T14:15:52Z" level=error msg="non-20x response when sending event" api_host="http://172.17.0.7:8081" dataset=beeline-go event_type=unknown status_code=500 target=unknown

No spans were dropped after repeating the test with this change. There's
always going to be a possible race condition with some inconsistent
sharding/sampling as the peers re-register and refresh the lists at
different intervals, but this will deal with the worst case.

Returning an error when there are no peers means that it will surface
the error to the logs and re-attempt later. I considered putting the
check beneath the URL parsing but there's no benefit since that will
return an error rather than continuing without a peer.

We originally discussed only keeping the peer list if there was
previously a non-zero peer list but that seems unnecessary to check
since it can't start with a non-zero peer list (`Start` will return an
error if it can't find itself).

### DIFF
--- a/config/mock.go
+++ b/config/mock.go
@@ -8,6 +8,7 @@ import (
 // MockConfig will respond with whatever config it's set to do during
 // initialization
 type MockConfig struct {
+	Callbacks            []func()
 	GetAPIKeysErr        error
 	GetAPIKeysVal        []string
 	GetCollectorTypeErr  error
@@ -46,8 +47,14 @@ type MockConfig struct {
 	PeerManagementType       string
 }
 
-func (m *MockConfig) ReloadConfig()                 {}
-func (m *MockConfig) RegisterReloadCallback(func()) {}
+func (m *MockConfig) ReloadConfig() {
+	for _, callback := range m.Callbacks {
+		callback()
+	}
+}
+func (m *MockConfig) RegisterReloadCallback(callback func()) {
+	m.Callbacks = append(m.Callbacks, callback)
+}
 func (m *MockConfig) GetAPIKeys() ([]string, error) { return m.GetAPIKeysVal, m.GetAPIKeysErr }
 func (m *MockConfig) GetCollectorType() (string, error) {
 	return m.GetCollectorTypeVal, m.GetCollectorTypeErr
--- a/sharder/deterministic.go
+++ b/sharder/deterministic.go
@@ -174,6 +174,10 @@ func (d *DeterministicSharder) loadPeerList() error {
 		return errors.Wrap(err, "failed to get peer list config")
 	}
 
+	if len(peerList) == 0 {
+		return errors.New("refusing to load empty peer list")
+	}
+
 	// turn my peer list into a list of shards
 	newPeers := make([]*DetShard, 0, len(peerList))
 	for _, peer := range peerList {
--- a/sharder/deterministic_test.go
+++ b/sharder/deterministic_test.go
@@ -4,6 +4,7 @@ import (
 	"testing"
 
 	"github.com/honeycombio/samproxy/config"
+	"github.com/honeycombio/samproxy/internal/peer"
 	"github.com/honeycombio/samproxy/logger"
 	"github.com/stretchr/testify/assert"
 )
--- a/sharder/deterministic_test.go
+++ b/sharder/deterministic_test.go
@@ -1,0 +1,42 @@
+package sharder
+
+import (
+	"testing"
+
+	"github.com/honeycombio/samproxy/config"
+	"github.com/honeycombio/samproxy/logger"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestWhichShard(t *testing.T) {
+	const (
+		selfAddr = "127.0.0.1:8081"
+		traceID  = "test"
+	)
+
+	peers := []string{
+		"http://" + selfAddr,
+		"http://2.2.2.2:8081",
+		"http://3.3.3.3:8081",
+	}
+	config := &config.MockConfig{
+		GetPeerListenAddrVal: selfAddr,
+		GetPeersVal:          peers,
+	}
+	sharder := DeterministicSharder{
+		Config: config,
+		Logger: &logger.NullLogger{},
+	}
+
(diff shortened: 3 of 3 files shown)


megan-bee:
Thanks for raising this Dan -- we will take a look at this PR and your issues in the next few days.

maplebed:
Thanks for the bug and PR!

Curious for your take - rather than returning `self`, would you prefer behavior that reuses the previous list when getting zero hosts? In other words:
* when there were previously multiple peers present in redis
* and redis returns zero peers
* the peer list remains unchanged

There's a choice about whether the deterministic sharder or the redis querier would implement this - feels like the deterministic sharder would be a better choice. 

The error that's showing up in your logs is sadly [squashed too early](https://github.com/honeycombio/samproxy/blob/main/config/redimem/redimem.go#L136-L142) to be useful. But if that error were to be returned, it could bubble up to the `watchPeers` function, which [already has logic to skip updating the peer list](https://github.com/honeycombio/samproxy/blob/main/config/redis.go#L227-L231) if Redis returns an error.

The only part of this I haven't dug into is how often those errors appear when everything is actually fine. I'd be curious if you have in your logs evidence that redis is throwing errors from that clause frequently or rarely, to help gauge the impact of changing `redimem.go` to surface them?

maplebed:
Re-reading my comment, I'm realizing it's too early and I've combined two comments.

1. when redis returns an error, we should keep the previous shard list? It looks like in this case, redis was returning an error, it was getting squashed, and so a "populated" peer list with zero entries was returned. This seems like it should be fixed so a redis error skips changing the list.
1. if, under different circumstances, redis actually does return zero entries, what should the behavior be? This would likely be changing code in the sharder rather than redis.

dcarley:
> when redis returns an error, we should keep the previous shard list? It looks like in this case, redis was returning an error, it was getting squashed, and so a "populated" peer list with zero entries was returned. This seems like it should be fixed so a redis error skips changing the list.

This part works fine in my experience. It keeps working happily if Redis disconnects the client or just disappears for a long period of time. It could be worth calling out in the documentation though, because we were sceptical enough to test it ourselves. We've written this in our internal docs:

>> A single Redis node is used for peer discovery between samproxy nodes. The state is cached by samproxy and it will continue working if Redis becomes unavailable for any period of time. However changes to the peer list during this time wouldn't be picked up and would result in dropped spans.
>> 
>> Changes to the peer list (e.g. deployment or resetting Redis) will result in spans being sharded to different nodes and previous sampling decisions will be lost. Traces will be incomplete because spans will be sampled inconsistently to the rest of the trace that they belong to. This inconsistency is temporary and new traces won't be affected once the peer list has stabilised again.

> if, under different circumstances, redis actually does return zero entries, what should the behavior be? This would likely be changing code in the sharder rather than redis.

The particular problem that we ran into, which is kinda unrelated to the Redis scan errors that precede it, is that we're running a single in-memory instance of Redis and restarting it wipes the existing peer records. When that happens there's a race condition between all of the peers refreshing the list and all of the peers re-registering themselves. A similar situation could also happen in disaster recovery of replicated Redis instances.

I think your suggestion of keeping the existing peer list is better, because it will result in less incoming-routed spans getting an inconsistent sampling decision. There's always going to be a possible race condition with some inconsistent sharding/sampling as the peers re-register and refresh the lists are different intervals, but this will deal with the worst case. Want me to update the PR to that approach?

dcarley:
I've rebased with that new approach and included some more detail:

> Returning an error when there are no peers means that it will surface
> the error to the logs and re-attempt later. I considered putting the
> check beneath the URL parsing but there's no benefit since that will
> return an error rather than continuing without a peer.
> 
> We originally discussed only keeping the peer list if there was
> previously a non-zero peer list but that seems unnecessary to check
> since it can't start with a non-zero peer list (Start will return an
> error if it can't find itself).

asdvalenzuela:
@dcarley Apologies for the delay on this. I'd like to merge this in, but we've made some changes that break the tests you wrote. Would you like me to update the tests and push to your branch, or would you prefer to make the changes?

dcarley:
No problem. You can go ahead and update it 👍

PR #65, opened 2020-06-30
ticker-based sends

martin308:
Moves samproxy to process accumulated traces on a single time interval instead of per trace. As per refinery:

> Note this doesn't move the benchmark needle because it's stuck waiting for the ticker, but it uses a whole lot less CPU to do the same job in the same time.

https://github.com/honeycombio/hound/pull/3042

### DIFF
--- a/collect/collect.go
+++ b/collect/collect.go
@@ -17,6 +17,10 @@ import (
 	"github.com/honeycombio/samproxy/types"
 )
 
+const (
+	sendTickerInterval = 100 * time.Millisecond
+)
+
 type Collector interface {
 	// AddSpan adds a span to be collected, buffered, and merged in to a trace.
 	// Once the trace is "complete", it'll be passed off to the sampler then
@@ -56,10 +60,10 @@ type InMemCollector struct {
 
 	sentTraceCache *lru.Cache
 
-	incoming chan *types.Span
-	fromPeer chan *types.Span
-	toSend   chan *sendSignal
-	reload   chan struct{}
+	incoming   chan *types.Span
+	fromPeer   chan *types.Span
+	sendTicker *time.Ticker
+	reload     chan struct{}
 }
 
 // sendSignal is an indicator that it's time to send a trace.
@@ -123,7 +127,6 @@ func (i *InMemCollector) Start() error {
 
 	i.incoming = make(chan *types.Span, capacity*3)
 	i.fromPeer = make(chan *types.Span, capacity*3)
-	i.toSend = make(chan *sendSignal, capacity)
 	i.reload = make(chan struct{}, 1)
 	// spin up one collector because this is a single threaded collector
 	go i.collect()
@@ -196,38 +199,21 @@ func (i *InMemCollector) AddSpanFromPeer(sp *types.Span) {
 // block is the only place we are allowed to modify any running data
 // structures.
 func (i *InMemCollector) collect() {
-	// make sure we send all pending traces when we finish
-	defer func() {
-		for sig := range i.toSend {
-			i.send(sig.trace)
-		}
-	}()
-
 	peerChanSize := cap(i.fromPeer)
+
+	i.sendTicker = time.NewTicker(sendTickerInterval)
+	defer i.sendTicker.Stop()
 
 	for {
 		// record channel lengths
-		i.Metrics.Histogram("collector_tosend_queue", float64(len(i.toSend)))
 		i.Metrics.Histogram("collector_incoming_queue", float64(len(i.incoming)))
 		i.Metrics.Histogram("collector_peer_queue", float64(len(i.fromPeer)))
-
-		// process traces that are ready to send first to make sure we can clear out
-		// any stuck queues that are eligible for clearing
-		select {
-		case sig, ok := <-i.toSend:
-			if ok {
-				i.send(sig.trace)
-			}
-		default:
-		}
 
 		// process peer traffic at 2/1 ratio to incoming traffic. By processing peer
 		// traffic preferentially we avoid the situation where the cluster essentially
 		// deadlocks because peers are waiting to get their events handed off to each
 		// other.
 		select {
-		case sig := <-i.toSend:
-			i.send(sig.trace)
 		case sp, ok := <-i.fromPeer:
 			if !ok {
 				// channel's been closed; we should shut down.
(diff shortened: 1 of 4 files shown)


collect/collect.go
```diff
@@ -371,30 +355,7 @@ func (i *InMemCollector) sendAfterRootDelay(trace *types.Trace) {
...
-}
-
-// pauseAndSend will block for pause time and then send the trace.
-func (i *InMemCollector) pauseAndSend(pause time.Duration, trace *types.Trace) {
-	select {
-	case <-time.After(pause):
-		trace.SendOnce.Do(func() {
-			// TODO fix FinishTime to be the time of the last span + its duration rather
-			// than whenever the timer goes off.
-			trace.FinishTime = time.Now()
```
asdvalenzuela, line 384: `pauseAndSend` added a `FinishTime` to the trace. Should this be happening in the new code as well?

collect/collect.go
```diff
@@ -246,9 +232,9 @@ func (i *InMemCollector) collect() {
 
 		// ok, the peer queue is low enough, let's wait for new events from anywhere
 		select {
-		case sig, ok := <-i.toSend:
-			if ok {
-				i.send(sig.trace)
+		case <-i.sendTicker.C:
```
asdvalenzuela, line 235: Why are we only checking the sendTicker in this select but not the others?

collect/collect.go
```diff
@@ -196,38 +199,21 @@ func (i *InMemCollector) AddSpanFromPeer(sp *types.Span) {
 // block is the only place we are allowed to modify any running data
 // structures.
 func (i *InMemCollector) collect() {
-	// make sure we send all pending traces when we finish
```
asdvalenzuela, line 199: What happens now on shut down? Will pending traces still be sent?

collect/collect.go
martin308: ooh good catch, I'll look into where we can do that now. It's only used for stats AFAICT

collect/collect.go
martin308: Pending traces should still be sent as the cache will be iterated over when we call `Stop` on the collector down [here](https://github.com/honeycombio/samproxy/blob/b5589bf57cd9b14a0e6d8e36326985b7f9796a92/collect/collect.go#L435)

collect/collect.go
martin308: A very good question, I'm just pondering this now 🤔

maplebed: I think this should go up in the previous `select` loop too. The multiple `select`s here with different options are to prioritize getting traffic _out of_ the system above accepting new traffic _into_ the system, so expunging traces that are ready to be sent feels like a good thing to do in the first loop through.

collect/collect.go
```diff
@@ -346,7 +333,7 @@ func isRootSpan(sp *types.Span) bool {
 
 // sendASAP sends a trace as soon as possible, aka no delay.
 func (i *InMemCollector) sendASAP(trace *types.Trace) {
-	go i.pauseAndSend(0, trace)
+	trace.SetDeadline(0)
```
maplebed, line 336: I think this will have to change to actually trigger sending the trace. `sendASAP` is invoked when a new trace is added to the buffer of pending traces but the buffer is close enough to full that an unsent trace is ejected from the ring. Setting the deadline on the trace won't trigger it getting sent since it's already been ejected from the set of traces that will be checked by the ticker.

Open question whether this sending should be in a goroutine or synchronous. Since it's in the path of `processSpan` (aka incoming traffic) I think nonblocking is better? Open to suggestions.

```suggestion
	go func(){
		// TODO fix FinishTime to be the time of the last span + its duration
		trace.FinishTime = time.Now()
		i.send(trace)
	){}
```

types/event.go
```diff
@@ -119,7 +113,16 @@ func (t *Trace) AddSpan(sp *Span) {
...
 func (t *Trace) GetSpans() []*Span {
 	return t.spans
+}
 
+// SetDeadline sets the deadline that this trace should be sent by
+// When we check for traces to send any trace that has exceeded the
+// deadline will be sent
+func (t *Trace) SetDeadline(d time.Duration) {
+	deadline := time.Now().Add(d)
+	if t.SendBy.IsZero() || t.SendBy.After(deadline) {
```
maplebed, line 123: This is an interesting test. I think it means "you're only allowed to set deadlines that are sooner, you can't push out when a trace should be sent".  Did I get that right? Do we need this protection? 

There's a future feature that might make us reconsider - instead of keeping a trace for 60s since the *first* span arrives, allow keeping a trace for 60s since the *most recent* span arrives. This is one approach to allowing very long traces (minutes, hours) while still getting them out the door at a reasonable time. We haven't accepted that work as a feature yet though - just flagging this as a thing that would have to change if we were.

```suggestion
	// only set the deadline if it has not yet been set or is sooner than
	// the existing deadline
	if t.SendBy.IsZero() || t.SendBy.After(deadline) {
```

types/event.go
```diff
@@ -119,7 +113,16 @@ func (t *Trace) AddSpan(sp *Span) {
 // GetSpans returns the list of spans in this trace
 func (t *Trace) GetSpans() []*Span {
 	return t.spans
+}
 
+// SetDeadline sets the deadline that this trace should be sent by
+// When we check for traces to send any trace that has exceeded the
+// deadline will be sent
```
maplebed, line 120:
```suggestion
// SetDeadline sets the deadline after which this trace should be sent. It
// will get sent sometime after the deadline expires (by a ticker check) but
// not before. If a deadline has already been set, this call will only shrink
// the deadline, not expand it. 
```

collect/collect.go
maplebed: Looking at `sendASAP` and realizing it can be triggered during peer span processing makes me believe more strongly that checking the send ticker should be in both places. If we're behind and reading a bunch of peer spans, the probability we'll overflow the trace buffer also increases. If `sendASAP` is triggering a goroutine for every ejected trace, falling behind could trigger a flurry of extra goroutines, increasing CPU, making us fall further behind. Womp womp. Moving this ticker check above will help keep the trace ring clean, reducing the chance of ejected traces, and minimizing calls to `sendASAP`.

If you decide on the non-goroutine based method for handling `sendASAP`, then I'm once again not sure if it's useful to put this in the upper `select`. Wheee!

collect/collect.go
martin308: A lot of the code in `send` modifies traces which may cause some issue. I think this is a good idea to do it async but perhaps if we want to go down this route then having a separate go routine to send traces is a better idea than creating arbitrary go routines like this?

types/event.go
martin308: This is pretty much verbatim from the refinery code https://github.com/honeycombio/hound/blob/ced8efbf733fa4d7ea3b9b72ec677412a0651d51/cmd/dalmatian/app/collect/collect.go#L555. I think if we need to change this in the future then it's no big deal. I'll try to figure out why refinery uses this check though

PR #76, opened 2020-07-21
Use time.duration for config

martin308:
Read `time.Duration` direct from the TOML file. 

🚨 This is a breaking config change 🚨

Also removes unused config value GetSpanSeenDelay


config.toml
```diff
@@ -85,7 +85,7 @@ HoneycombAPI = "https://api.honeycomb.io"
 # trace. This timer is in seconds. Set to 0 for immediate
```
asdvalenzuela, line 85: The comment reads `This timer is in seconds.` Does that need to be updated to state this is a duration? Does setting to 0 for immediate sends still work?

martin308: ah good catch, I'll update the comment. 0 does work as expected

PR #81, opened 2020-07-27
sample: Add key to all spans

dcarley:
Add the key to all spans within the trace rather than just the root.
There is no additional cost to users storing these in Honeycomb and they
will make it easier to see which keys are contributing to the total rate
by number of spans.

We need this to verify what keys are being used for a large number of
events that we suspect belong to a small number of deep traces. We can't
currently figure out how to query that from Honeycomb because only the
root spans appear when grouped by sampler key. It also can't be combined
with the server-side method for counting the number of spans within a
trace:

- https://docs.honeycomb.io/working-with-your-data/tracing/#span-count-per-trace

We've added simple tests that verify the trace/span pointers are updated
correctly. These tests could be extended to verify other dynamic sampler
functionality in the future, but we don't intend to do that ourselves.
The functions for finding a root span are removed because they are no
longer used.

Paired with Claire Alvis.

### DIFF
--- a/sample/dynamic.go
+++ b/sample/dynamic.go
@@ -227,39 +227,11 @@ func (d *DynamicSampler) buildKey(trace *types.Trace) string {
 		key += strconv.FormatInt(int64(len(spans)), 10)
 	}
 
-	// if we should add the key used by the dynsampler to the root span, let's
-	// do so now.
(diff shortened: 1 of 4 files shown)


dcarley:
Thanks for the swift merge!

PR #82, opened 2020-07-27
add debug service and enable with debug flag

asdvalenzuela:
Add `-debug` flag which when used, will run a debug service on port 6060.  The debug service allows you to use pprof to visualize and analyze profiling data.

### DIFF
--- a/cmd/samproxy/main.go
+++ b/cmd/samproxy/main.go
@@ -36,7 +36,7 @@ type Options struct {
 	RulesFile  string `short:"r" long:"rules_config" description:"Path to rules config file" default:"/etc/samproxy/rules.toml"`
 	PeerType   string `short:"p" long:"peer_type" description:"Peer type - should be redis or file" default:"file"`
 	Version    bool   `short:"v" long:"version" description:"Print version number and exit"`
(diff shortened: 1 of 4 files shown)


cmd/samproxy/main.go
```diff
@@ -173,6 +175,13 @@ func main() {
 		fmt.Printf("failed to provide injection graph. error: %+v\n", err)
 		os.Exit(1)
 	}
+
+	if opts.Debug {
+		g.Provide(
```
martin308, line 180: should we handle the error here like do above?

cmd/samproxy/main.go
```diff
@@ -35,6 +36,7 @@ type Options struct {
 	RulesFile  string `short:"r" long:"rules_config" description:"Path to rules config file" default:"/etc/samproxy/rules.toml"`
 	PeerType   string `short:"p" long:"peer_type" description:"Peer type - should be redis or file" default:"file"`
 	Version    bool   `short:"v" long:"version" description:"Print version number and exit"`
+	Debug      bool   `short:"d" long:"debug" description:"If enabled, runs debug service on port 6060"`
```
martin308, line 39: specifying port 606 is not quite true right as it will try to find a port higher than 6060 if it's in use?

PR #87, opened 2020-07-30
Basic integration testing skeleton

martin308:


martin308:
We can re-open this if need be

PR #91, opened 2020-08-03
separate quickstart from complete config, set defaults

asdvalenzuela:


config/file_config.go
```diff
@@ -51,7 +51,34 @@ type samplerConfigType struct {
...
+	defaultPeers = append(defaultPeers, "http://127.0.0.1:8081")
+
+	var defaultAPIKeys []string
+	defaultAPIKeys = append(defaultAPIKeys, "*")
+
+	c.SetDefault("ListenAddr", "0.0.0.0:8080")
+	c.SetDefault("PeerListenAddr", "0.0.0.0:8081")
+	c.SetDefault("APIKeys", defaultAPIKeys)
+	c.SetDefault("Peers", defaultPeers)
+	c.SetDefault("RedisHost", "localhost:6379")
```
martin308, line 65: I think this probably shouldn't have a default as if you are using redis it will be in a cluster and you probably won't be using localhost

config/file_config.go
```diff
@@ -51,7 +51,34 @@ type samplerConfigType struct {
...
+	c.SetDefault("RedisIdentifier", "192.168.1.1")
```
martin308, line 66: I think this shouldn't have a default as you only need to do this if you can't use the default host identification

config/file_config.go
```diff
@@ -51,7 +51,34 @@ type samplerConfigType struct {
...
+	c.SetDefault("IdentifierInterfaceName", "eth0")
```
martin308, line 67: I think this shouldn't have a default as you should only specify this if we can't automatically determine the interface to use

config/file_config.go
```diff
@@ -51,7 +51,34 @@ type samplerConfigType struct {
...
+
+	var defaultPeers []string
+	defaultPeers = append(defaultPeers, "http://127.0.0.1:8081")
+
+	var defaultAPIKeys []string
+	defaultAPIKeys = append(defaultAPIKeys, "*")
+
+	c.SetDefault("ListenAddr", "0.0.0.0:8080")
+	c.SetDefault("PeerListenAddr", "0.0.0.0:8081")
+	c.SetDefault("APIKeys", defaultAPIKeys)
```
martin308, line 63: would this work?

```suggestion
	c.SetDefault("APIKeys", []string{"*"})
```

config/file_config.go
```diff
@@ -51,7 +51,34 @@ type samplerConfigType struct {
...
+	c.SetDefault("Peers", defaultPeers)
```
martin308, line 64: would this work?

```suggestion
	c.SetDefault("Peers",[]string{"http://127.0.0.1:8081"})
```

asdvalenzuela:
> A few suggestions inline. It's a shame there is not a "no metrics" option and then you wouldn't even need the config file at all in a basic setup

I had to add `InMemCollector` to the config file as well since `CacheCapacity` is a nested setting.

martin308:
> > A few suggestions inline. It's a shame there is not a "no metrics" option and then you wouldn't even need the config file at all in a basic setup
> 
> I had to add `InMemCollector` to the config file as well since `CacheCapacity` is a nested setting.

GetOtherConfig strikes again! We really need to get rid of that concept. 

Probably means we should add fetching more things from the config to the tests? How did you catch that cache capacity needed to be in there?

PR #93, opened 2020-08-03
add test for dataset name with whitespace

asdvalenzuela:


config/config_test.go
```diff
@@ -107,7 +107,7 @@ func TestGetSamplerTypes(t *testing.T) {
 		Sampler = "DeterministicSampler"
 		SampleRate = 2
 
-	[SamplerConfig.dataset1]
+	[SamplerConfig.'dataset 1']
```
martin308, line 110: might be worth adding a comment to the `rules.toml` to make sure people know about escaping the dataset name if it contains a space

PR #105, opened 2020-08-19
Configure peer management only via config

martin308:
Shuffle the config.toml file to have a peer management section which specifies file or redis for peer management

Remove the cli flag which switches to redis mode

This relies on #104 to run the tests now


config_complete.toml
```diff
@@ -137,6 +94,57 @@ Logger = "logrus"
...
+#########################
+
+# [PeerManagement]
+# Type = "file"
+# Peers is the list of all servers participating in this proxy cluster. Events
+# will be sharded evenly across all peers based on the Trace ID. Values here
+# should be the base URL used to access the peer, and should include scheme,
+# hostname (or ip address) and port. All servers in the cluster should be in
+# this list, including this host.
+# This list is only used when the 'peer_type' command line flag is set to
```
asdvalenzuela, line 108: This comment and the one below both still reference the `peer-type` command line flag.

martin308: 🤦‍♂️ will update!

PR #112, opened 2020-08-24
Remove unused UnmarshalJSON method

martin308:
Extracted this from my work on msgpack support. This seems to be unused.


asdvalenzuela:
I thought this was called when json.Unmarshal() is called, such as here: https://github.com/honeycombio/samproxy/blob/main/route/route.go#L163

martin308:
Oh you might be right!

PR #113, opened 2020-08-24
allow debug service address to be specified by config

asdvalenzuela:
I tested this locally both by specifying the debug service address in config, and allowing it to choose it's own.

### DIFF
--- a/cmd/samproxy/main.go
+++ b/cmd/samproxy/main.go
@@ -36,7 +36,7 @@ type Options struct {
 	ConfigFile string `short:"c" long:"config" description:"Path to config file" default:"/etc/samproxy/samproxy.toml"`
 	RulesFile  string `short:"r" long:"rules_config" description:"Path to rules config file" default:"/etc/samproxy/rules.toml"`
 	Version    bool   `short:"v" long:"version" description:"Print version number and exit"`
-	Debug      bool   `short:"d" long:"debug" description:"If enabled, runs debug service (runs on the first open port between 0.0.0.0:6060 and 0.0.0.0:6069 by default)"`
+	Debug      bool   `short:"d" long:"debug" description:"If enabled, runs debug service (runs on the first open port between localhost:6060 and :6069 by default)"`
 }
 
 func main() {
--- a/cmd/samproxy/main.go
+++ b/cmd/samproxy/main.go
@@ -174,7 +174,7 @@ func main() {
 	}
 
 	if opts.Debug {
-		err = g.Provide(&inject.Object{Value: &debug.DebugService{}})
+		err = g.Provide(&inject.Object{Value: &debug.DebugService{Config: c}})
 		if err != nil {
 			fmt.Printf("failed to provide injection graph. error: %+v\n", err)
 			os.Exit(1)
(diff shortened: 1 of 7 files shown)


config_complete.toml
```diff
@@ -67,6 +67,11 @@ LoggingLevel = "debug"
 UpstreamBufferSize = 10000
 PeerBufferSize = 10000
 
+# DebugServicePort sets the port the debug service will run on
+# The debug service will only run if the command line flag -d is specified
+# The debug service runs on the first open port between 6060 and 6069 if no port is specified
+# DebugServicePort = "6060"
```
martin308, line 73: I wonder if this should look more like `ListenAddr` rather than just the port?

cmd/samproxy/main.go
```diff
@@ -36,7 +36,7 @@ type Options struct {
 	ConfigFile string `short:"c" long:"config" description:"Path to config file" default:"/etc/samproxy/samproxy.toml"`
 	RulesFile  string `short:"r" long:"rules_config" description:"Path to rules config file" default:"/etc/samproxy/rules.toml"`
 	Version    bool   `short:"v" long:"version" description:"Print version number and exit"`
-	Debug      bool   `short:"d" long:"debug" description:"If enabled, runs debug service (default port 6060)"`
+	Debug      bool   `short:"d" long:"debug" description:"If enabled, runs debug service (runs on the first open port between 0.0.0.0:6060 and 0.0.0.0:6069 by default)"`
```
martin308, line 39: is this true? Looks to be `localhost`, not `0.0.0.0`?

config_complete.toml
```diff
@@ -67,6 +67,11 @@ LoggingLevel = "debug"
 UpstreamBufferSize = 10000
 PeerBufferSize = 10000
 
+# DebugServiceAddr sets the IP and port the debug service will run on
+# The debug service will only run if the command line flag -d is specified
+# The debug service runs on the first open port between 0.0.0.0:6060 and 0.0.0.0:6069 by default
+# DebugServiceAddr = "0.0.0.0:8085"
```
martin308, line 73: Would localhost be better as a suggestion here?
```suggestion
# DebugServiceAddr = "localhost:8085"
```

service/debug/debug_service.go
```diff
@@ -48,28 +50,38 @@ func (s *DebugService) Start() error {
...
 	go func() {
-		// Prefer to listen on addr, but will try to bind to the next 9 ports
-		// in case you have multiple services running on the same host.
-		for i := 0; i < 10; i++ {
-			host, portStr, _ := net.SplitHostPort(addr)
-			port, _ := strconv.Atoi(portStr)
-			port += i
-			addr := net.JoinHostPort(host, fmt.Sprint(port))
-
+		defaultAddr := s.Config.GetDebugServiceAddr()
```
martin308, line 53: is this the default address? It looks like if this is set then it is the specified address and we won't use the default?

asdvalenzuela:
> Does this mean that if you set the `DebugServiceAddr` config field that you would still need to specify the flag to be able to use it? I'm not sure which way would be the best but might be worth thinking about?

Good question. I am leaning towards always needing the flag, because I think it would be odd require it when you use the default address but not when you supply a custom address. Open to other options though!

martin308:
> > Does this mean that if you set the `DebugServiceAddr` config field that you would still need to specify the flag to be able to use it? I'm not sure which way would be the best but might be worth thinking about?
> 
> Good question. I am leaning towards always needing the flag, because I think it would be odd require it when you use the default address but not when you supply a custom address. Open to other options though!

That makes sense to me!

PR #114, opened 2020-08-25
Add msgpack support

martin308:
Adds support for msgpack encoded events/batches. 

Supports the same `Content-Type` headers that our API does.

Although there are unit tests to cover this new unmarshal method, some manual testing as part of the review would be great, as we have minimal coverage on code outside of this new test 😬


asdvalenzuela:
Will resolve https://github.com/honeycombio/samproxy/issues/53#issue-562970518

PR #120, opened 2020-08-27
add dry run mode which marks spans and sends all traces

asdvalenzuela:
Not sure if `samproxy_kept` is the best name for the field. Let me know if you have any suggestions.

### DIFF
--- a/collect/collect_test.go
+++ b/collect/collect_test.go
@@ -176,6 +176,7 @@ func TestDryRunMode(t *testing.T) {
 		GetDefaultSamplerTypeVal: "DeterministicSampler",
 		SendTickerVal:            2 * time.Millisecond,
 		GetOtherConfigVal:        `{"SampleRate":10}`,
+		DryRun:                   true,
 	}
 	samplerFactory := &sample.SamplerFactory{
 		Config: conf,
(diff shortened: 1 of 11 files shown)


collect/collect_test.go
```diff
@@ -164,3 +164,121 @@ func TestAddSpan(t *testing.T) {
...
+	go coll.collect()
+
+	var traceID1 = "abc123"
+	var traceID2 = "def456"
+	var traceID3 = "ghi789"
+	// sampling decisions based on trace ID
+	_, keepTraceID1 := coll.defaultSampler.GetSampleRate(&types.Trace{TraceID: traceID1})
+	_, keepTraceID2 := coll.defaultSampler.GetSampleRate(&types.Trace{TraceID: traceID2})
+	_, keepTraceID3 := coll.defaultSampler.GetSampleRate(&types.Trace{TraceID: traceID3})
+
```
martin308, line 219: maybe we should assert that each of these traces has been sampled as we expect

config/file_config.go
```diff
@@ -64,6 +65,7 @@ func NewConfig(config, rules string) (Config, error) {
 	c.SetDefault("SendTicker", 100*time.Millisecond)
 	c.SetDefault("UpstreamBufferSize", libhoney.DefaultPendingWorkCapacity)
 	c.SetDefault("PeerBufferSize", libhoney.DefaultPendingWorkCapacity)
+	c.SetDefault("DryRun", false)
```
martin308, line 68: maybe we should add a test for this default

PR #126, opened 2020-09-08
add simple collect benchmark

ianwilkes:
I'd like a more comprehensive integration benchmark, but this is a good start.


martin308 (approved):
💖

PR #127, opened 2020-09-08
avoid allocations for unused log entries

ianwilkes:
Uses a new Logger interface to no-op log operations on disabled log levels. It's a break from the logrus model, but I think it's worth it to avoid unexpected perf issues from a misplaced debug line.


ianwilkes:
This reduces but doesn't eliminate allocations - the value argument to "WithField" tends to escape to the heap, which is harder to solve.

martin308 (approved):
A little bit of a more awkward API, but if it improves performance substantially then it seems like a worthwhile tradeoff to make

PR #128, opened 2020-09-08
simplify collect() loop, always drain fromPeer channel first

ianwilkes:
I found it hard to reason about the original 3-select logic; this is simpler and a bit faster. Always drain fromPeer first, then wait on any input.

### DIFF
--- a/collect/collect.go
+++ b/collect/collect.go
@@ -197,62 +197,43 @@ func (i *InMemCollector) collect() {
 	ticker := time.NewTicker(tickerDuration)
 	defer ticker.Stop()
 
-	peerChanSize := cap(i.fromPeer)
-
 	for {
 		// record channel lengths
 		i.Metrics.Histogram("collector_incoming_queue", float64(len(i.incoming)))
 		i.Metrics.Histogram("collector_peer_queue", float64(len(i.fromPeer)))
 
-		select {
-		case <-ticker.C:
-			i.sendTracesInCache(time.Now())
-		default:
-		}
-
-		// process peer traffic at 2/1 ratio to incoming traffic. By processing peer
+		// Always drain peer channel before doing anyhting else. By processing peer
 		// traffic preferentially we avoid the situation where the cluster essentially
 		// deadlocks because peers are waiting to get their events handed off to each
 		// other.
 		select {
-		case <-ticker.C:
-			i.sendTracesInCache(time.Now())
 		case sp, ok := <-i.fromPeer:
 			if !ok {
 				// channel's been closed; we should shut down.
 				return
 			}
 			i.processSpan(sp)
-			// additionally, if the peer channel is more than 80% full, restart this
-			// loop to make sure it stays empty enough. We _really_ want to avoid
-			// blocking peer traffic.
-			if len(i.fromPeer)*100/peerChanSize > 80 {
-				i.Metrics.IncrementCounter("peer_queue_too_large")
-				continue
-			}
 		default:
-		}
-
-		// ok, the peer queue is low enough, let's wait for new events from anywhere
-		select {
-		case <-ticker.C:
-			i.sendTracesInCache(time.Now())
-		case sp, ok := <-i.incoming:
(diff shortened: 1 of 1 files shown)


collect/collect.go
```diff
@@ -194,62 +194,43 @@ func (i *InMemCollector) collect() {
...
 		i.Metrics.Histogram("collector_peer_queue", float64(len(i.fromPeer)))
 
-		select {
-		case <-ticker.C:
-			i.sendTracesInCache(time.Now())
-		default:
-		}
-
-		// process peer traffic at 2/1 ratio to incoming traffic. By processing peer
+		// Always drain peer channel before doing anyhting else. By processing peer
```
martin308, line 202:
```suggestion
		// Always drain peer channel before doing anything else. By processing peer
```

collect/collect.go
```diff
@@ -194,62 +194,43 @@ func (i *InMemCollector) collect() {
...
 		// traffic preferentially we avoid the situation where the cluster essentially
 		// deadlocks because peers are waiting to get their events handed off to each
 		// other.
 		select {
-		case <-ticker.C:
-			i.sendTracesInCache(time.Now())
 		case sp, ok := <-i.fromPeer:
```
martin308, line 207: will this 'drain' the `fromPeer` channel? It looks like it will just preferentially grab an event from this channel first before continuing to do the other steps? I think I tried this approach when moving to the ticker and @maplebed provided some context that without the 80% block that clusters would lock up in practice

martin308 (changes requested):
This looks kind of similar to my earlier attempts to simplify this when moving to the ticker, I think we need to actually drain the `fromPeer` channel when it is greater than 80% full. @maplebed may have some more context here

collect/collect.go
ianwilkes: It does drain `fromPeer`. The second `select` is only hit in the `default` case from the first select, which only happens when `fromPeer` is empty. If there's data in `fromPeer`, it just loops on the first `select` over and over.

martin308: oh right, I see it now. What if the peer queue always has waiting events to process? Do we need some kind of escape hatch here?

ianwilkes: The degenerate mode is that it will just process these peer spans exclusively until shutdown. Since there's no idle time, we don't care about missing the ticker signals; all traces will be sent on cache eviction. We also probably don't want to absorb any local spans if this is what's going on. Missing the reload signal might be a problem though.

I can see this condition happening in short bursts, but not for an extended period unless the cluster is irredeemably fubar.

ianwilkes:
Let's keep this on ice for now.

ianwilkes:
@maplebed it'd be good to hear any specific concerns you have around this.

collect/collect.go
maplebed: This is interesting - I hadn't considered that missing the ticker would be ok. My thought had been that the first priority would be getting finished traces out (to free up room for new content). Second is processing peer traffic (aka work the cluster had already "accepted"), and third incoming spans (accepting new work).
Changing that mental model a bit to something more like muster, where there are both time-based and throughput-based triggers to send existing traces. The ticker is the time-based one, which can be ignored if the throughput-based one has sufficient volume. Here, the throughput based one is going all the way around the ring and ejecting traces that haven't been sent yet.

My only question here is whether keeping the peer channel completely empty is too strict a requirement and will limit total throughput. I suppose both the peer and incoming channels have a capacity, so really short bursts will collect there. If the cluster can't keep the peer channel empty then it's falling behind anyways and shouldn't be accepting additional incoming traffic. 

I'd be curious about the metrics on the `collector_peer_queue` and what it shows about the volume in that queue during normal operation. 

I think I'm game to try this and see how it behaves.

maplebed:
I think this is in the state where it's ready to experiment. I don't think it will get congested in the way it used to so from my perspective this is ready to go.

maplebed (approved):

PR #129, opened 2020-09-09
Reload all sampler settings on change

martin308:
- don't store a default sampler to use, create the default for each dataset that uses the default
- remove reloading from each sampler and push this up into the collector so that we discard all of the samplers when the config changes so that they can be re-created with the new config data


collect/collect_test.go
```diff
@@ -287,3 +289,69 @@ func TestDryRunMode(t *testing.T) {
...
+	}, conf.GetTraceTimeoutVal*2, conf.SendTickerVal)
+
+	conf.ReloadConfig()
+
+	assert.Eventually(t, func() bool {
+		_, ok := coll.datasetSamplers[dataset]
+		return !ok
+	}, conf.GetTraceTimeoutVal*2, conf.SendTickerVal)
+
+	span = &types.Span{
```
asdvalenzuela, line 344: What does adding this new span test that wasn't tested previously?

asdvalenzuela (approved):

collect/collect_test.go
martin308: the new reload functionality, we first check that when a span is added that the map contains the dataset/sampler mapping, then we "reload" the configs that the map gets cleared and then when we send another span that the sampler is created again

asdvalenzuela: But you're sending two spans?

asdvalenzuela: Please ignore! I read it wrong.

PR #130, opened 2020-09-10
add validation for config

asdvalenzuela:
- Add go-validator and use it to validate config fields
- Use specific methods for pulling nested configuration instead of the more general 'GetOtherConfig'
- Validate configurations that are needed if another config value is present, such as HoneycombMetrics


config/file_config.go
```diff
@@ -43,6 +48,40 @@ type samplerConfigType struct {
...
+type HoneycombMetricsConfig struct {
+	MetricsHoneycombAPI      string `validate:"required,url"`
+	MetricsAPIKey            string `validate:"required"`
+	MetricsDataset           string `validate:"required"`
+	MetricsReportingInterval int64  `validate:"required"`
+}
+
+type PeerManagementConfig struct {
+	Type                    string   `validate:"required,oneof= file redis"`
+	Peers                   []string `validate:"required,dive,url"`
```
martin308, line 78: what does `dive` here mean?

martin308 (commented):
Looks like it's heading in the right direction! How do you feel about those config methods just returning the struct/pointer instead of needing to pass in something to get the values?

config/file_config.go
asdvalenzuela: > This tells the validator to dive into a slice, array or map and validate that level of the slice, array or map with the validation tags that follow.

It validates that each string in the array is a URL.

config/mock.go
```diff
@@ -84,6 +112,22 @@ func (m *MockConfig) GetRedisHost() (string, error) { return m.GetRedisHostVal,
 func (m *MockConfig) GetMetricsType() (string, error) {
 	return m.GetMetricsTypeVal, m.GetMetricsTypeErr
 }
+func (m *MockConfig) GetHoneycombMetricsConfig() (HoneycombMetricsConfig, error) {
+	hmConfig := &HoneycombMetricsConfig{}
+	err := json.Unmarshal([]byte(m.GetHoneycombMetricsConfigVal), hmConfig)
```
martin308, line 117: do we still need to unmarshal here? We should just allow setting an instance of the struct to the mock?

martin308 (approved):
Looks good, one comment on making the mock easier to work with would be to be able to specify the structs directly on it instead of needing to marshal them from JSON

PR #131, opened 2020-09-11
add integrated benchmark

ianwilkes:
This is single-node, but captures a lot of the work involved in sending traces. The code's pretty rough and ready but it gets the job done. This shows many more allocations in today's code vs the pre-ticker code, and even more before my log changes. I think that's the source of circle's regression.

Also moves the signal handler out of App.Start into main, so you can Start an app without blocking.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -1,10 +1,6 @@
 package app
 
 import (
-	"os"
-	"os/signal"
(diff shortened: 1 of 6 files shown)


martin308 (approved):
Makes sense to me

PR #132, opened 2020-09-11
remove traces from cache after they are sent

ianwilkes:
Moves some of this logic into the cache so it can be done efficiently. This should lower the memory footprint when the cache capacity is set larger than necessary, and should fix the perf regression associated with repeatedly touching already-sent traces in sendTracesInCache().

The improvement is visible using the benchmark in #131 

```
benchmark                    old ns/op     new ns/op     delta
BenchmarkTraces/batch-8      23715         16990         -28.36%

benchmark                    old allocs     new allocs     delta
BenchmarkTraces/batch-8      286            164            -42.66%
```

### DIFF
--- a/collect/cache/cache_test.go
+++ b/collect/cache/cache_test.go
@@ -2,6 +2,7 @@ package cache
 
 import (
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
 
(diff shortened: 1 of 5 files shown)


collect/cache/cache_test.go
```diff
@@ -53,3 +54,43 @@ func TestBufferOverrun(t *testing.T) {
...
+	expired := c.TakeExpiredTraces(now)
+	assert.Equal(t, 3, len(expired))
+	assert.Equal(t, traces[0], expired[0])
+	assert.Equal(t, traces[1], expired[1])
+	assert.Equal(t, traces[3], expired[2])
+
+	assert.Equal(t, 1, len(c.cache))
+
+	all := c.GetAll()
+	assert.Equal(t, 10, len(all))
```
martin308, line 88: I think this is existing behavior but it seems odd to return a slice that has nil entries in it

collect/cache/cache_test.go
```diff
@@ -53,3 +54,43 @@ func TestBufferOverrun(t *testing.T) {
...
+	c := &DefaultInMemCache{
+		Config:  CacheConfig{10},
+		Metrics: s,
+		Logger:  &logger.NullLogger{},
+	}
+	c.Start()
+
+	now := time.Now()
+	traces := []*types.Trace{
+		&types.Trace{TraceID: "1", SendBy: now.Add(-time.Minute), Sent: true},
```
martin308, line 70: `Sent` is not used right? Is this just to show that?

collect/collect.go
```diff
@@ -434,9 +429,7 @@ func (i *InMemCollector) Stop() error {
 		traces := i.Cache.GetAll()
 		for _, trace := range traces {
 			if trace != nil {
-				if !trace.GetSent() {
```
martin308, line 437: is `GetSent` redundant/can be removed now?

martin308 (approved):
I think this looks good, left a few questions/comments inline

collect/cache/cache_test.go
ianwilkes: Right.

collect/cache/cache_test.go
ianwilkes: True. Fixed it.

collect/collect.go
ianwilkes: It was always redundant. Removed.

PR #133, opened 2020-09-11
fix broken log line, add WithString for efficiently skipped logs

ianwilkes:
Turns out logrus doesn't do what I want when I pass a pointer to a string to WithField. Instead, add WithString() to the Logger interface, which avoids the heap allocation for skipped logs. It's a bit awkward to have both but I think that's a small price to pay for a simple way to avoid vast quantities of garbage.

For background, the problem with `WithField()` is that passing a scalar into the `interface{}` value param causes it to escape to the heap, even if it's ultimately not used. This can add up to a lot of allocations for frequent log lines, and is a huge waste when those logs are disabled. However, a typed `string` value doesn't escape here, so calling `WithString()` is essentially free for disabled log levels.

### DIFF
--- a/collect/collect.go
+++ b/collect/collect.go
@@ -408,9 +408,9 @@ func (i *InMemCollector) send(trace *types.Trace) {
 
 	// ok, we're not dropping this trace; send all the spans
 	if i.Config.GetIsDryRun() && !shouldSend {
-		i.Logger.Info().WithField("trace_id", &trace.TraceID).WithField("dataset", &trace.Dataset).Logf("Trace would have been dropped, but dry run mode is enabled")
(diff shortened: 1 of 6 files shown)


martin308 (approved):
Makes sense. 

I wonder if it also makes sense to just always logrus and then just configure outputs. It looks like logrus is pretty pluggable so we could just have a honeycomb and null output mode so we don't need to reimplement logrus in all of these implmentations

ianwilkes:
> I wonder if it also makes sense to just always logrus and then just configure outputs. It looks like logrus is pretty pluggable so we could just have a honeycomb and null output mode so we don't need to reimplement logrus in all of these implmentations

Unfortunately logrus is what got us the problematic `WithField` interface to begin with. AFAIK there's no way to use the logrus API efficiently.

martin308:
:sad_trombone:

PR #135, opened 2020-09-12
add distributed benchmark

ianwilkes:
Adds a new benchmark with a 5-node cluster. Results are pretty much as you'd expect, but profiling this does show that we spend a lot of time locking metrics. Maybe not a chief concern at this time, but something to keep an eye on.

Also includes a shutdown code for Router.


martin308 (approved):
👾

PR #137, opened 2020-09-15
update debug line comment for clarity

maplebed:
The comment in code had aged a bit while other parts of the code changed. Updated it to indicate the ticker might be triggering this message in addition to racing timers coming down a channel.


ianwilkes:
It doesn't though - traces are now expunged when sent.

maplebed:
Ah, thanks. I misunderstood and thought they were removed when they were done getting sent (still allowing a race) rather than removed in order to be sent. 👍

PR #138, opened 2020-09-15
fix data races

ianwilkes:
Fixes shutdown data race in InMemCollector.Stop, and fixes test races found with `go test -race`


martin308 (approved):
Seems to make sense, worth putting some before/after benchmarks in the PR body to show the impact of the new lock?

ianwilkes:
There's no impact from the new lock; it's only touched once per tick. (Compare with the various metrics, with take a lock each time they're updated.)

PR #139, opened 2020-09-15
implement optional soft allocation ceiling

ianwilkes:
Adds a new MaxAlloc config value. If > 0, InMemCollector will periodically check to see if the process's total allocation is above the limit. If it is, it will send and discard 10% of traces, and reduce the cache capacity to the new size. For this to work well, it should be set to about 75%-90% of available RAM, to leave room for scratch space and various fluctuations.

The intention is that this is a failsafe for if the cache capacity has been set too high; if it triggers regularly, users should lower their starting cache capacity and/or add machines to their samproxy cluster.

Depends on #138

### DIFF
--- a/collect/collect_test.go
+++ b/collect/collect_test.go
@@ -1,6 +1,7 @@
 package collect
 
 import (
+	"runtime"
 	"strconv"
 	"testing"
 	"time"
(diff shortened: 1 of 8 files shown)


collect/collect_test.go
```diff
@@ -371,3 +373,103 @@ func TestSampleConfigReload(t *testing.T) {
...
+			},
+		}
+		coll.AddSpan(span)
+	}
+
+	for len(coll.incoming) > 0 {
+		time.Sleep(conf.SendTickerVal)
+	}
+
+	// Now there should be 1000 traces in the cache.
```
asdvalenzuela, line 433:
```suggestion
	// Now there should be 500 traces in the cache.
```

config_complete.toml
```diff
@@ -166,6 +166,15 @@ Metrics = "honeycomb"
 # temporary odd sampling decisions).
 CacheCapacity = 1000
 
+// MaxAlloc is optional. If set, it must be an integer >= 0. 64-bit values are
```
asdvalenzuela, line 169: I think the comments need to start with `#`

asdvalenzuela (approved):

PR #140, opened 2020-09-15
enable -race in circle test

ianwilkes:
This should work now. I increased the timeout since race tests can be quite a bit slower.


martin308 (approved):
🙌

PR #141, opened 2020-09-17
use more efficient wire serialization

ianwilkes:
Switch to msgpack for internal (peer to peer and upstream) communication.

Switch to jsoniter for batch event unmarshal, which is notably faster than stock json.

The perf win here is pretty clear, and this benchmark doesn't capture the gains from encoding upstream traffic with msgpack, since libhoney's Writer transmission always uses json.

```
benchmark                              old ns/op     new ns/op     delta
BenchmarkDistributedTraces/batch-8     25991         17567         -32.41%

benchmark                              old allocs     new allocs     delta
BenchmarkDistributedTraces/batch-8     296            226            -23.65%

benchmark                              old bytes     new bytes     delta
BenchmarkDistributedTraces/batch-8     23634         19938         -15.64%
```

### DIFF
--- a/app/app_test.go
+++ b/app/app_test.go
@@ -162,6 +162,7 @@ func newStartedApp(
 			},
 			BlockOnSend:            true,
 			DisableGzipCompression: true,
+			EnableMsgpackEncoding:  true,
 			Metrics:                sdPeer,
 		},
 	})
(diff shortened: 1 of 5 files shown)


asdvalenzuela (approved):

martin308 (approved):
Nice!

PR #142, opened 2020-09-17
send excess traces when reducing cache size

ianwilkes:
Also fixes an off-by-one error in the cache building loop, and adds a mutex to mock config to avoid test races.

### DIFF
--- a/collect/collect.go
+++ b/collect/collect.go
@@ -53,7 +53,7 @@ type InMemCollector struct {
 
 	// mutex must be held whenever non-channel internal fields are accessed.
 	// This exists to avoid data races in tests and startup/shutdown.
-	mutex sync.Mutex
+	mutex sync.RWMutex
 
 	cache           cache.Cache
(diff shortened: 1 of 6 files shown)


collect/collect.go
```diff
@@ -120,6 +112,18 @@ func (i *InMemCollector) Start() error {
...
 
+func (i InMemCollector) createStartedCache(capacity int) *cache.DefaultInMemCache {
+	c := &cache.DefaultInMemCache{
+		Config: cache.CacheConfig{
+			CacheCapacity: capacity,
+		},
+		Metrics: i.Metrics,
+		Logger:  i.Logger,
+	}
+	c.Start()
```
martin308, line 123: maybe we could just get rid of the start function? We don't have a matching stop function and it seems to just create some data structures that could be created upfront when we make the initial struct?

martin308 (approved):
Makes sense, one question about just removing the start function from the cache

asdvalenzuela (approved):

PR #143, opened 2020-09-17
add mutex locking to fileConfig

ianwilkes:
The trouble with config reloads is that it means we've got a background goroutine writing new config values - at the same time our code is occasionally reading these values. The real-world risk here is pretty low but in the spirit of cleaning up `go -race` it seems worth getting out of the way.

### DIFF
--- a/config/file_config.go
+++ b/config/file_config.go
@@ -18,7 +18,7 @@ type fileConfig struct {
 	rules     *viper.Viper
 	conf      *configContents
 	callbacks []func()
-	mux       sync.Mutex
+	mux       sync.RWMutex
 }
 
 type configContents struct {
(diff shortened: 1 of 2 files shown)


config/file_config.go
```diff
@@ -290,10 +335,16 @@ func (f *fileConfig) GetHoneycombLoggerConfig() (HoneycombLoggerConfig, error) {
...
 func (f *fileConfig) GetCollectorType() (string, error) {
+	f.mux.RLock()
+	defer f.mux.RUnlock()
+
 	return f.conf.Collector, nil
 }
 
 func (f *fileConfig) GetSamplerConfigForDataset(dataset string) (interface{}, error) {
+	f.mux.Lock()
+	defer f.mux.Unlock()
```
martin308, line 346: should this one be just a `RLock`?

martin308 (approved):
Looks good, just one question around one of the lock/unlock pairs in what is I think a read only method

PR #144, opened 2020-09-18
validate sampling rules configuration

asdvalenzuela:
Validate sampling rules configuration on start and reload.

### DIFF
--- a/cmd/samproxy/main.go
+++ b/cmd/samproxy/main.go
@@ -64,7 +64,11 @@ func main() {
 		Version: version,
 	}
 
-	c, err := config.NewConfig(opts.ConfigFile, opts.RulesFile, func(err error) { a.Logger.Error().WithField("error", err).Logf("error reloading config") })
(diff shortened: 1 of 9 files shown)


cmd/samproxy/main.go
```diff
@@ -60,7 +60,11 @@ func main() {
 		os.Exit(0)
 	}
 
-	c, err := config.NewConfig(opts.ConfigFile, opts.RulesFile)
+	a := app.App{
+		Version: version,
+	}
+
+	c, err := config.NewConfig(opts.ConfigFile, opts.RulesFile, func(err error) { a.Logger.Error().WithField("error", err).Logf("error reloading config") })
```
martin308, line 67: I wonder if it's worth checking that the logger is not nil here? I think the chances are pretty slim that the config reloads in the time before the logger gets assigned, but better to be safe than sorry?

martin308 (approved):
makes sense to me

PR #145, opened 2020-09-21
only re-route incoming spans

ianwilkes:
Don't sent spans from peers to other peers - we'd rather collect these on the "wrong" host then bounce them around the cluster forever.

Note this doesn't fix the event endpoint, which shouldn't see any peer traffic anyway. I think that endpoint needs some remedial attention to remove copy pasta and I'll deal with that in a separate PR.

### DIFF
--- a/route/route.go
+++ b/route/route.go
@@ -409,10 +409,12 @@ func (r *Router) batch(w http.ResponseWriter, req *http.Request) {
 			r.UpstreamTransmission.EnqueueEvent(ev)
 			continue
 		}
-		// ok, we're a span. Figure out if we should handle locally or pass on to a peer
+		// ok, we're a span. Figure out if we should handle locally or pass on
+		// to a peer. We won't pass anything that came from a peer; if there's
+		// confusion about who owns what, this avoids loops.
 		targetShard := r.Sharder.WhichShard(traceID)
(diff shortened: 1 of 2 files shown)


route/route.go
```diff
@@ -409,10 +409,12 @@ func (r *Router) batch(w http.ResponseWriter, req *http.Request) {
...
 			continue
 		}
-		// ok, we're a span. Figure out if we should handle locally or pass on to a peer
+		// ok, we're a span. Figure out if we should handle locally or pass on
+		// to a peer. We won't pass anything that came from a peer; if there's
+		// confusion about who owns what, this avoids loops.
 		targetShard := r.Sharder.WhichShard(traceID)
-		if !targetShard.Equals(r.Sharder.MyShard()) {
-			r.Metrics.IncrementCounter(r.incomingOrPeer + "_router_peer")
+		if r.incomingOrPeer == "incoming" && !targetShard.Equals(r.Sharder.MyShard()) {
```
martin308, line 416: the collection of magic string based features grows!

martin308 (approved):
👾

route/route.go
ianwilkes: I would have used a bool, but I'm not rewriting this whole thing today.

PR #148, opened 2020-09-21
peers list should not be required in config

asdvalenzuela:
Peers list is not required if file type is redis. https://github.com/honeycombio/samproxy/issues/147

### DIFF
--- a/config/file_config.go
+++ b/config/file_config.go
@@ -70,7 +70,7 @@ type HoneycombMetricsConfig struct {
 
 type PeerManagementConfig struct {
 	Type                    string   `validate:"required,oneof= file redis"`
-	Peers                   []string `validate:"required,dive,url"`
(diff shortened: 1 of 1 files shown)


config/file_config.go
```diff
@@ -70,7 +70,7 @@ type HoneycombMetricsConfig struct {
 
 type PeerManagementConfig struct {
 	Type                    string   `validate:"required,oneof= file redis"`
-	Peers                   []string `validate:"required,dive,url"`
+	Peers                   []string `validate:"dive,url"`
```
martin308, line 73: can we do some conditional require here? It is required for `file` type right? Oh wait it has a default! Is that why?

martin308 (approved):

config/file_config.go
asdvalenzuela: It does have a default set 👍

PR #149, opened 2020-09-22
consolidate event handling code

ianwilkes:
Share the important bits between the batch and event handlers. Add an integration test for the event endpoint. Add testing of meta fields. Remove unused event metadata. Clean up some test code. Remove dead code.

### DIFF
--- a/route/route.go
+++ b/route/route.go
@@ -202,8 +202,13 @@ func (r *Router) version(w http.ResponseWriter, req *http.Request) {
 func (r *Router) event(w http.ResponseWriter, req *http.Request) {
 	defer req.Body.Close()
 
-	// get out request ID for logging
-	reqBod, _ := ioutil.ReadAll(req.Body)
+	bodyReader, err := r.getMaybeCompressedBody(req)
+	if err != nil {
+		r.handlerReturnWithError(w, ErrPostBody, err)
+		return
+	}
+
+	reqBod, _ := ioutil.ReadAll(bodyReader)
 	ev, err := r.requestToEvent(req, reqBod)
 	if err != nil {
 		r.handlerReturnWithError(w, ErrReqToEvent, err)
--- a/route/route.go
+++ b/route/route.go
@@ -204,19 +204,7 @@ func (r *Router) event(w http.ResponseWriter, req *http.Request) {
 	defer req.Body.Close()
 
 	// get out request ID for logging
-	reqID := req.Context().Value(types.RequestIDContextKey{})
-	debugLog := r.iopLogger.Debug().WithField("request_id", reqID)
-
 	reqBod, _ := ioutil.ReadAll(req.Body)
-	var trEv eventWithTraceID
-	// pull out just the trace ID for use in routing
-	err := unmarshal(req, bytes.NewReader(reqBod), &trEv)
-	if err != nil {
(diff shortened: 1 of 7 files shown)


route/route.go
```diff
@@ -203,114 +202,24 @@ func (r *Router) version(w http.ResponseWriter, req *http.Request) {
 func (r *Router) event(w http.ResponseWriter, req *http.Request) {
 	defer req.Body.Close()
 
-	// get out request ID for logging
-	reqID := req.Context().Value(types.RequestIDContextKey{})
-	debugLog := r.iopLogger.Debug().WithField("request_id", reqID)
```
martin308, line 208: we removed this from here but not from the batch endpoint, should it be only in the batch endpoint?

route/route.go
```diff
@@ -203,114 +202,24 @@ func (r *Router) version(w http.ResponseWriter, req *http.Request) {
...
-			"api_host": ev.APIHost,
-			"dataset":  ev.Dataset,
-			"trace_id": trEv.TraceID,
-			"peer":     targetShard.GetAddress(),
-		}).Logf("Sending span to my peer")
-		r.PeerTransmission.EnqueueEvent(ev)
-		return
-	}
-	// we're supposed to handle it
+	reqBod, _ := ioutil.ReadAll(bodyReader)
```
martin308, line 211: in the batch endpoint we handle an error being returned from this call, should they both do the same thing?

route/route.go
```diff
@@ -561,7 +459,7 @@ type batchedEvent struct {
 
 func (b *batchedEvent) getEventTime() time.Time {
 	if b.MsgPackTimestamp != nil {
-		return *b.MsgPackTimestamp
+		return b.MsgPackTimestamp.UTC()
```
martin308, line 462: what are the implications of this?

route/route.go
```diff
@@ -631,16 +529,6 @@ func makeDecoders(num int) (chan *zstd.Decoder, error) {
...
-		// First try to decode with the fast but fussy msgp. If that fails,
-		// defer to the much more friendly msgpack.
-		if decodable, ok := v.(msgp.Decodable); ok {
-			err := msgp.Decode(data, decodable)
-
-			if err == nil {
-				return nil
-			}
-		}
-
```
martin308, line 643: why do we not want to do this?

martin308 (approved):
Seems like a pretty good cleanup, just have a couple of questions about some the bits

route/route.go
```diff
@@ -203,114 +202,24 @@ func (r *Router) version(w http.ResponseWriter, req *http.Request) {
 func (r *Router) event(w http.ResponseWriter, req *http.Request) {
 	defer req.Body.Close()
```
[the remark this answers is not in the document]
ianwilkes: foo_router_event is already used to count non-span events. I'll rename that one.

route/route.go
ianwilkes: good catch, I put it back

route/route.go
ianwilkes: It prevents us from creating serialized times with the local server timezone, ie `2020-09-24T09:38:59.169594-07:00`. Functionally it shouldn't matter, especially since our servers should all be on UTC, but for local tests it keeps everything consistent.

Go's local-time-by-default is infuriating AF.

route/route.go
ianwilkes: We never did do it. This codebase contains no msgp.Decodable data structures. Looks like it was just cargo-culted over from shepherd.

PR #156, opened 2020-09-29
exclude TestErrorReloading from go_test run with --race

asdvalenzuela:


config/config_test_reload_error.go
```diff
@@ -0,0 +1,89 @@
+// +build all
```
martin308, line 1: is it worth adding `!race` here so it's clear we don't want to run this with the race detector?

martin308 (changes requested):
✨ one question about including the `!race` tag

Also I think the new test file will also need to end in `*_test.go` or it won't be run

martin308 (approved):
🚀

PR #157, opened 2020-09-29
Make response consistent with Honeycomb API

martin308:
The alive check for the honeycomb API uses a string value "yes" to respond to its health check.

https://api.honeycomb.io/x/alive

Update to match


martin308:
I think the tests are failing due to the race detector which will be fixed in #156

asdvalenzuela (approved):

PR #158, opened 2020-09-30
add dataset as field when logging sampler type

asdvalenzuela:
`DEBU[0028] created implementation for sampler type *config.DynamicSamplerConfig  dataset="API Calls"`

### DIFF
--- a/config/sampler_config.go
+++ b/config/sampler_config.go
@@ -5,8 +5,8 @@ type DeterministicSamplerConfig struct {
 }
 
 type DynamicSamplerConfig struct {
-	SampleRate                   int64 `validate:"required,gte=1"`
-	ClearFrequencySec            int64
+	SampleRate                   int64    `validate:"required,gte=1"`
+	ClearFrequencySec            int64    `validate:"required,gt=0"`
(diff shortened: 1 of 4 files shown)


sample/sample.go
```diff
@@ -28,23 +28,26 @@ func (s *SamplerFactory) GetSamplerImplementationForDataset(dataset string) Samp
 		return nil
 	}
 
-	s.Logger.Debug().WithField("dataset", dataset).Logf("creating sampler implementation")
```
martin308, line 31: could this line move down to the bottom before we return and do something like the default case for the log message?

config/sampler_config.go
```diff
@@ -5,11 +5,11 @@ type DeterministicSamplerConfig struct {
 }
 
 type DynamicSamplerConfig struct {
-	SampleRate                   int64    `validate:"required,gte=1"`
-	ClearFrequencySec            int64    `validate:"required,gt=0"`
```
martin308, line 9: maybe this should be in a separate PR?

martin308 (approved):

PR #167, opened 2020-10-02
Add customisable dry run field name

martin308:
Allows customizable dry run field name and updates the default to reference refinery and not samproxy

### DIFF
--- a/rules.toml
+++ b/rules.toml
@@ -6,7 +6,7 @@
 # and sends all traces regardless
 # DryRun = false
 
-# DryRunFieldName - the key to add to use to add to event data when using DryRun mode above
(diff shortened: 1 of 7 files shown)


rules.toml
```diff
@@ -6,6 +6,9 @@
 # and sends all traces regardless
 # DryRun = false
 
+# DryRunFieldName - the key to add to use to add to event data when using DryRun mode above
```
asdvalenzuela, line 9:
```suggestion
# DryRunFieldName - the key to add to use to add to event data when using DryRun mode above, defaults to refinery_kept
```

asdvalenzuela (approved):

PR #186, opened 2020-10-21
Handle all the datatypes in the EMA sampler

martin308:
Now that we support msgpack the types that might be in the data hash are more varied than they used to be. 

Adds the ability to format any type as a string in order to handle this

### DIFF
--- a/sample/dynamic_ema_test.go
+++ b/sample/dynamic_ema_test.go
@@ -21,7 +21,7 @@ func TestDynamicEMAAddSampleRateKeyToTrace(t *testing.T) {
 
 	sampler := &EMADynamicSampler{
 		Config: &config.EMADynamicSamplerConfig{
-			FieldList:                    []string{"http.status_code"},
+			FieldList:                    []string{"http.status_code", "request.path", "app.team.id", "important_field"},
 			AddSampleRateKeyToTrace:      true,
 			AddSampleRateKeyToTraceField: "meta.key",
 		},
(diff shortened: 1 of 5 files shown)


martin308:
It looks like the regular dynamic sampler is also going to be impacted by this, will update that as well

sample/dynamic_ema_test.go
```diff
@@ -43,11 +46,15 @@ func TestDynamicEMAAddSampleRateKeyToTrace(t *testing.T) {
...
 	assert.Len(t, spans, spanCount, "should have the same number of spans as input")
 	for _, span := range spans {
 		assert.Equal(t, span.Event.Data, map[string]interface{}{
-			"http.status_code": "200",
-			"meta.key":         "200•,",
+			"http.status_code": 200,
+			"app.team.id":      float64(4),
+			"important_field":  true,
+			"request.path":     "/{slug}/fun",
+			"meta.key":         "4•,200•,true•,/{slug}/fun•,",
```
maplebed, line 57: while we're in here improving the test, would it be useful to ensure the "multiple instances of the same field get collapsed to their smallest set of unique values" behavior is working as well?  This may be pushing the boundaries of this test and an explicit test of the `buildKey` function would fit better, but I don't have a strong opinion there.
(Since there are 5 spans in this trace, the "collapse similar values" portion of that is being tested by only having one `200` show up instead of 5, but if we had 3 `200`s and 2 `429`s it'd be nice to see `200•429` as the value in the key.)

maplebed (approved):
This LGTM! Approving now to let you decide whether to expand the test. Either road is fine by me.

sample/dynamic_ema_test.go
martin308: added some tests to cover this scenario!

PR #191, opened 2020-11-11
Make TotalThroughputSampler a valid sampler.

magnusstahre:


martin308 (changes requested):
It's probably worth adding a bunch of extra tests to cover these points we missed! 

Could we add:

- a sample config to rules.toml similar to the others
- a test in config_test for this sampler config

magnusstahre:
@martin308 updated the PR per the requests, please take a look!

asdvalenzuela (approved):

martin308 (approved):
✨

PR #195, opened 2020-11-18
Add quotes for example DryRunFieldName value

### LINKED ISSUE #194: Sample config is missing some quotes
Inside the sample [rules.toml](https://github.com/honeycombio/refinery/blob/442e269/rules.toml#L10) there are sample values (commented out) for setting these:
```
DryRun = true
...
DryRunFieldName = refinery_kept
```

If you uncomment them the second needs quotes:
```
% docker run --rm -it --volume $(pwd):/etc/refinery --expose 8080 --publish 44444:8080 honeycombio/refinery:latest /usr/bin/refinery
unable to load config: While parsing config: (10, 19): no value can start with r
```

MikeGoldsmith:
Fixes #194.

The example config includes the property `DryRunFieldName` that has a string value. However, the value is not wrapped in quotes which makes the config fail to load.

### DIFF
--- a/rules.toml
+++ b/rules.toml
@@ -7,7 +7,7 @@
 # DryRun = false
 
 # DryRunFieldName - the key to add to use to add to event data when using DryRun mode above, defaults to refinery_kept
-# DryRunFieldName = refinery_kept
+# DryRunFieldName = "refinery_kept"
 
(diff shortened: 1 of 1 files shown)


martin308 (approved):
✅

PR #196, opened 2020-11-18
add blank rules file to build-pkg, use old rules file as example

asdvalenzuela:
In adding more detail to the getting started docs, we noticed that the rules file had been left out of the build package. Since the existing rules file was more of an example file, and so as not to pollute the config with those examples, we renamed `rules.toml` to `rules_complete.toml`. Since defaults for the rules configuration are set in `file_config.go`, `rules.toml` is now blank until folks start adding their own rules. 

Drive-by cleanup of a few things in `file_config.go` and regarding our config tests.


config/file_config.go
```diff
@@ -146,6 +143,8 @@ func NewConfig(config, rules string, errorCallback func(error)) (Config, error)
 
 	r.SetDefault("Sampler", "DeterministicSampler")
```
MikeGoldsmith, line 144: Does these need to happen after the config has been loaded? Could they all be done beforehand?

asdvalenzuela: They don't need to happen after. I think this bit is just meant to continue the distinction between the general config file and the rules file, since these values are all set in the rules file. @martin308 Is that correct?

martin308: we need to do them beforehand AFAIK, which is the case here as we don't load the config until line 150

MikeGoldsmith: Ah, just seen the `SetDefaults` are on the `r` not the `c` object so can't happen any sooner 👍

MikeGoldsmith (approved):

martin308 (approved):
✨

PR #199, opened 2020-12-21
[logger] Properly format duration

dstrelau:
`%d` prints ns, not ms. `%s` will format appropriately for all values.


martin308 (approved):
The actual code change here looks good. Not sure why the build would be failing on the race detection from this change though 🤔

asdvalenzuela (approved):

PR #200, opened 2020-12-22
Rate limit log messages sent to honeycomb logger

vreynolds:
We have encountered a failure case in Refinery where logging overwhelms downstream shepherd. Adding configurable rate-limiting to the Honeycomb logger to prevent this.

* prevent overwhelming the downstream service, when getting a spike of
similar messages (e.g. the same error)
* the sampling strategy will rate-limit per [log level + log message] key, with a configurable throughput
* sampler is disabled by default, enabled via config

### Considerations
* I looked into adding an integration test. I could see the outbound events traffic by injecting into the http transport, but the sampling is determined by events/sec so it was a spaghetti-o of thread.sleeps and log messages.
  * Realistically, we'd need to inject the Honeycomb WriterSender as the logger transmission (similarly to the app integration test), but that felt like a larger change than I was up for this round.
* I didn't update anything in the README, because that seemed too high level, but I will update the public config docs as a follow-up: https://docs.honeycomb.io/working-with-your-data/best-practices/refinery/configuration/


vreynolds:
@martin308 @asdvalenzuela would love your thoughts here, especially around testing. It seemed that this was largely a configuration change, and there aren't really any tests around log producers currently.

config/file_config.go
```diff
@@ -131,6 +132,7 @@ func NewConfig(config, rules string, errorCallback func(error)) (Config, error)
 	c.SetDefault("UpstreamBufferSize", libhoney.DefaultPendingWorkCapacity)
 	c.SetDefault("PeerBufferSize", libhoney.DefaultPendingWorkCapacity)
 	c.SetDefault("MaxAlloc", uint64(0))
+	c.SetDefault("HoneycombLogger.LoggerSamplerThroughput", 5)
```
vreynolds, line 135: @martin308 @asdvalenzuela this didn't work to set the default. Did we miss something?

logger/honeycomb.go
```diff
@@ -66,6 +69,11 @@ func (h *HoneycombLogger) Start() error {
 		}
 	}
 
+	err = h.maybeStartSampler(loggerConfig)
```
vreynolds, line 72: Not my favorite: passing the parsed config into the receiver func, but parsing the logger config _again_ feels worse.

vreynolds:
@maplebed @martin308 @asdvalenzuela could ya'll give it another 👀 , it's no longer live reloading the sampler

asdvalenzuela (approved):

PR #202, opened 2021-01-07
Add OTLP ingest

MikeGoldsmith:
Adds OTLP ingest support to refinery.

Initially this PR adds ingest support but does not add support for sending events to peer refinery processes or to the Honeycomb ingest API over gRPC and continues to use HTTP/JSON for that communication.

The primary use-case for adding OTLP here is for ingesting spans directly from OTel SDKs and Collector using OTLP over gRPC. This will allow users of OpenTelemetry to send data to refinery without having to use a custom span exporter and instead use the SDK provided OTLP exporter.


route/route.go
```diff
@@ -330,6 +363,124 @@ func (r *Router) batch(w http.ResponseWriter, req *http.Request) {
...
+					APIKey:     apiKey,
+					Dataset:    dataset,
+					SampleRate: uint(sampleRate),
+					Timestamp:  timestamp,
+					Data:       eventAttrs,
+				}
+
+				err := r.processEvent(event, requestId)
+				if err != nil {
+					// log failure? We don't send per-event responses back to sender for OTLP
```
asdvalenzuela, line 475: Why not log? If we're worried about too many logs we could add a metric instead.

asdvalenzuela (approved):

route/route.go
MikeGoldsmith: Not sure what to do. Log's are useful here as it would give context on _why_ it failed, but potentially could generate a lot of noise.

config/file_config.go
```diff
@@ -53,8 +53,9 @@ func (r *RulesBasedSamplerConfig) String() string {
 }
 
 type configContents struct {
-	ListenAddr         string        `validate:"required"`
-	PeerListenAddr     string        `validate:"required"`
+	ListenAddr         string `validate:"required"`
```
vreynolds, line 56: go fmt?

MikeGoldsmith: I too thought the formatting was wrong. but running `go fmt` doesn't change it. I think it's because the `GRPCListenAddr` entry below doesn't have a descriptor which seems to reset the spacing.

route/route.go
```diff
@@ -165,6 +177,24 @@ func (r *Router) LnS(incomingOrPeer string) {
...
+		r.iopLogger.Error().Logf("failed to get grpc listen addr config: %s", err)
+		return
+	} else if grpcAddr != "" {
+		l, err := net.Listen("tcp", grpcAddr)
+		if err != nil {
+			r.iopLogger.Error().Logf("failed to listen to grpc addr: " + grpcAddr)
+		}
+
+		r.iopLogger.Info().Logf("gRPC listening on %s", grpcAddr)
+		serverOpts := []grpc.ServerOption{}
```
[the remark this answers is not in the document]
MikeGoldsmith: Add log entry in commit [c58a30c](https://github.com/honeycombio/refinery/pull/202/commits/c58a30ccc3e99fe270868a834f1f4845eabf23f3)

route/route.go
MikeGoldsmith: Add log entry in commit [1d4df3e](https://github.com/honeycombio/refinery/pull/202/commits/1d4df3ed5ba2667176b479f2bf7c1ef69f754ab8)

vreynolds:
@MikeGoldsmith I tried to review to the best of my knowledge :)

- When I set this branch up locally, Refinery fails to start, because it attempts to bind a peer GRPC connection to the same `GRPCListenAddr` as main. I could be doing something wrong? Or we need to add a configuration for a peer GRPC address.
- Need to add new config value to config_complete.toml (unless you plan on doing that in a separate PR)
- [config live reload](https://houndsh.slack.com/archives/CDE86Q55X/p1608749552192000) is not something we intend to support, so I think you can omit those tests

route/route.go
```diff
@@ -330,6 +375,117 @@ func (r *Router) batch(w http.ResponseWriter, req *http.Request) {
...
+		resourceAttrs := make(map[string]interface{})
+
+		if resourceSpan.Resource != nil {
+			addAttributesToMap(resourceAttrs, resourceSpan.Resource.Attributes)
+		}
+
+		for _, librarySpan := range resourceSpan.InstrumentationLibrarySpans {
+			library := librarySpan.InstrumentationLibrary
+			if library != nil {
+				if library.Name != "" {
```
paulosman, line 414: nit, but for clarity, we should stick to checking for an empty string or checking that the length is greater than 0, not both. I vote `if len(library.Name) > 0`.

route/route.go
```diff
@@ -330,6 +375,117 @@ func (r *Router) batch(w http.ResponseWriter, req *http.Request) {
...
+					"duration_ms":    float64(span.EndTimeUnixNano-span.StartTimeUnixNano) / float64(time.Millisecond),
+					"status_code":    int32(span.Status.Code),
+				}
+				if span.ParentSpanId != nil {
+					eventAttrs["trace.parent_id"] = hex.EncodeToString(span.ParentSpanId)
+				}
+				if r.getSpanStatusCode(span.Status) == trace.Status_STATUS_CODE_ERROR {
+					eventAttrs["error"] = true
+				}
+				if len(span.Status.Message) > 0 {
```
paulosman, line 441: See above.

route/route.go
```diff
@@ -330,6 +375,117 @@ func (r *Router) batch(w http.ResponseWriter, req *http.Request) {
...
+				sampleRate := 1
+				if eventAttrs["sampleRate"] != nil {
+					switch eventAttrs["sampleRate"].(type) {
+					case string:
+						sampleRate, err = strconv.Atoi(eventAttrs["sampleRate"].(string))
+						if err != nil {
+							r.Logger.Info().WithField("Unable to parse sampleRate in OTLP span attribute", eventAttrs["sampleRate"])
+						}
+					case int:
+						sampleRate = eventAttrs["sampleRate"].(int)
```
paulosman, line 457: Could add a `default:` here with a log statement stating that it couldn't parse the sampleRate (in case it's something other than a string or an int - say, a float).

route/route.go
```diff
@@ -546,3 +702,104 @@ func unmarshal(r *http.Request, data io.Reader, v interface{}) error {
...
+	apiKey = getFirstValueFromMetadata(types.APIKeyHeader, md)
+	if apiKey == "" {
+		apiKey = getFirstValueFromMetadata(types.APIKeyHeaderShort, md)
+	}
+	dataset = getFirstValueFromMetadata(types.DatasetHeader, md)
+
+	return apiKey, dataset
+}
+
+// safeGetValueFromMeta returns the first value of a metadata entry using a
```
paulosman, line 716: The other helper functions are jealous that this one gets a comment and the rest don't. :P Also the method name is different. 

```suggestion
// getFirstValueFromMetadata returns the first value of a metadata entry using a
```

route/route_test.go
```diff
@@ -234,3 +235,63 @@ func TestUnmarshal(t *testing.T) {
 		t.Error("Expecting", now, "Received", b)
 	}
 }
+
+func TestGetAPIKeyAndDatasetFromMetadataCaseInsensitive(t *testing.T) {
```
paulosman, line 239: I'd like to see some tests covering the parsing of the sample rate from the OTLP request - especially since we're doing some type assertions.

paulosman (approved):
Approving w/ some suggestions. Overall looks good!

route/route.go
```diff
@@ -165,6 +179,32 @@ func (r *Router) LnS(incomingOrPeer string) {
 		Handler: muxxer,
 	}
 
+	// GRPC listen addr is optional, err means addr was not empty and invalid
+	grpcAddr, err := r.Config.GetGRPCListenAddr()
+	if err != nil {
+		r.iopLogger.Error().Logf("failed to get grpc listen addr config: %s", err)
+		return
+	} else if grpcAddr != "" {
```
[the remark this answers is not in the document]
MikeGoldsmith: This was fixed in [076c089](https://github.com/honeycombio/refinery/pull/202/commits/076c08952d4342f5fcf854b48303faacd5e9e37d) by separating the grpcAddr assignment based in "incoming" and then checking if it's set.

route/route.go
MikeGoldsmith: It's copied from hound/shepherd 😄

vreynolds (approved):
🏅

route/route.go
MikeGoldsmith: Added default case that creates err with info that goes in debug logs.

route/route_test.go
MikeGoldsmith: Added testing for getting sampleRatee from attr. Includes missing and different value types 👍

route/route.go
```diff
@@ -166,6 +187,27 @@ func (r *Router) LnS(incomingOrPeer string) {
 		Handler: muxxer,
 	}
 
+	if len(grpcAddr) > 0 {
+		l, err := net.Listen("tcp", grpcAddr)
+		if err != nil {
+			r.iopLogger.Error().Logf("failed to listen to grpc addr: " + grpcAddr)
```
asdvalenzuela, line 193: Do we want to return if there's an error here?

route/route.go
```diff
@@ -337,6 +382,108 @@ func (r *Router) batch(w http.ResponseWriter, req *http.Request) {
...
+func (r *Router) Export(ctx context.Context, req *collectortrace.ExportTraceServiceRequest) (*collectortrace.ExportTraceServiceResponse, error) {
+
+	md, ok := metadata.FromIncomingContext(ctx)
+	if !ok {
+		r.Logger.Error().Logf("Unable to retreive metadata from OTLP request.")
+		return &collectortrace.ExportTraceServiceResponse{}, nil
+	}
+
+	var requestID types.RequestIDContextKey
+	debugLog := r.iopLogger.Debug().WithField("request_id", requestID)
```
asdvalenzuela, line 394: It might be helpful to have a note here about why requestID is blank.

asdvalenzuela (approved):

route/route.go
MikeGoldsmith: I think we should fail to start if a grpc address has been set and we're unable to bind to the port. We return if the HTTP server is unable to start. If something is reliant on gRPC running and the process doesn't fail to start, it could look like we're working but would reject gRPC traffic.

route/route.go
MikeGoldsmith: Added comment in [eaaaa52](https://github.com/honeycombio/refinery/pull/202/commits/eaaaa5227dbc67f04cd07fe283bba3d1ffabe4a0).

PR #204, opened 2021-01-13
Add /debug/trace/{traceID} endpoint

MikeGoldsmith:
Add a diagnostics HTTP GET endpoint that returns the debug trace information, including the assigned node address using the configured sharding algorithm.

The endpoint can be called like this:
`http://localhost:9000/debug/trace/123abc`

The response is a JSON object two two properties and looks like the following:
`{"traceID":"123abc","node":"http://localhost:12345"}`

### DIFF
--- a/route/route.go
+++ b/route/route.go
@@ -132,7 +132,7 @@ func (r *Router) LnS(incomingOrPeer string) {
 	muxxer.HandleFunc("/alive", r.alive).Name("local health")
 	muxxer.HandleFunc("/panic", r.panic).Name("intentional panic")
 	muxxer.HandleFunc("/version", r.version).Name("report version info")
-	muxxer.HandleFunc("/getNode/{traceID}", r.getNode).Name("get node address for given trace ID")
+	muxxer.HandleFunc("/debug/trace/{traceID}", r.debugTrace).Name("get debug information for given trace ID")
 
 	// require an auth header for events and batches
 	authedMuxxer := muxxer.PathPrefix("/1/").Methods("POST").Subrouter()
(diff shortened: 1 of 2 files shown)


route/route.go
```diff
@@ -132,6 +132,7 @@ func (r *Router) LnS(incomingOrPeer string) {
 	muxxer.HandleFunc("/alive", r.alive).Name("local health")
 	muxxer.HandleFunc("/panic", r.panic).Name("intentional panic")
 	muxxer.HandleFunc("/version", r.version).Name("report version info")
+	muxxer.HandleFunc("/getNode/{traceID}", r.getNode).Name("get node address for given trace ID")
```
vreynolds, line 135: Preference for `/get-node/` or `/node/` or `/shard/` here, but that's NBD

vreynolds (approved):
Looks good! Is there any security considerations here with exposing a hostname on a non-authed endpoint? I'm assuming it's information that's already public.

maplebed:
✨

route/route.go
paulosman: +1 

Case sensitivity in URL paths stands out to me so I'd prefer to ignore it. 

Also, thinking about future-proofing - what if we just had a debug endpoint that accepted a traceID? 

```
muxxer.HandleFunc("/debug/{traceID}", r.debug).Name("get debug information about a specific trace")
```

Then we're free to add more information to this as we go.

MikeGoldsmith:
> Looks good! Is there any security considerations here with exposing a hostname on a non-authed endpoint? I'm assuming it's information that's already public.

Currently Refinery is expected to only work within a secure network and doesn't setup any authentication credentials so I think we're good for now. If we decide to add authentication later, this endpoint should be secured too.

route/route.go
MikeGoldsmith: Yeah, I think that's a good idea and agree having all lower cased paths is a good idea.

I think the path `/debug/trace/{traceID}` is more descriptive and more REST like incase we want to provide debug info on other types of data in the future.

vreynolds (approved):
✅

PR #209, opened 2021-02-17
detect int64 sample rate

paulosman:
Detect int64 / int32 sample rates. Some OTLP exporters will send them as these types - so handle appropriately.

### DIFF
--- a/route/route.go
+++ b/route/route.go
@@ -819,7 +819,9 @@ func getSampleRateFromAttributes(attrs map[string]interface{}) (int, error) {
 	var err error
 	switch v := attrs[sampleRateKey].(type) {
 	case string:
-		sampleRate, err = strconv.Atoi(v)
+		var i int64
+		i, err = strconv.ParseInt(v, 10, 32)
+		sampleRate = int(i)
 	case int:
 		if v > int32MaxValue {
 			sampleRate = int32MaxValue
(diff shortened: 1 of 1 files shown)


asdvalenzuela (approved):

route/route.go
```diff
@@ -804,21 +805,40 @@ func (r *Router) getSpanStatusCode(status *trace.Status) trace.Status_StatusCode
...
-		switch attributes["sampleRate"].(type) {
-		case string:
-			sampleRate, err = strconv.Atoi(attributes["sampleRate"].(string))
-		case int:
-			sampleRate = attributes["sampleRate"].(int)
-		default:
-			err = fmt.Errorf("Unrecognised sampleRate datatype - %T", attributes["sampleRate"])
+	switch v := attrs[sampleRateKey].(type) {
+	case string:
+		sampleRate, err = strconv.Atoi(v)
```
asdvalenzuela, line 822: Do we need to check for this also or does it not matter as much here? `strconv.ParseInt(v, 10, 32)`

paulosman: I don't think so. From the Golang docs:

`Atoi is equivalent to ParseInt(s, 10, 0), converted to type int.`

So I think this is safe enough.

paulosman: > Do we need to check for this also or does it not matter as much here? `strconv.ParseInt(v, 10, 32)`

Yeah, good point - let's keep it consistent with shepherd. I'll update.

vreynolds (approved):
🔢

PR #210, opened 2021-02-22
Error response from libhoney does not always mean a non-20x response

vreynolds:
* it could be a 200 API response that got an error on parsing the body

### DIFF
--- a/transmit/transmit.go
+++ b/transmit/transmit.go
@@ -160,7 +160,7 @@ func (d *DefaultTransmission) processResponses(
 				if r.Err != nil {
 					log = log.WithField("error", r.Err.Error())
 				}
-				log.Logf("non-20x response when sending event")
+				log.Logf("error when sending event")
 				if honeycombAPI == apiHost {
 					// if the API host matches the configured honeycomb API,
 					// count it as an API error


transmit/transmit.go
```diff
@@ -160,7 +160,7 @@ func (d *DefaultTransmission) processResponses(
 				if r.Err != nil {
 					log = log.WithField("error", r.Err.Error())
 				}
-				log.Logf("non-20x response when sending event")
+				log.Logf("error when sending event")
```
robbkidd, line 163: Would `"error sending event or processing the response"` be more helpful to avoid narrowing in the mind of the reader the scope of where the error popped up?

vreynolds: Mentioning response processing feels like leaking libhoney internals into here, but maybe there's a different way to say it that encapsulates sending/processing? 

Error transmitting event?
Event transmission error?

Also, with the additional context in libhoney, it will say something like:
`ERROR: error when sending event: got OK HTTP response, but couldn't read response body`

robbkidd: Works for me then!

robbkidd (approved):
👍

PR #213, opened 2021-02-23
Add tests for int64 sample rate values

### LINKED ISSUE #211: Add tests for int64 sample rates
Handling of int64 sample rates was added in #209. We should extend the tests in [route_test.go](https://github.com/honeycombio/refinery/blob/main/route/route_test.go#L301) to test int64 sample rates are handled as expected and are not accidentally broken in the future.

Add test for the following cases:
- int64 value lower/equal to than int32 max - uses sample rate value
- int64 value greater than int32 max - uses int32 max value

MikeGoldsmith:
Resolves #211.

Adds testing for parsing int64 sample rate values. Test cases include parsing a value that can be parsed into an int32 and a value greater than int32 max where the int32 max value is used instead.

Additionally, replace the custom max int32 value with the math package const `math.MaxInt32`.

### DIFF
--- a/route/route.go
+++ b/route/route.go
@@ -48,7 +48,6 @@ const (
 	traceIDLongLength      = 16
 	GRPCMessageSizeMax int = 5000000 // 5MB
 	defaultSampleRate      = 1
-	int32MaxValue          = 2147483647 // (2**31)-1
 )
 
 type Router struct {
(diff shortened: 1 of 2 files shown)


vreynolds (approved):
👍

PR #216, opened 2021-03-01
Update rules-based sampler examples

shelbyspees:
The name `500 errors` doesn't describe the rule completely, and it wasn't clear that [rule conditions use AND](https://github.com/honeycombio/refinery/blob/a861dee6a345f497ce9ae08f458e6064cbb322e4/sample/rules.go#L122) (am I understanding that correctly?).

I also changed the rule for 200 responses to downsample instead of dropping them. `drop` is a special case that should be used rarely on entire traces, and I think using it here makes the example less realistic. This is also an opportunity to give a large number for the `SampleRate` to help set expectations for users that sending 1/1000 "boring" events is safe and reasonable.

### DIFF
--- a/rules_complete.toml
+++ b/rules_complete.toml
@@ -208,8 +208,9 @@ SampleRate = 1
 	Sampler = "RulesBasedSampler"
 
 	[[dataset4.rule]]
-		name = "500 errors"
+		name = "slow 500 errors"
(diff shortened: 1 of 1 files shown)


shelbyspees:
Need to update tests. Will reopen if I get bandwidth.

PR #217, opened 2021-03-01
Update README with recommendation to use IPs when DNS is unreliable or slow

robbkidd:
Also some tweaks to the wording to bring in line with the recent documentation updates.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,9 @@ Refinery is a trace-aware sampling proxy. It collects spans emitted by your appl
 
 ## Setting up Refinery
 
-Refinery is designed to sit within your infrastructure where all sources of Honeycomb events (aka spans if you're doing tracing) can reach it. A standard deployment will have a cluster of servers running Refinery accessible via a load balancer. Refinery instances must be able to communicate with each other to concentrate traces on single servers.
+Refinery is designed to sit within your infrastructure where all sources of Honeycomb events (aka spans if you're doing tracing) can reach it.
+A standard deployment will have a cluster of two or more Refinery processes accessible via a separate load balancer.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -10,7 +10,9 @@ Refinery is a trace-aware sampling proxy. It collects spans emitted by your appl
 
 ## Setting up Refinery
 
-Refinery is designed to sit within your infrastructure where all sources of Honeycomb events (aka spans if you're doing tracing) can reach it. A standard deployment will have a cluster of servers running Refinery accessible via a load balancer. Refinery instances must be able to communicate with each other to concentrate traces on single servers.
+Refinery is designed to sit within your infrastructure where all sources of Honeycomb events (aka spans if you're doing tracing) can reach it.
+A standard deployment will have a cluster of two or more Refinery processes accessible via a separate load balancer.
+Refinery processes must be able to communicate with each other to concentrate traces on single servers.
```
lizthegrey, line 15: again, using common language, "correct shard for the trace" >> "concentrate traces on single servers"

lizthegrey (approved):

MikeGoldsmith (approved):
Other than :+1: for @lizthegrey's comment, I think this looks good.

README.md
robbkidd: I'm a bit unclear on the call to action for this comment. Are you suggesting something like the following change?

```suggestion
Refinery processes must be able to communicate with each other to shuttle traces to the correct shard.
```

Should we bring up sharding in this introductory paragraph? Currently, Sharding isn't introduced until the architecture section of this README much farther down.

robbkidd:
I think I'll merge this as-is and we can pick up adding more about sharding in another PR.

PR #228, opened 2021-03-22
Log sampler config and validation errors

MikeGoldsmith:
When loading sampler rules from file based config, it can be difficult to see what rules have been set or any validation errors.

This PR logs the sampler config in `Debug` level and also any unmarshalling / validation errors in `Warn` level.

### DIFF
--- a/config/file_config.go
+++ b/config/file_config.go
@@ -12,6 +12,7 @@ import (
 	"github.com/go-playground/validator"
 	libhoney "github.com/honeycombio/libhoney-go"
 	viper "github.com/spf13/viper"
+	"github.com/sirupsen/logrus"
 )
 
 type fileConfig struct {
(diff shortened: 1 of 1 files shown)


config/file_config.go
```diff
@@ -311,11 +314,13 @@ func (f *fileConfig) validateSamplerConfigs() error {
 			}
 			err := f.rules.Unmarshal(i)
 			if err != nil {
+				logrus.WithError(err).Warn("Failed to unmarshal sampler rule")
```
vreynolds, line 317: Looks like these validation errors get logged upstream. Would it make sense to add context to the returned error instead of adding a log statement, so that logging is consolidated to the topmost place?

MikeGoldsmith: Yeah, good idea. Maybe something like this to wrap the error?

```golang
return fmt.Errorf("Failed to unmarshal sampler rule: %w", err)
```

I used this as [reference](https://blog.golang.org/go1.13-errors) on how to extend/wrap an `err`.

vreynolds: That looks great!

robbkidd (approved):
👍

PR #235, opened 2021-03-25
rules-complete.toml: add some more example rules for RulesBasedSampler.

isnotajoke:
The example sampler rules in `rules_complete.toml` cover some of the basic boolean operators, but miss things like `contains`, `starts-with`, etc that are pretty useful and powerful. It's easy enough to look at the code and figure out which rules actually exist (so I would totally understand if you don't want to go down the path of enumerating every rule type in the example config), but I would have found it helpful to have a more complete set of examples to work off of so I made this PR.

### DIFF
--- a/rules_complete.toml
+++ b/rules_complete.toml
@@ -220,12 +220,38 @@ SampleRate = 1
 		value = 1000.789
 
 	[[dataset4.rule]]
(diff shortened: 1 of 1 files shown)


isnotajoke:
Oops, looks like some tests are failing here. I'll reopen once I fix those.

PR #237, opened 2021-03-25
Add metrics qualifiers for structs that inject metrics

MikeGoldsmith:
When using dependency injection, all dependencies listed in structs must use the same name qualifier. The dependency builder does not auto inject based on property name match.

This PR sets the metrics inject qualifier for the main metrics config that is registered when executing main.go.

This was missed because tests manually create their dependencies and do not rely on dependency injection which I've now added for any struct that injects a metrics instance.


vreynolds (approved):
That's a sneaky one, I wish this surfaced at compile time, but I suppose that's the DI world we live in. ❤️ the tests

PR #247, opened 2021-04-12
Remove redundant peer/api suffix from response error metric

### LINKED ISSUE #244: API errors are incorrectly reported as peer errors
It seems refinery checks the API host option in outgoing events to see if the message is destined for a peer, or the honeycomb API:

https://github.com/honeycombio/refinery/blob/003f55f4c6a86fdc33f3740c9b0432523439a2fe/transmit/transmit.go#L164-L171

During a conversation in the pollinators slack we noticed that log lines emitted by refinery for some messages showed that the event's `api_host` option was empty, even though the event was being sent to honeycomb:

```
Apr 01 12:13:01 production-refinery-i-00c73eea958d0b46c refinery[11758]: time="2021-04-01T12:13:01Z" level=error msg="non-20x response when sending event" api_host= dataset= error="Post \"https://api.honeycomb.io/1/batch/production.traces\": context deadline exceeded (Client.Timeout exceeded while awaiting headers)" event_type= status_code=0 target=
```

This caused timeout errors in the upstream honeycomb API to be treated as peer errors:

Image 2021-04-06 at 10 36 30 am

This is a bit misleading as it causes the operator to believe there's a problem with their cluster, when the issue is really coming from the honeycomb API.

Given metrics are already prefixed with `upstream_` or `peer_`, I'm wondering whether we really need the `_api` and `_peer` suffixes, as it's a bit confusing. e.g. How should `upstream_response_errors_peer` differ from `peer_response_errors_peer`?

Would it be possible to remove this suffix so that operators can determine whether a metric is related to the honeycomb API, or the peer, by its prefix?

vreynolds:
* transmission metrics are already prefixed with peer/upstream
* fixes #244

### DIFF
--- a/transmit/transmit.go
+++ b/transmit/transmit.go
@@ -22,10 +22,9 @@ type Transmission interface {
 }
 
 const (
-	counterEnqueueErrors      = "enqueue_errors"
-	counterResponse20x        = "response_20x"
-	counterResponseErrorsAPI  = "response_errors_api"
-	counterResponseErrorsPeer = "response_errors_peer"
+	counterEnqueueErrors  = "enqueue_errors"
+	counterResponse20x    = "response_20x"
+	counterResponseErrors = "response_errors"
 )
 
 type DefaultTransmission struct {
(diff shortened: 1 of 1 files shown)


robbkidd (approved):
hey, you're redundant, spiderman

PR #248, opened 2021-05-03
Verify that sample rate trace field key is specified, if needed.

paulosman:
Currently, if `AddSampleRateKeyToTrace` is set to true in a sampler config, and no `AddSampleRateKeyToTraceField` is specified, refinery will attempt to send the sample rate in a field with an empty key. This will result in hard to debug 400s being returned from the Honeycomb Events API. It's probably best to just prevent this configuration scenario. This change makes refinery fail fast if `AddSampleRateKeyToTrace` is true and `AddSampleRateKeyToTraceField` is not set, or is empty.

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -255,6 +255,46 @@ func TestPeerManagementType(t *testing.T) {
 	}
 }
 
+func TestAbsentTraceKeyField(t *testing.T) {
+	tmpDir, err := ioutil.TempDir("", "")
+	assert.NoError(t, err)
+	defer os.RemoveAll(tmpDir)
+
+	configFile, err := ioutil.TempFile(tmpDir, "*.toml")
+	assert.NoError(t, err)
+
+	rulesFile, err := ioutil.TempFile(tmpDir, "*.toml")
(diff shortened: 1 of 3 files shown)


martin308:
Can we just use `required_with`? https://pkg.go.dev/github.com/go-playground/validator/v10#hdr-Required_With

paulosman:
> Can we just use `required_with`? https://pkg.go.dev/github.com/go-playground/validator/v10#hdr-Required_With

That is _much_ better. Let me retool this PR to use that.

paulosman:
> > Can we just use `required_with`? https://pkg.go.dev/github.com/go-playground/validator/v10#hdr-Required_With
> 
> That is _much_ better. Let me retool this PR to use that.

Done. @martin308 PTAL

martin308:
@paulosman looks good! I think maybe some tests need updating and would be good to have some new tests to cover this

martin308 (approved):
I would love to see a specific test for the failing case too!

vreynolds (approved):
🍿

PR #249, opened 2021-05-05
Move from garyburd Redigo to supported redigo

verajohne:
This PR moves from Garyburd Redigo to go module redigo. It also explicitly sets InsecureSkipVerify on the tls config as this was only set if tlsConfig was not set. See [redigo code](https://github.com/gomodule/redigo/blob/72af8129e040d6f962772a8c582e5e9f22085788/redis/conn.go#L246)

### DIFF
--- a/internal/peer/redis.go
+++ b/internal/peer/redis.go
@@ -179,11 +179,11 @@ func buildOptions(c config.Config) []redis.DialOption {
 	useTLS, _ := c.GetUseTLS()
 	if useTLS {
 		tlsConfig := &tls.Config{
-			MinVersion: tls.VersionTLS12,
+			MinVersion:         tls.VersionTLS12,
+			InsecureSkipVerify: true,
 		}
 		options = append(options,
 			redis.DialTLSConfig(tlsConfig),
-			redis.DialTLSSkipVerify(true),
 			redis.DialUseTLS(true))
 	}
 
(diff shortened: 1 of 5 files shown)


internal/peer/redis.go
```diff
@@ -179,11 +179,11 @@ func buildOptions(c config.Config) []redis.DialOption {
 	useTLS, _ := c.GetUseTLS()
 	if useTLS {
 		tlsConfig := &tls.Config{
-			MinVersion: tls.VersionTLS12,
+			MinVersion:         tls.VersionTLS12,
+			InsecureSkipVerify: true,
```
vreynolds, line 183:
```suggestion
```

Thanks for catching that skip verify was a noop! We came to the conclusion that it ought not to have been set in the first place, so let's just remove the noop line, and allow it to keep verifying as it is doing today.

vreynolds (changes requested):
Thank you! See comment about TLS skip verify, otherwise LGTM

internal/peer/redis.go
vreynolds: Ah, understood. We thought it was just something ya'll noticed while bumping the dependency. If you'd like, we can just merge the dependency change in this PR, and open a new issue for the skipVerify? On the issue, can you provide some context for your use case?

verajohne: @vreynolds Removed the code for InsecureSkipVerify from this PR. And will open an issue

vreynolds (approved):

PR #250, opened 2021-05-06
Add 'meta.refinery.local_hostname' to all spans

jharley:
The intention is to help determine issues with a single Refinery instance.

### DIFF
--- a/config_complete.toml
+++ b/config_complete.toml
@@ -87,6 +87,14 @@ PeerBufferSize = 10000
 # The debug service runs on the first open port between localhost:6060 and :6069 by default
 # DebugServiceAddr = "localhost:8085"
 
+# AddHostMetadataToTrace determines whether or not to add information about
+# the host that Refinery is running on to the spans that it processes.
+# If enabled, information about the host will be added to each span with the
+# prefix `meta.refinery.`.
+# Currently the only value added is 'meta.refinery.local_hostname'.
+# Eligible for live reload
+AddHostMetadataToTrace = true
+
 ############################
 ## Implementation Choices ##
 ############################
--- a/config_complete.toml
+++ b/config_complete.toml
@@ -92,8 +92,8 @@ PeerBufferSize = 10000
 # If enabled, information about the host will be added to each span with the
 # prefix `meta.refinery.`.
 # Currently the only value added is 'meta.refinery.local_hostname'.
-# Eligible for live reload
-AddHostMetadataToTrace = true
+# Not eligible for live reload
+AddHostMetadataToTrace = false
 
 ############################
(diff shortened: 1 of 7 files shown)


vreynolds:
Hurm. I tried it out locally, and I'm not seeing the extra metadata on the events. I think the `AddField` needs to happen before we grab the builder?

jharley:
Urgh, thank you and my apologies! :\

Adding the field before the builder is created (or on the builder directly) does give the desired result, but it breaks the testsuite as the field is being added all the event data.

If you feel this is useful -- I think it's a helpful addition -- perhaps it should be a config option anyway? Then it can be disabled for the testsuite to avoid having to add the field to all the expected test results. Perhaps something like `Config.GetEnableEventHostMetadata`?

vreynolds:
No worries! It wasn't intuitive to me, either.

I like the idea of adding a config option, because there's already something akin to it: `AddSampleRateKeyToTrace` in the rules config, except this new one (maybe something like `AddHostMetadataToTrace = true/false`) would be in the general config. That way, a single test that flips that to true can test the hostname, and all other tests should work the same.

Does that make sense?

jharley:
Makes good sense to me! I'll pick that up and run with it. 🚀

config_complete.toml
```diff
@@ -87,6 +87,14 @@ PeerBufferSize = 10000
 # The debug service runs on the first open port between localhost:6060 and :6069 by default
 # DebugServiceAddr = "localhost:8085"
 
+# AddHostMetadataToTrace determines whether or not to add information about
+# the host that Refinery is running on to the spans that it processes.
+# If enabled, information about the host will be added to each span with the
+# prefix `meta.refinery.`.
+# Currently the only value added is 'meta.refinery.local_hostname'.
+# Eligible for live reload
```
vreynolds, line 95: I don't believe it's live reloading, nor do we need to have that. Let's just update the doc comment that it's not eligible.

I'm inclined to default the value to false, since that's how the other metadata field works. WDYT?

vreynolds (changes requested):
Question about the default value, and small update about live reload, otherwise LGTM!

config_complete.toml
jharley: Good catch on the live reload comment: thank you!

Updated default value to false as well.

vreynolds (approved):

vreynolds:
@jharley could you update the docs as well? https://docs.honeycomb.io/manage-data-volume/refinery/configuration/#general-configuration

jharley:
> @jharley could you update the docs as well? https://docs.honeycomb.io/manage-data-volume/refinery/configuration/#general-configuration

No problem!

Issue #258, opened 2021-05-21
Span Events from OTLP are dropped


dstrelau:
It looks as if `span.Events` is never looked at here, meaning span events are just dropped on the floor 😢 

https://github.com/honeycombio/refinery/blob/main/route/route.go#L446-L478

PR #261, opened 2021-05-27
Propagate span events

dstrelau:
Implements #258 to pass through OTLP span events as regular events with `meta.annotation_type: "span_event"`. I don't see many tests for this codepath, but let me know if there's someplace I could easily add a test for this.

I can confirm that running this locally I'm able to see span events in a production dataset.


maplebed (commented):
LGTM.

paulosman (approved):

PR #264, opened 2021-06-03
Ensure span links and events generate events and get resource attrs

MikeGoldsmith:
When generating events for OTLP span events and links, each sub-event should also get resource attributes added.

NOTE:
- Fixed a couple of linter warnings in https://github.com/honeycombio/refinery/pull/264/commits/258cf313cc3776733fdd56efbba25ec6aba56c86
- Testing event content is challenging because if the number of dependencies required, collector, metrics, sampler. I've tested the events are created as expected but fixing testing framework is larger task

cc @bdarfler

### DIFF
--- a/route/route.go
+++ b/route/route.go
@@ -923,7 +923,7 @@ func getSampleRateFromAttributes(attrs map[string]interface{}) (int, error) {
 			sampleRate = int(v)
 		}
 	default:
-		err = fmt.Errorf("Unrecognised sampleRate datatype - %T", sampleRate)
+		err = fmt.Errorf("unrecognised sampleRate datatype - %T", sampleRate)
 		sampleRate = defaultSampleRate
 	}
 	// remove sampleRate from event fields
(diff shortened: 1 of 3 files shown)


route/route.go
```diff
@@ -501,22 +501,60 @@ func (r *Router) Export(ctx context.Context, req *collectortrace.ExportTraceServ
...
 						APIKey:     apiKey,
 						Dataset:    dataset,
 						SampleRate: uint(sampleRate),
 						Timestamp:  timestamp,
 						Data:       attrs,
+					})
+				}
+
+				for _, slink := range span.Links {
+					var timestamp time.Time
```
dstrelau, line 526: It doesn't look like this is getting set anywhere?

MikeGoldsmith: I copied this from Shepeherd. Span links don't have timestamps, so instead we should just omit the `timestamp` event field.

route/route_test.go
```diff
@@ -441,7 +446,97 @@ func TestOTLPHandler(t *testing.T) {
 		if err != nil {
 			t.Errorf(`Unexpected error: %s`, err)
 		}
+		assert.Equal(t, 2, len(mockTransmission.Events))
+		mockTransmission.Flush()
 	})
+
+	// TODO: (MG) figuure out how we can test JSON created from OTLP requests
```
bdarfler, line 453: Suggestion: use `t.skip` rather than commenting out

paulosman (approved):
LGTM

Issue #265, opened 2021-06-04
Feature: Support "does-not-contain" as operator


tr-fteixeira:
Hello, 

Seems like not all ui/api operators from the platform are supported on the RuleBasedSampler, [here](https://github.com/honeycombio/refinery/blob/main/sample/rules.go#L56). 

When using one of the unsupported operators, there is no indication of error, it just doesn't match the desired trace/span. 
It is also not documented anywhere (that i could find). 

Im looking specifically for that one, but would be nice to have them all =)

Regards,

MikeGoldsmith:
Closing this issue now #267 was merged. #266 will be use to track enhanced rules validation.

Issue #266, opened 2021-06-04
Enhancement: Validate operators supported on RulesBasedSampler when parsing the rules


tr-fteixeira:
This is related to https://github.com/honeycombio/refinery/issues/265 . 

It would have been a LOT easier to understand why the rules were not being applied/evaluated properly if we had the indication that those are not supported. 

This does not need to fail the parse, or could, this depends on the team preference for those failure modes.  Let me know what you guys would prefer 👍 

Regards,

irvingpop:
We're noticing a lot of customers are getting burned by Refinery silently ignoring rules that are malformed in any way.  It's rather difficult to troubleshoot.

irvingpop:
Couple of recent examples, from customers who used the bad dot.notation examples from the helm repo. Namely these two: 
* https://github.com/honeycombio/helm-charts/pull/169/files 
* https://github.com/honeycombio/helm-charts/pull/163/files

PR #267, opened 2021-06-05
Add support to "does-not-contain" operator on RulesBasedSampler

tr-fteixeira:
Solves https://github.com/honeycombio/refinery/issues/265

Took must of the code from the "contains", including the test. 
Let me know what you think, and if it needs anything else. 

Docs and extra logging for unsupported operators to be added on 266 👍.


MikeGoldsmith (approved):
Hey @tr-fteixeira - thanks for the contribution and sorry for the delay in getting back to you.

This looks good and I'm happy to accept it. I agree further work to better validate rules would be good and #266 is a good place to track that.

Issue #269, opened 2021-06-09
Using refinery and otel-collector to route traces based on content (add dataset to inmemory collector cache key)


tr-fteixeira:
Started the discussion on the pollinators slack, [here](https://honeycombpollinators.slack.com/archives/C01BZ3J7RMJ/p1622991532061900).

Im trying to use refinery as one of the tools to achieve "trace routing" or "trace multiplexing" capabilities, it is a unusual use case, but here is the context of the ask. 

this might be an unconventional use case, and seems like it runs into some problems [here](https://github.com/honeycombio/refinery/blob/main/collect/collect.go#L316), because it uses only traceID to group spans. 

Details :point_down:
**Context:**
Using istio and a shared ingress gateway for multiple applications/environments/teams (each mapping to its own dataset)
This means i have to split the destination of the traces based on content they carry (not doable in istio alone, or any app)..

<img width="1057" alt="Screen Shot 2021-06-06 at 5 58 09 PM" src="https://user-images.githubusercontent.com/55003893/121408085-fa00cd80-c92d-11eb-8e03-ab44eb68ee1d.png">

**What am i trying to do?**
Get those istio traces to the correct dataset, by means of:
Shared istio gateway -> otel-collector (fan out to 2 exporters, one for each dataset) -> refinery (rulesbasedSampler - drop traces of other environments/namespaces) -> HC
This effectively duplicates all traces generated by istio, and adds the meta of different datasets to each,
After that, the thought was to drop based on Refinery rules the ones going to the wrong dataset.

**What ends up happening?**
For a single exporter, it works great.. when multiple exporters are enabled, traces don’t get evaluated properly, some being kept/dropped, seemly in a random fashion.

**What i think the problem is:**
(No batching is enabled in otel)
When refinery collect the spans in memory and collate them to form a trace [here](https://github.com/honeycombio/refinery/blob/main/collect/collect.go#L316), only the TraceID is taken into account, so this is combining (sometimes) both exporters data, with different datasets as upstream into one single trace, making the sampling decision look wrong.

**To The actual questions**
Does this sound right? or did i get it all wrong :smiley:
Was this a conscious decision or i am trying to use the tool on an unexpected use case?

**Possible solutions**
- Adding `sp.Dataset`  to the cache object key as a prefix should fix it..
- using multiple refinery deployments for each exporter

Would prefer to do 1, but wanted to hear your thoughts before working on a PR :smiley:

paulosman:
[...]

PR #272, opened 2021-06-22
Adds dataset to inMemory cache key patterns

tr-fteixeira:
Solves https://github.com/honeycombio/refinery/issues/269

The issue has more details on the use case and the problems i've encountered. 

Not sure if this follows best practices in go, so feel free to review it harshly =) 
I've tested this on our stack, open to discuss specifics, reach out here or on pollinators 😄 

Did some thinking on how to add tests for this specific use case, but was not happy with any of the solutions as it transposes collect and rules "domains", open to inputs on how/if those are required.

### DIFF
--- a/collect/cache/cache.go
+++ b/collect/cache/cache.go
@@ -13,7 +13,7 @@ type Cache interface {
 	// Set adds the trace to the cache. If it is kicking out a trace from the cache
 	// that has not yet been sent, it will return that trace. Otherwise returns nil.
 	Set(trace *types.Trace) *types.Trace
-	Get(traceID string) *types.Trace
+	Get(Dataset string, traceID string) *types.Trace
 	// GetAll is used during shutdown to get all in-flight traces to flush them
 	GetAll() []*types.Trace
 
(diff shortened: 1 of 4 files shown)


collect/cache/cache.go
```diff
@@ -13,7 +13,7 @@ type Cache interface {
 	// Set adds the trace to the cache. If it is kicking out a trace from the cache
 	// that has not yet been sent, it will return that trace. Otherwise returns nil.
 	Set(trace *types.Trace) *types.Trace
-	Get(traceID string) *types.Trace
+	Get(Dataset string, traceID string) *types.Trace
```
tyler-boyd, line 16: I believe parameters should be lower camel cased

```suggestion
	Get(dataset string, traceID string) *types.Trace
```

paulosman:
Discussing in the associated issue:

https://github.com/honeycombio/refinery/issues/269#issuecomment-884501551

vreynolds:
https://github.com/honeycombio/refinery/issues/269#issuecomment-888653478

PR #317, opened 2021-09-15
Add dynamic sampler support to rules based samplers

### LINKED ISSUE #283: Feature Request: Support Rules based & Dynamics Samplers on single dataset
I want to be able to have rules based sampling then fall back to the dynamic sampler when not matched.

puckpuck:
The RulesBased sample is exclusive to the other samplers. This PR allows you to use Dynamic and EMADynamic samplers to determine the sample rate of a RulesBased sampler rule.

This adds a new `Downstream` option to a rule configuration. This option can be used to contain either a `DynamicSampler` or `EMADynamicSampler` option that is a complete configuration of their respective samplers.  When defined the downstream sampler will be used to determine the sample rate when all rule conditions are matched.

A popular use case is to use RulesBased sampling to keep all error and long duration traces while using a dynamic sampler for all other data, based on key fields.

Closes #283 

notes: 

- I'm not sure the word `Downstream` is the right one here. Naming things is hard.
- I would also prefer to have just the initial option contain the entire downstream sampler config, without having to wrap it into another config option.  However, because of Go's strong typing, this doesn't seem possible without significant re-architecture of how the config is parsed for all samplers.

### DIFF
--- a/config/sampler_config.go
+++ b/config/sampler_config.go
@@ -2,7 +2,6 @@ package config
 
 import (
 	"fmt"
-	"github.com/honeycombio/refinery/sample"
 )
 
 type DeterministicSamplerConfig struct {
@@ -58,12 +57,11 @@ type RulesBasedDownstreamSampler struct {
 }
 
 type RulesBasedSamplerRule struct {
-	Name              string
-	SampleRate        int
-	Downstream        *RulesBasedDownstreamSampler
-	DownstreamSampler sample.Sampler
-	Drop              bool
-	Condition         []*RulesBasedSamplerCondition
+	Name       string
+	SampleRate int
+	Downstream *RulesBasedDownstreamSampler
+	Drop       bool
+	Condition  []*RulesBasedSamplerCondition
 }
 
 func (r *RulesBasedSamplerRule) String() string {
--- a/config/sampler_config.go
+++ b/config/sampler_config.go
@@ -1,5 +1,10 @@
 package config
 
+import (
+	"fmt"
+	"github.com/honeycombio/refinery/sample"
+)
+
 type DeterministicSamplerConfig struct {
 	SampleRate int `validate:"required,gte=1"`
 }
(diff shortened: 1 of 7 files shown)


config/sampler_config.go
```diff
@@ -36,3 +40,38 @@ type TotalThroughputSamplerConfig struct {
...
+
+type RulesBasedDownstreamSampler struct {
+	DynamicSampler    *DynamicSamplerConfig
+	EMADynamicSampler *EMADynamicSamplerConfig
+}
+
+type RulesBasedSamplerRule struct {
+	Name       string
+	SampleRate int
+	Downstream *RulesBasedDownstreamSampler
```
MikeGoldsmith, line 62: I agree `Downstream` doesn't feel like the right word. Maybe we can use `Sampler`, so we'd have both a `rule.Sampler` and `rule.SampleRate`.

sample/rules.go
```diff
@@ -127,7 +115,32 @@ func (s *RulesBasedSampler) GetSampleRate(trace *types.Trace) (rate uint, keep b
...
-		if matched == len(rule.Condition) {
+		if rule.Condition == nil || matched == len(rule.Condition) {
+			var rate uint
+			var keep bool
+
+			if rule.Downstream != nil {
+
+				var sampler Sampler
+				var found bool
+				if sampler, found = s.samplers[rule.String()]; !found {
```
MikeGoldsmith, line 126: The mutex needs to be used to ensure thread safety when reading/writing with the map to prevent multiple threads reading an empty sampler and trying to create one then each trying to store it in the map. The creation of a sampler doesn't really need to be synchronised as it doesn't consume any external resources that need to be disposed of manually.

Instead, we could populate the samplers map during `Start`. This will separate the creation from usage and not needing to lazy load the samplers. We can do the error logs there too if we can't create a sampler.

sample/rules.go
```diff
@@ -138,15 +151,39 @@ func (s *RulesBasedSampler) GetSampleRate(trace *types.Trace) (rate uint, keep b
 				"rate":      rate,
 				"keep":      keep,
 				"drop_rule": rule.Drop,
-				"rule_name": rule.Name,
 			}).Logf("got sample rate and decision")
+
```
MikeGoldsmith, line 155: note: remove whitespace
```suggestion
```

MikeGoldsmith (changes requested):
This looks really nice and is a great start.

It's a shame we can't use `sample.Sampler` as the type in `RulesBasedSamplerRule` because it introduces a circular dependency. Maybe that's something we can look at resolving.

I'd like to see the sampler creation moved to `Start` and a couple of other suggestions.

sample/rules.go
puckpuck: Moving all initialization to Start means we can just check if the sampler exists, and if not log an error and return `1, true`, negating the need to use a mutex (unless I'm missing something)

MikeGoldsmith: Yeah, that's right. We wouldn't need the mutex at all. If the sampler doesn't exist, just return `1, true` 👍🏻 

If we're going to log for each failed attempt to get a sampler, it'll need to be debug.

puckpuck:
I added `Start()` to the `sample.Sampler` interface. All samplers called this anyhow, and it allowed to simplify the code when moving all initialization to the rule start.

I also added support for the TotalThroughput sampler, based on a comment thread in Slack.

MikeGoldsmith (approved):
Looks great, thanks @puckpuck 👍🏻

PR #328, opened 2021-10-04
Only login to docker if going to publish

### LINKED ISSUE #325: testing docker image build+publish fails from forks
We started [running the docker image build and publish for all builds](https://github.com/honeycombio/refinery/pull/323/commits/215afe64353b5fc086bdd9d5dd0359b129d463f5) to test image building. But the implementation there requires CI secrets which forks do not have causing spurious failures for pull requests from forks. Consider a separate job for local image build and publish, requiring no secrets.

An example from a dependabot PR:

Screen Shot 2021-10-01 at 1 32 22 PM

MikeGoldsmith:
Forked builds will fail the `publish_docker` step because it tries to login to docker, but the username and password are in CircleCI secrets that are not shared on forked builds.

- Resolves #325 

- only login to docker if `CIRCLE_TAG` env var is set


vreynolds (approved):
as a future optimization: we can cache `dockerize` and `ko` if we download the executables and cache those in a temp directory ala buildevents: https://github.com/honeycombio/buildevents-orb/blob/main/orb.yml

PR #338, opened 2021-10-13
Re-add missing docker login when publishing

MikeGoldsmith:
We need to provide login credentials for publishing to DockerHub, these were accidentally removed while refactoring.

- re-add docker login command when publishing

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -145,7 +145,7 @@ jobs:
           environment:
             KO_DOCKER_REPO: honeycombio
           command: |
-            echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -144,7 +144,9 @@ jobs:
           name: build docker images and publish to Docker Hub
           environment:
             KO_DOCKER_REPO: honeycombio
-          command: ./build-docker.sh
+          command: |
+            echo "$DOCKERHUB_PASS" | docker login -u "$DOCKERHUB_USERNAME" --password-stdin
```
JamieDanielson, line 148:
```suggestion
            echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
```

JamieDanielson: We removed the old env vars, these are the new ones to use in the Honeycomb Secrets context

PR #341, opened 2021-10-29
Replace internal duplicated code with Husky

MikeGoldsmith:
We have duplicated a number of OTLP trace utility functions into refinery from other systems. This PR replaces those duplicated functions with a new utility module [husky](https://github.com/honeycombio/husky).

- Remove internal opentelemetry-proto files
- Add Husky v0.2.0
- Use husky/otlp and remove duplicated functionality


robbkidd (approved):
happy husky

PR #342, opened 2021-11-01
Add an --interface-names flag

ismith:
Our config lets you specify which network name to use for peers; this
flag makes it easier to discover what interfaces the binary has
available.

Example, on a locally-built image:
```
± docker run  -t ko.local/refinery:e2915c3075fd17b54c155f53b04acf0cff1a027a5610cc7bf25052d0725bd7ed --interface-names
lo
eth0
```

### DIFF
--- a/cmd/refinery/main.go
+++ b/cmd/refinery/main.go
@@ -33,10 +33,11 @@ var BuildID string
 var version string
 
 type Options struct {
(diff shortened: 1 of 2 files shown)


ismith:
These tests - run with `make test`, so including the `-race` flag - pass locally for me, but seem to fail in Circle. With different errors on all 3 runs so far ...

I suspect this is unrelated to my change.

MikeGoldsmith (approved):
Looks good 👍🏻 

Thanks for updating the flaky test timings too.

PR #343, opened 2021-11-01
Use alpine as base image

ismith:
Alpine is not _that_ much bigger than google's distroless image. and
having this OS layer underneath will make it easier to debug.

(Currently: do we have a problem with the network interface name in our
config?)

### DIFF
--- a/.ko.yaml
+++ b/.ko.yaml
@@ -1,0 +1,1 @@
+defaultBaseImage: alpine:3.13
--- a/collect/collect_test.go
+++ b/collect/collect_test.go
@@ -320,14 +320,14 @@ func TestCacheSizeReload(t *testing.T) {
 	coll.AddSpan(&types.Span{TraceID: "2", Event: event})
 
 	expectedEvents := 1
-	wait := 2 * time.Millisecond
+	wait := 1 * time.Second
 	check := func() bool {
 		transmission.Mux.RLock()
 		defer transmission.Mux.RUnlock()
 
 		return len(transmission.Events) == expectedEvents
 	}
-	assert.Eventually(t, check, 10*wait, wait, "expected one trace evicted and sent")
+	assert.Eventually(t, check, 60*wait, wait, "expected one trace evicted and sent")
 
 	conf.Mux.Lock()
 	conf.GetInMemoryCollectorCacheCapacityVal.CacheCapacity = 2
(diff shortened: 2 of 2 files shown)


ismith:
Test: run `./build-docker.sh`, output contains:
`2021/11/01 16:34:23 Using base alpine:3.13 for github.com/honeycombio/refinery/cmd/refinery`

Vlaaaaaaad:
I am very much in favour of this (it'll also make it [easier to run Refinery using ECS on Fargate](https://github.com/Vlaaaaaaad/terraform-aws-fargate-refinery/issues/2), related [Slack thread here](https://honeycombpollinators.slack.com/archives/CKWS8M71C/p1597937188007400)), but any chance something other than Alpine can be used?

Alpine is [somewhat](https://github.com/gliderlabs/docker-alpine/blob/master/docs/caveats.md#dns) well [known](https://github.com/gliderlabs/docker-alpine/issues/255) to have [runtime issues](https://stackoverflow.com/questions/65181012/does-alpine-have-known-dns-issue-within-kubernetes) due to `musl`. I had massive DNS issues in Kubernetes due to this, so I am now hyper-aware, even if some bugs were fixed.

Compared to [Alpine's 2MB size](https://hub.docker.com/layers/alpine/library/alpine/latest/images/sha256-e841d0f0881ea22080e84088337646ada15871abbc3ce19b3219e8fc2cb0cc22?context=explore), alternatives are indeed ten times bigger with [Ubuntu images being ~27MB](https://hub.docker.com/layers/ubuntu/library/ubuntu/latest/images/sha256-7cc0576c7c0ec2384de5cbf245f41567e922aab1b075f3e8ad565f508032df17?context=explore) and [Debian slim images being ~22MB](https://hub.docker.com/layers/debian/library/debian/stretch-slim/images/sha256-3e6f082b2511516bb3f7f216d9611b837e1ce9891ef145c85cd5570b60170ed6?context=explore), but I think the added size has minimal impact.

MikeGoldsmith (approved):
I'm good with changing the default image to alpine.

@ismith do you have an opinion on @Vlaaaaaaad's recommendation regarding Ubuntu/Debian images instead?

MikeGoldsmith:
@Vlaaaaaaad I'm going to merge this PR using alpine as base image. Please raise an issue if you feel you'd still like use to consider us using another base image.

PR #347, opened 2021-11-04
Bump libhoney-go to v1.15.6

MikeGoldsmith:
## Which problem is this PR solving?
Bumps libhoney-go to the latest v1.15.6

## Short description of the changes
- update libhoney-go to v1.15.6 in go.mod/sum

PR #348, opened 2021-11-04
Prepares v1.6.0 release

MikeGoldsmith:
Prepares v1.6.0 release

- adds changelog entry


robbkidd (approved):
I assume this depends on #347 merging first.

Let's do this!

PR #355, opened 2021-11-23
Improves histogram buckets over the default set

### LINKED ISSUE #115: Bucket sizes for Prometheus histograms
The [default bucket sizes](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus?tab=doc#pkg-variables) for [Prometheus histograms](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus?tab=doc#HistogramOpts) are too small to usefully represent metrics like `trace_duration_ms` or `trace_span_count`. I imagine that one size will never suit all and that all of the histograms need to be reviewed independently for appropriate values.

### LINKED ISSUE #346: Use gauges not histograms

Prometheus Histograms are confusing. User's don't understand them. A quick google search turns up articles like "Hisograms with Prometheus: A Tale of Woe" and more importantly, we have confused customers that reach out to us in [Pollenators](https://honeycombpollinators.slack.com/archives/C01BZ3J7RMJ/p1635509296050900) about them as well as customers that open up [issues](https://github.com/honeycombio/refinery/issues/339#issuecomment-942646548) based in part on misinterpretations of them.

Additionally, we have failed to understand them in the past which resulted in #115. Trying to fix our very broken Prometheus histograms is non-trivial because Refinery's internal metrics interface doesn't allow different settings per metric and even if it did it's unclear we could find a good value that works across the variety of traffic and host configurations that our customers run Refinery in.

Rather than continue to leverage Prometheus histograms which are confusing to our customers and difficult to properly configure we should remove histograms altogether and use gauges instead. 

Currently, the Honeycomb metrics reporter reports every 1s which seems granular enough that switching to a gauge wouldn't impact operability. We would want to recommend that Prometheus users scrape between every 1 and 10 seconds to ensure they get the right level of granularity.

bdarfler:
This is an attempt at a usable set of buckets for a wide range of metrics

Fixes #115

As noted in the issue, the default buckets in the go client are as follows:

```
DefBuckets = []float64{.005, .01, .025, .05, .1, .25, .5, 1, 2.5, 5, 10}
```

… and from the docs … 

> DefBuckets are the default Histogram buckets. The default buckets are tailored to broadly measure the response time (in seconds) of a network service. Most likely, however, you will be required to define buckets customized to your use case.

Additionally from the docs … 

> Buckets defines the buckets into which observations are counted. Each element in the slice is the upper inclusive bound of a bucket. The values must be sorted in strictly increasing order. There is no need to add a highest bucket with +Inf bound, it will be added implicitly. The default value is DefBuckets.

In the end, the `cache_entries` metric therefore always ends up in the implicit +inf bucket and makes the metric nearly useless. You just know your `cache_entries` are more than 10.

For example "your stuff is more than 10!":

graphs that show that these only go to 10

### DIFF
--- a/metrics/prometheus.go
+++ b/metrics/prometheus.go
@@ -72,6 +72,9 @@ func (p *PromMetrics) Register(name string, metricType string) {
 			Name:      name,
 			Namespace: p.prefix,
 			Help:      name,
+			// This is an attempt at a usable set of buckets for a wide range of metrics
+			// 16 buckets, first upper bound of 1, each following upper bound is 4x the previous
+			Buckets: prometheus.ExponentialBuckets(1, 4, 16),
 		})
 	}
 


akasprzok:
I'm on board with this!

robbkidd (approved):
These buckets appear to be more better.

a bucket

PR #361, opened 2021-12-02
Bump husky to v0.3.0

MikeGoldsmith:
Updates the Husky to v0.3.0 and removes duplicate code Husky now provides.

- Bump husky to v0.3.0
- Update otlp_trace to use new interface
- Remove getSampleRateFromAttributes func and tests


robbkidd (approved):
👍🏻 

I think it's safe to merge this even though we have a pending decision around which version of otel-proto-go to depend on in husky as a result of the transitive dep on grpc brining in surprising changes in other husky-using projects. Refinery can benefit from this change on main and get an updated husky (with back-versioned proto-go and grpc) as a followup.

a husky poking its head in

MikeGoldsmith:
I'll hold on this and bump to v0.4.0 instead.

PR #364, opened 2021-12-06
EIAE - ramp down upstream batch size 500 -> 100

jharley:

- experimental change for upstream communications

- swap hard-coded upstream max. batch size from 500 to 100


MikeGoldsmith:
I think #365 supersedes this as it makes the batch size configurable.

jharley:
@MikeGoldsmith 💯 Closing this one out

PR #382, opened 2022-01-25
Add retries when connecting to redis during init

MikeGoldsmith:
The refinery process only attempts to connect to redis ones during initialization. This can be challenging when redis is started at the same time and causes refinery to error out.

This change allows refinery to keep attempting to connect to redis for up to 10 seconds, with a one second delay between attempts.

- Add timeout of 10 seconds when calling redis.dial with a 1 second delay between attempts


vreynolds:
Can we keep reporting the underlying connection error? I think it would be useful to have more information when the connection fails.

vreynolds (approved):
🎇

kentquirk (approved):
Looks good. one extra semicolon.

Issue #275, opened 2021-06-29
Configuration through environment variables


thundering-herd:
Hej,

**tldr; make all configuration properties available via environment variables to provide the means to dynamically adjust `Refinery`**

we are introducing `Refinery` into our setup and stumbled upon the problem that the environment variables are not evaluated within the configuration file.

For example a configuration like following

```
 18 [PeerManagement]
 19 Type = "redis"
 20 RedisHost =  "refinery-leader-redis-${ENVIRONMENT}.mydomain:6379"
```

will not work. (I don't know if this is the behaviour of `viper`).

Looking at the implementation of `Refinery` we can see that only four environment variables are exposed for usage https://github.com/honeycombio/refinery/blob/a8684973018395e11c56e2559535333038660a5d/config/file_config.go#L125

Taking the 12factor principle into account, I can't think about a design decision why not all properties are exposed through environment variables. It seems to be possible with `viper` by defining a `SetEnvPrefix` and running `AutomaticEnv`.

A use case in our scenario is following:

- we are running different environments (e.g. dev, stage, prod)
- each of them uses a dedicated `Refinery` instance (plus `redis`)
- we want to send the metrics of `Refinery` to our analyse backend **but** into different datasets
--> at this moment it is not possible to do this without baking in the different dataset names

Thanks for any input and maybe I am missing something simple here :)

olane:
I'm running into the same issue - I've ended up creating a wrapper Dockerfile which injects the `config.toml` file with various environment variables that we want configurable (or secret) for certain environments. 

It works, but it's still annoying - because it's baked in at the image level we're going to have to build a separate image for each environment. Being able to just pass environment variables in directly would vastly simplify things.

olane:
Ah - so for my use case I didn't realize that there actually do seem to be env variables for the things I care about right now, it's just that the `MetricsAPIKey` value isn't documented as having an env var alternative on https://docs.honeycomb.io/manage-data-volume/refinery/configuration/ 

I'd still be in favour of having as many values as possible have environment variable alternatives, though.

cartermp:
Since Refinery 2.0, more things were added that you can configure via environment variables. So we're thinking of closing this one out since it seems that most things people need are handled now.

PR #386, opened 2022-01-27
Honor environment variable to set gRPC listener address

### LINKED ISSUE #383: Introduce environment variable for Refinery's gRPC listener's bind address

When choosing whether to serve Refinery behind a proxy terminating TLS, sometimes we want to bind Refinery's gRPC listener to all network interfaces (specified via the IP address _0.0.0.0_) in order to allow unencrypted calls from clients anywhere, but sometimes we want to bind this listener to only the loopback interface (via address _127.0.0.1_).

At present, we can only configure the gRPC listener's bind address via the Refinery configuration file. When operating with or without TLS in front of Refinery, this is the only detail that has to change in the configuration file. Much like with [the "REFINERY_REDIS_HOST" environment variable](https://github.com/honeycombio/refinery/blob/5d57c7468ded2bbb5c4b95868b9d2a6befb22539/config/file_config.go#L97), we would like to introduce an environment variable to set the gRPC listener's bind address, setting [the `configContents.GRPCListenAddr` field](https://github.com/honeycombio/refinery/blob/423f235c88de424acc34e2a1fc822d878254973c/config/file_config.go#L31) from afar. Doing so would allow us to use a common configuration file, and vary only this environment variable in our Kubernetes pod spec for Refinery.

Introduce a new environment variable consumed by Refinery, named something like "REFINERY_GPRC_LISTEN_ADDR" for parity with the similar environment variable names. As with other Viper-managed configuration, [setting this environment variable would take precedence over a value set in the configuration file](https://github.com/spf13/viper#why-viper).

I expect the implementation would look something like the following in the `config.NewConfig` function:
```go
c.BindEnv("GRPCListenAddr", "REFINERY_GPRC_LISTEN_ADDR")
```

We can generate different configuration files for each of these two scenarios, but doing so introduces more complexity into our Kubernetes manifest preparation system, and we wind up with two nearly identical configurations in play, without hoisting out the lone detail that varies to make the intended similarity and difference clear.

**Additional context**

Viper also offers [its `AutomaticEnv` facility](https://github.com/spf13/viper/#working-with-environment-variables), opening **all** the configuration fields to access via environment variables, albeit with some clumsy names. [I mentioned this possibility in the "discuss-refinery" channel](https://honeycombpollinators.slack.com/archives/C01BZ3J7RMJ/p1643061358024000) of the "Honeycomb Pollinators" Slack workspace.

seh:
Introduce the "REFINERY_GRPC_LISTEN_ADDRESS" environment variable to allow setting Refinery's `GRPCListenAddr` configuration field, overriding any such configuration read from a file.

Fixes #383.

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -12,10 +12,28 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+func TestGRPCListenAddrEnvVar(t *testing.T) {
+	const address = "127.0.0.1:4317"
+	const envVarName = "REFINERY_GRPC_LISTEN_ADDRESS"
+	os.Setenv(envVarName, address)
+	defer os.Unsetenv(envVarName)
+
+	c, err := NewConfig("../config.toml", "../rules.toml", func(err error) {})
+
+	if err != nil {
+		t.Error(err)
+	}
+
+	if a, _ := c.GetGRPCListenAddr(); a != address {
+		t.Error("received", a, "expected", address)
+	}
+}
+
 func TestRedisHostEnvVar(t *testing.T) {
-	host := "redis.magic:1337"
-	os.Setenv("REFINERY_REDIS_HOST", host)
-	defer os.Unsetenv("REFINERY_REDIS_HOST")
+	const host = "redis.magic:1337"
+	const envVarName = "REFINERY_REDIS_HOST"
+	os.Setenv(envVarName, host)
+	defer os.Unsetenv(envVarName)
 
 	c, err := NewConfig("../config.toml", "../rules.toml", func(err error) {})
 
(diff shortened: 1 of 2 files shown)


config/config_test.go
```diff
@@ -1,4 +1,4 @@
-// +build all race
+//go:build all || race
```
MikeGoldsmith, line 1: I'd prefer to update all go build tags in the project at the same time. We can do that in a separate PR 👍🏻

MikeGoldsmith (approved):
Thanks for the PR @seh - it looks great. Just one small ask on reverting the build tags.

It helps address some of #275 too.

config/config_test.go
seh: Please see my most recent commit f65b90c64a790eafb64ab2fe47f6742abbe7b164 in which I reverted that change. _go fmt_ isn't happy about it, but it will come around.

MikeGoldsmith: Thanks - I agree we need to move the newer standard and will do the change package wide separately in another PR 👍🏻

PR #387, opened 2022-01-31
Properly set meta.refinery.local_hostname field

### LINKED ISSUE #385: The meta.refinery.local_hostname field, is not set properly
**Versions**

- Go:
- Refinery: 1.8.1

**Steps to reproduce**

1. Setup a multi-node cluster
2. Set the `AddHostMetadataToTrace` config property to true
3. query filtered on a single trace, grouped by `meta.refinery.local_hostname` and notice a trace will have multiple hosts used when it should only be 1 

**Additional context**

After further digging into the code, this is set at the Libhoney client level, which is used for both peer communication and to send to honeycomb. It seems the peer that receives the span initially is what the hostname will be set to.

jharley:
Ensure the hostname is set before transmitting upstream to the current hostname.

- Closes #385 

- As suggested by @puckpuck in #385, the host metadata is now added just before the trace is sent upstream

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,5 @@ refinery
 !/cmd/refinery
 test_redimem
 !/cmd/test_redimem
+
+dockerize*
--- a/collect/collect.go
+++ b/collect/collect.go
@@ -467,6 +467,12 @@ func (i *InMemCollector) send(trace *types.Trace) {
 			field := i.Config.GetDryRunFieldName()
 			sp.Data[field] = shouldSend
 		}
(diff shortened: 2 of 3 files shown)


puckpuck:
Since this will get called with each outgoing span, instead of just once at startup, does it make sense to cache the hostname and reuse it instead of calling `os.Hostname()` for each span?

MikeGoldsmith (approved):
Thanks for working on this @jharley.

I made a small adjustment to get the hostname during the collector Start func to be reused across all traces.

PR #395, opened 2022-02-03
build: add ARM64 RPM artifact

jharley:

Missing ARM64 RPM artifact (Amazon Linux, RedHat, Rocky, etc)

Adds an ARM64 RPM to the released artifacts.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -83,6 +83,9 @@ jobs:
           name: build_rpm_amd64
           command: ./build-pkg.sh -m amd64 -v "${CIRCLE_TAG:-${CIRCLE_SHA1:0:7}}" -t rpm && mv *.rpm ~/artifacts
       - run:
+          name: build_rpm_arm64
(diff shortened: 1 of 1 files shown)


MikeGoldsmith (approved):
Thanks @jharley 🎉

MikeGoldsmith:
Looking at the CircleCI build, the output file is named `refinery-ea71155-1.aarch64.rpm`. Would it be more appropriate to rename to `refinery-ea71155-1.arm64.rpm`?

jharley:
> Looking at the [CircleCI build](https://app.circleci.com/pipelines/github/honeycombio/refinery/1025/workflows/45234fe7-335b-429b-b919-4829cfd5fca2/jobs/2867/parallel-runs/0/steps/0-112), the output file is named `refinery-ea71155-1.aarch64.rpm`. Would it be more appropriate to rename to `refinery-ea71155-1.arm64.rpm`?

Looking at the [Rocky Linux packages](https://download.rockylinux.org/pub/rocky/8/BaseOS/aarch64/os/Packages/l/) I think `aarch64` is correct

PR #402, opened 2022-02-12
fix: race condition in config rebuilds

vreynolds:

-

-


kentquirk:
We just ran into this again, and the consensus online is that there's no good fix. So we still have a !race in the build tags for this test. So it goes.

PR #403, opened 2022-02-15
E&S part 2

MikeGoldsmith:

-

-

### DIFF
--- a/collect/collect.go
+++ b/collect/collect.go
@@ -437,10 +437,10 @@ func (i *InMemCollector) send(trace *types.Trace) {
 	var sampler sample.Sampler
 	var found bool
 
-	if sampler, found = i.datasetSamplers[trace.Dataset]; !found {
-		sampler = i.SamplerFactory.GetSamplerImplementationForDataset(trace.Dataset)
-		// save sampler for later
-		i.datasetSamplers[trace.Dataset] = sampler
+	samplerKey := i.getSamplerKey(trace)
+	if sampler, found = i.datasetSamplers[samplerKey]; !found {
+		sampler = i.SamplerFactory.GetSamplerImplementationForDataset(samplerKey)
+		i.datasetSamplers[samplerKey] = sampler
 	}
 
 	// make sampling decision and update the trace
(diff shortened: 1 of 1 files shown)


route/otlp_trace.go
```diff
@@ -65,18 +64,27 @@ func processTraceRequest(
...
-			APIHost:    apiHost,
-			APIKey:     apiKey,
-			Dataset:    datasetName,
-			SampleRate: uint(ev.SampleRate),
-			Timestamp:  ev.Timestamp,
-			Data:       ev.Attributes,
-		}
-		if err = router.processEvent(event, requestID); err != nil {
-			router.Logger.Error().Logf("Error processing event: " + err.Error())
+	// get environment name - will be empty for legacy keys
```
vreynolds, line 67: it will be `$legacy$` for legacy keys, right? does that change anything downstream?

route/route.go
```diff
@@ -626,3 +646,119 @@ func getFirstValueFromMetadata(key string, md metadata.MD) string {
...
+	return ""
+}
+
+func (c *environmentCache) getOrSet(key string, ttl time.Duration, getFn func(string) (string, error)) (string, error) {
+	val := c.get(key)
+	if val != "" {
+		return val, nil
+	}
+
+	c.mutex.Lock()
```
vreynolds, line 685: we could probably move the write lock down to `addItem` function - it's the lowest level that writes, and we might call it from elsewhere

types/event.go
```diff
@@ -64,11 +65,29 @@ func (t *Trace) AddSpan(sp *Span) {
...
+
+func (t *Trace) GetSamplerKey() (string, bool) {
+	if IsLegacyAPIKey(t.APIKey) {
+		return t.Dataset, true
+	}
 
+	env := ""
+	for _, sp := range t.GetSpans() {
+		if sp.Event.Environment != "" {
+			env = sp.Event.Environment
```
vreynolds, line 78: assuming all events have the same environment, we can just break out of the loop after we find the first one?

vreynolds:
looks good, just some optimization suggestions. I checked it out locally and did a cursory test with legacy key + dataset-based rules, and new key + environment-based rules

route/route.go
MikeGoldsmith: I was aiming to let the write block happen on concurrent requests so we don't do parallel calls to auth endpoint, but on reflection it just makes them serial instead which isn't any better.

Instead, I'll do another cache check after acquiring the lock and if still no result then do the auth check. Then if multiple calls see a cache miss, only the first one to get the write lock will do the auth check and populate the cache. The following write locks will then get a cache hit and skip the auth check.

kentquirk: I ended up following exactly this pattern in Basenji, so I agree with this decision.

route/route.go
```diff
@@ -626,3 +646,119 @@ func getFirstValueFromMetadata(key string, md metadata.MD) string {
 	}
 	return ""
 }
+
+type environmentCache struct {
```
kentquirk, line 650: It probably doesn't matter because this is only called in one place, but if this were to be used in multiple places, I'd rather see the getFn passed in to the environmentCache on construction, and then called with the key when the value is not found or the ttl expires. Then the cache basically looks like a key/value map. 

However, the way you've done it here is not wrong and I'm not requesting you to change it.

kentquirk (approved):
My review was mostly limited to looking over the cache stuff.

route/route.go
MikeGoldsmith: Thanks for the feedback - I like your suggestion as it simplifies the usage fo the cache so dod the update.

route/route.go
MikeGoldsmith: Thanks both 👍🏻

vreynolds (approved):

build-pkg.sh
```diff
@@ -30,6 +30,8 @@ if [ -z "$version" ]; then
     version=v0.0.0-dev
 fi
 
+version="v1.11.0-dev"
```
MikeGoldsmith, line 33: Need to revert this

MikeGoldsmith (commented):
Need to do a couple of tidy up things before merging

MikeGoldsmith:
Created follow-up tickets:
- https://github.com/honeycombio/refinery/issues/406
- https://github.com/honeycombio/refinery/issues/407
- https://github.com/honeycombio/refinery/issues/408

Issue #413, opened 2022-02-25
allow nested JSON fields for rules sampling config


ecobrien29:
**Is your feature request related to a problem? Please describe.**
Yes, our logs have nested JSON fields that we would like to use for sampling decisions. The current code looks like it does not allow this functionality. 

**Describe the solution you'd like**
I would like to modify the code [here](https://github.com/honeycombio/refinery/blob/main/sample/rules.go#L72) to also use a [get function](https://lodash.com/docs/4.17.15#get) if the value isn't found in order to have it read nested JSON, but also retain the original functionality for fields with a name including a '.'

**Describe alternatives you've considered**

**Additional context**

MikeGoldsmith:
Thanks for raising this @ecobrien29. We will take a look and add to our backlog for prioritisation.

We're also very open to community contributions, so if you feel so inclined we'd welcome a PR 👍🏻

PR #418, opened 2022-03-07
Add parsing for nested json fields in the rules sampler

ecobrien29:
Adds support for checking nested JSON when validating rules. The option can be enabled using the `CheckNestedFields` configuration option.

- Closes https://github.com/honeycombio/refinery/issues/413

- Adds a [json parser module](https://github.com/tidwall/gjson) to allow retrieval of nested fields for the rules sampler. If the field is not found in the top level of the span, it calls the get function to check if it exists in the nested json.

### DIFF
--- a/sample/rules.go
+++ b/sample/rules.go
@@ -74,7 +74,7 @@ func (s *RulesBasedSampler) GetSampleRate(trace *types.Trace) (rate uint, keep b
 				value, exists := span.Data[condition.Field]
 				if !exists {
 					jsonStr, err := json.Marshal(span.Data)
-					if err != nil {
+					if err == nil {
 						result := gjson.Get(string(jsonStr), condition.Field)
 						if result.Exists() {
 							value = result.String()
--- a/sample/rules.go
+++ b/sample/rules.go
@@ -72,7 +72,7 @@ func (s *RulesBasedSampler) GetSampleRate(trace *types.Trace) (rate uint, keep b
 			for _, span := range trace.GetSpans() {
 				var match bool
 				value, exists := span.Data[condition.Field]
-				if !exists {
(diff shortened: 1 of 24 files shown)


sample/rules.go
```diff
@@ -70,6 +72,16 @@ func (s *RulesBasedSampler) GetSampleRate(trace *types.Trace) (rate uint, keep b
 			for _, span := range trace.GetSpans() {
 				var match bool
 				value, exists := span.Data[condition.Field]
+				if !exists {
```
MikeGoldsmith, line 75: I'm concerned this would impact users that does not want to pay the cost of attempting to JSON parse span data when trying to resolve rule conditions.

I think we should look into adding an opt-in configuration option to enable this behaviour. Default would be disabled.

MikeGoldsmith (changes requested):
I like the overall solution but I'm concerned with enabling JSON parsing of span data for all condition failures. This should be configuration driven to enable his behaviour.

sample/rules.go
ecobrien29: That makes sense. I can add an optional parameter in the rules config file to enable it (similar to the AddSampleRateKeyToTrace and UseTraceLength fields), unless you think there is a better place for it.

ecobrien29:
I made the requested changes and tested with the refinery I have running with the nested fields parameter both enabled and disabled. The nested traces were correctly sampled when enabled and didn't affect normal sampling rules.

vreynolds:
Thanks @ecobrien29! We've had a busy couple weeks, but we'll take another look as soon as we can!

MikeGoldsmith (approved):
Looks good, thank you for the follow-up and sorry for the delay in getting back to you.

PR #419, opened 2022-03-08
Update to YAML configs

puckpuck:
The popular deploy mechanism for refinery is in Kubernetes using Helm where all configuration is done in YAML format. The current configuration examples and docs on done in TOML format which leads to confusion with operators when deploying and configuring refinery.  

This PR is meant to align all configurations into the YAML format and avoid that confusion. After this PR is merged, we will also need to update docs to reflect YAML format for configuration.

- Creates YAML equivalent sample configurations based on existing TOML setup
- Migrates all tests to use a YAML format
- `refinery.service` and `refinery.upstart` reference a YAML based config
- Adds a note to readme about upgrading from prior versions. This note should only apply to those upgrading on non-Kubernetes environments, still using TOML format for config.

### DIFF
--- a/config/config_test_reload_error_test.go
+++ b/config/config_test_reload_error_test.go
@@ -1,3 +1,4 @@
+//go:build all || !race
 // +build all !race
 
 package config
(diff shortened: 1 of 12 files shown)


MikeGoldsmith:
To confirm; is this is a breaking change for anyone using refinery not in Kubernetes because they need to add `refinery.service` and `refinery.upstart`?

config/config_test_reload_error_test.go
```diff
@@ -17,31 +18,31 @@ func TestErrorReloading(t *testing.T) {
 	assert.NoError(t, err)
 	defer os.RemoveAll(tmpDir)
 
-	rulesFile, err := ioutil.TempFile(tmpDir, "*.toml")
```
cartermp, line 20: It's kinda lame, but I feel like we should load and run tests against both if we continue to support both file formats. Wouldn't want to let the toml config bitrot when the majority of our customers (for quite a while) will be using it.

refinery.service
```diff
@@ -3,7 +3,7 @@ Description=Refinery Honeycomb Trace-Aware Sampling Proxy
 After=network.target
 
 [Service]
-ExecStart=/usr/bin/refinery -c /etc/refinery/refinery.toml -r /etc/refinery/rules.toml
+ExecStart=/usr/bin/refinery -c /etc/refinery/refinery.yaml -r /etc/refinery/rules.yaml
```
cartermp, line 6: Is there a way to load this like so?

* Load either the toml or yaml file, depending on which is present
* Error if both are present, ask to pick just one

puckpuck: This would require changing the way refinery loads configs, or bootstrapping all this with a script to check for file existence.

martin308: changing both `refinery.service` and `refinery.upstart` like this seems like it would be a breaking change to me

MikeGoldsmith:
Sorry for delay getting back to this, we've been super busy.

While I like the idea of moving towards using standard config file types, I think it would be unnecessarily disruptive to switch from one format to another and instead would prefer to siupport both toml and json.

MikeGoldsmith:
Moving to icebox because we're not going to be able to work on this soon. If you want to continue on it, please feel free @puckpuck.

PR #432, opened 2022-03-30
WIP: Add debug fields to error response metadata

MikeGoldsmith:
When debugging error responses it would be useful to capture additional information from events. This PR adds support for an array of field names that will be copied from events into the debug metadata that is used to process failed event responses.

If no fields are specified, no additional data is logged.

Example config:
```toml
DebugFieldsToLog = [ "trace.trace_id", "service.name" ]
```

- Introduce `DebugFieldsToLog` in Config interface, with implementations for both file and mock configs
- Check for the fields and use them to copy field data from events into debug messages when handling failed responses

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -139,4 +139,6 @@ type Config interface {
 	GetAddHostMetadataToTrace() bool
 
 	GetEnvironmentCacheTTL() time.Duration
+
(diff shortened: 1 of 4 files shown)


cartermp:
We'll close due to inactivity.

PR #438, opened 2022-04-20
Add support for environment and dataset rules with same names

### LINKED ISSUE #408: Add default environment prefix to allow environments and dataset with same name in rules file
When configuring refinery rules, you cannot specify both an environment and dataset with the same name. This is possible to configure in Honeycomb - eg a legacy production dataset and a production environment.

Add a legacy dataset prefix config setting that differentiate between environments and datasets.

The default would be empty, so no prefix is used.

eg
config.toml
```
...
LegacyDatasetPrefix = "legacy"
...
```

rules.toml
```

# environment
[production]
...

# dataset
[legacy.production]
...
```

**Additional context**

MikeGoldsmith:
Adds support for config files to contain both a dataset and environment with the same name in the same rules config. This is achieved by adding a DatasetPrefix to the config and then using that prefix when setting up rule definitions.

config.toml
```
DatasetPrefix = "dataset"
```

rules.toml
```
Sampler = "DeterministicSampler"
SampleRate = 1

[production] # environment
  Sampler = "DeterministicSampler"
  SampleRate = 10

[dataset.production] # dataset
  Sampler = "DeterministicSampler"
  SampleRate = 20
```

- Closes #408 

- Adds new config option DatasetPrefix, defaults to ""
- If a legacy is provided to the SamplerFactory, it uses the configured prefix along with the sample key to locate rules definitions
- Adds test to verify behaviour

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,36 @@ By default, a Refinery process will register itself in Redis using its local hos
 In environments where domain name resolution is slow or unreliable, override the reliance on name lookups by specifying the name of the peering network interface with the `IdentifierInterfaceName` configuration option.
 See the [Refinery documentation](https://docs.honeycomb.io/manage-data-volume/refinery/) for more details on tuning a cluster.
 
+
+### Mixing Legacy and Environment & Services Rule Definitions
+
+With the change to support environemt and services in Honeycomb, some users will want to support both sending telemetry to a legacy dataset and a new environment called the same thing (eg `production`).
+
+This can be achomplihsed by leveraging the new `DataserPrefix` configuration property and then using that prefix in the rules definitions for the legacy datasets.
+
+When Refinery receives telemetry using an API key associated to a legacy dataset, it will then use the prefix in the form `{prefix}.{dataset}` when trying to resolve the rules definition.
+
+For example
+config.toml
+```toml
+DatasetPrefix = "legacy"
+```
+
+rules.toml
+```toml
+# default rules
+Sampler = "DeterministicSampler"
+SampleRate = 1
+
+    [production] # environment called "production"
+    Sampler = "DeterministicSampler"
+    SampleRate = 5
+
+    [legacy.production] # dataset called "prodiction"
+    Sampler = "DeterministicSampler"
+    SampleRate = 10
+```
+
 ## How sampling decisions are made
 
 In the configuration file, you can choose from a few sampling methods and specify options for each. The `DynamicSampler` is the most interesting and most commonly used. It uses the `AvgSampleRate` algorithm from the [`dynsampler-go`](https://github.com/honeycombio/dynsampler-go) package. Briefly described, you configure Refinery to examine the trace for a set of fields (for example, `request.status_code` and `request.method`). It collects all the values found in those fields anywhere in the trace (eg "200" and "GET") together into a key it hands to the dynsampler. The dynsampler code will look at the frequency that key appears during the previous 30 seconds (or other value set by the `ClearFrequencySec` setting) and use that to hand back a desired sample rate. More frequent keys are sampled more heavily, so that an even distribution of traffic across the keyspace is represented in Honeycomb.
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ See the [Refinery documentation](https://docs.honeycomb.io/manage-data-volume/re
 
 With the change to support environemt and services in Honeycomb, some users will want to support both sending telemetry to a legacy dataset and a new environment called the same thing (eg `production`).
 
-This can be achomplihsed by leveraging the new `DataserPrefix` configuration property and then using that prefix in the rules definitions for the legacy datasets.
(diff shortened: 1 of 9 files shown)


kentquirk (approved):
Generally approved, but two notes:
* I know there's probably going to be a docs-related PR for this as well, but I think it also deserves a note in the README.
* It took me a while to understand the description of this PR. Can you please add a bit of a note as to why there's a problem that needs this fix?

config/file_config.go
```diff
@@ -127,6 +128,7 @@ func NewConfig(config, rules string, errorCallback func(error)) (Config, error)
 	c.SetDefault("HoneycombLogger.LoggerSamplerThroughput", 5)
 	c.SetDefault("AddHostMetadataToTrace", false)
 	c.SetDefault("EnvironmentCacheTTL", time.Hour)
+	c.SetDefault("DatasetPrefix", "")
```
seh, line 131: Though this isn't the first time we've done so, I wonder whether it matters if we set a default value that's equivalent to the field type's zero value. Here, and, as another example, for `AddHostMetadataToTrace` two lines above, what happens if we elide these calls?

sample/sample.go
```diff
@@ -21,10 +22,16 @@ type SamplerFactory struct {
 	Metrics metrics.Metrics `inject:"metrics"`
 }
 
-// GetSamplerImplementationForDataset returns the sampler implementation for the dataset,
-// or nil if it is not defined
-func (s *SamplerFactory) GetSamplerImplementationForDataset(dataset string) Sampler {
-	c, err := s.Config.GetSamplerConfigForDataset(dataset)
+// GetSamplerImplementationForKey returns the sampler implementation for the given
+// samplerKey (dataset or environment), or nil if it is not defined
```
seh, line 26:
```suggestion
// samplerKey (dataset for legacy keys, environment otherwise), or nil if it is not defined
```

sample/sample.go
```diff
@@ -21,10 +22,16 @@ type SamplerFactory struct {
...
+func (s *SamplerFactory) GetSamplerImplementationForKey(samplerKey string, isLegacyKey bool) Sampler {
+	if isLegacyKey {
+		if prefix := s.Config.GetDatasetPrefix(); prefix != "" {
+			samplerKey = fmt.Sprintf("%s.%s", prefix, samplerKey)
```
seh, line 30: You could mandate that the intervening period be part of the specified prefix. If I set the `DatasetPrefix` field to "x," the prefix really becomes "x."—even though I didn't mention the period when specifying my preferred prefix. I might prefer a prefix like "dataset-" (with a separating hyphen instead of a period).

Had you considered including this separator as part of the prefix? Were you attempting to avoid false aliasing if there was no obvious separator character included in the prefix?

config/file_config.go
MikeGoldsmith: Yes, strictly these are not necessary and can be removed for simplicity.

sample/sample.go
MikeGoldsmith: Yes, I agree additional docs on the prefix format where the `.` separator is added for you would be useful 👍🏻 

The separator is not necessary but I think helps create a mental divide. The intent of this change is to support users migrate from legacy datasets to newer environment and services rule sets and should not be a permanent solution. I think a simple solution of always using a single `.` that's well documented should be good enough.

README.md
```diff
@@ -70,6 +70,36 @@ By default, a Refinery process will register itself in Redis using its local hos
 In environments where domain name resolution is slow or unreliable, override the reliance on name lookups by specifying the name of the peering network interface with the `IdentifierInterfaceName` configuration option.
 See the [Refinery documentation](https://docs.honeycomb.io/manage-data-volume/refinery/) for more details on tuning a cluster.
 
+
+### Mixing Legacy and Environment & Services Rule Definitions
+
+With the change to support environemt and services in Honeycomb, some users will want to support both sending telemetry to a legacy dataset and a new environment called the same thing (eg `production`).
+
+This can be achomplihsed by leveraging the new `DataserPrefix` configuration property and then using that prefix in the rules definitions for the legacy datasets.
```
kentquirk, line 78:
```suggestion
This can be accomplished by leveraging the new `DatasetPrefix` configuration property and then using that prefix in the rules definitions for the legacy datasets.
```

README.md
```diff
@@ -70,6 +70,36 @@ By default, a Refinery process will register itself in Redis using its local hos
...
+```toml
+# default rules
+Sampler = "DeterministicSampler"
+SampleRate = 1
+
+    [production] # environment called "production"
+    Sampler = "DeterministicSampler"
+    SampleRate = 5
+
+    [legacy.production] # dataset called "prodiction"
```
kentquirk, line 98:
```suggestion
    [legacy.production] # dataset called "production"
```

kentquirk (approved):
Thanks for the docs improvement. I added suggestions for a couple of typos.

Issue #420, opened 2022-03-08
Add support for all rule conditions to be applied per-span


sozuuuuu:
I am currently creating a sampling rule to use refinery in production. I created the following rule to exclude health checks against my_server. 

```
[ds]

	Sampler = "RulesBasedSampler"

	[[ds.rule]]
		name = "drop healtchecks"
		drop = true
		[[ds.rule.condition]]
			field = "http.target"
			operator = "="
			value = "/health-check"
```

However, unintentionally all the traces like below were dropped.

![Screen Shot 2022-03-08 at 17 56 45](https://user-images.githubusercontent.com/5513797/157202407-9a819d28-3152-4348-aac4-e67e726ef84f.png)

The reason is that the rule traverses all spans and applies the rule if it matches the given condition. In this case, my_server uses the health check endpoint of my_server2.

I have added a condition to the rule to fix this.

```
[ds]

	Sampler = "RulesBasedSampler"

	[[ds.rule]]
		name = "drop healtchecks"
		drop = true
		[[ds.rule.condition]]
			field = "trace.parent_id"
			operator = "not-exists"
		[[ds.rule.condition]]
			field = "http.target"
			operator = "="
			value = "/health-check"
```

However, due to the logic I mentioned earlier, such a trace would be dropped.

Is there a way to set up the filter that "the filter is applied to the traces that contain at least one span that matches the n conditions" rather than "given n conditions, if the trace contains n spans that match the condition, apply the filter". If not, I would appreciate if you support this.

MikeGoldsmith:
Hi @sozuuuuu 

Thanks for creating the issue. I'm not sure how we'd work around this, I'll check with some other folks who are better versed in setting up more complicated rule sets and report back.

MikeGoldsmith:
Hi @sozuuuuu - I'm sorry for the delay in getting back to you, we've been really busy the past few weeks.

Have you found a workaround for the issue?

isnotajoke:
FWIW, this is also something we'd like to be able to do. For context, we have a number of microservices logging to a single dataset in Honeycomb, and we use the rules-based sampler to sample these services. These services sometimes interact with each other: we often see traces that start at service A and then make calls to services B, C, D, etc. 

These services can see dramatically different load levels between them, and we'd like to be able to sample per-service – it would be hard to accurately and predictably account for differing request volume with a single across the board sampling setting.

What we'd like to be able to do is something like:

* For traces originating at service X, sample at 10
[...]

PR #440, opened 2022-04-27
Add rule Scope configuration option to rules-based sampler

isnotajoke:

We'd like to be able to apply all of the conditions in a rules-based sampler rule to a single span. In other words, we'd like a rule to be considered matched only if *all* of the conditions in that rule are satisfied by a single span in the trace. This allows us to more accurately make per-service sampling decisions in a dataset written to by multiple services.

- Closes see #420

At a high level, we add a new `Scope` configuration option to the rule definition. By default (and if unspecified) it is "trace". If "span", all conditions in the rule must be satisfied together by some span in a trace. This is a subtle but important difference with the default behavior, effectively making a rule more strict than it would otherwise be.

`Scope` introduces some complexity to the rule based sampler. While many of the operations performed within the 
`GetSampleRate` method (extracting a value, checking for a match) are similar between the two approaches, the shape of the logic to compute a match is totally different when `Scope` is 'trace' than when it is 'span'. To avoid duplication, we refactor the common methods into private functions; this also makes the logic to loop through & compute a match a little easier to understand at a glance.

I added a basic spec to cover the new behavior.

### DIFF
--- a/config/sampler_config.go
+++ b/config/sampler_config.go
@@ -62,6 +62,7 @@ type RulesBasedSamplerRule struct {
 	SampleRate int
 	Sampler    *RulesBasedDownstreamSampler
 	Drop       bool
+	MatchSpan  bool
 	Condition  []*RulesBasedSamplerCondition
 }
 
--- a/config/sampler_config.go
+++ b/config/sampler_config.go
@@ -62,7 +62,7 @@ type RulesBasedSamplerRule struct {
 	SampleRate int
 	Sampler    *RulesBasedDownstreamSampler
 	Drop       bool
-	MatchSpan  bool
+	Scope      string `validate:"oneof=span trace"`
 	Condition  []*RulesBasedSamplerCondition
 }
 
--- a/rules_complete.toml
+++ b/rules_complete.toml
@@ -250,7 +250,7 @@ SampleRate = 1
 		#
 		# this is especially helpful when sampling a dataset written to
 		# by multiple services that call one another in normal operation –
-		# you can set MatchSpan to true to attribute traces to an origin
+		# you can set Scope to 'span' to attribute traces to an origin
 		# service in a way that would be difficult without it.
 		Scope = "span"
 		[[dataset4.rule.condition]]
--- a/rules_complete.toml
+++ b/rules_complete.toml
@@ -243,6 +243,25 @@ SampleRate = 1
 			AddSampleRateKeyToTraceField = "meta.refinery.dynsampler_key"
 
 	[[dataset4.rule]]
+		name = "sample traces originating from a service"
+		# if true, a single span must match all of the conditions in this
+		# rule in order for the rule to match the trace.
+		#
+		# this is especially helpful when sampling a dataset written to
+		# by multiple services that call one another in normal operation –
+		# you can set MatchSpan to true to attribute traces to an origin
+		# service in a way that would be difficult without it.
+		MatchSpan = true
+		[[dataset4.rule.condition]]
+			field = "service name"
+			operator = "="
+			value = "users"
+		[[dataset4.rule.condition]]
+			field = "meta.span_type"
+			operator = "="
+			value = "root"
+
+	[[dataset4.rule]]
 		SampleRate = 10 # default when no rules match, if missing defaults to 10
 
 [dataset5]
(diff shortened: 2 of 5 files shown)


config/sampler_config.go
```diff
@@ -62,6 +62,7 @@ type RulesBasedSamplerRule struct {
 	SampleRate int
 	Sampler    *RulesBasedDownstreamSampler
 	Drop       bool
+	MatchSpan  bool
```
isnotajoke, line 65: Probably need to add this to the documentation + example rule configurations.

MikeGoldsmith (changes requested):
Hey @isnotajoke - thank you for putting this together. Sorry for the delay in getting it reviewed.

Overall, I really like what you've done and think it's a great solution. I've left a suggestion for changing the configuration option - let me know you thoughts.

isnotajoke:
@MikeGoldsmith makes sense to me. I'll go ahead and make that change.

isnotajoke:
We're planning to test this PR in our infra this week. Will update with any concerns/follow up tweaks after that, and then it should be good for final review.

sample/rules.go
```diff
@@ -67,10 +67,17 @@ func (s *RulesBasedSampler) GetSampleRate(trace *types.Trace) (rate uint, keep b
...
 			matched = ruleMatchesSpanInTrace(trace, rule, s.Config.CheckNestedFields)
-		} else {
+		case "trace", "":
 			matched = ruleMatchesTrace(trace, rule, s.Config.CheckNestedFields)
+		default:
+			logger.WithFields(map[string]interface{}{
+				"rule_name": rule.Name,
+				"scope":     rule.Scope,
+			}).Logf("invalid scope %s given for rule: %s", rule.Scope, rule.Name)
+			matched = false
```
MikeGoldsmith, line 80: I think we should default to send traces for misconfigured rules.

```suggestion
			matched = true
```

MikeGoldsmith (dismissed):
Looks great, thanks @isnotajoke.

Yes, please do test and report back any findings. We'll hold back on merging this for your feedback.

rules_complete.toml
```diff
@@ -242,6 +242,26 @@ SampleRate = 1
...
+	[[dataset4.rule]]
+		name = "sample traces originating from a service"
+		# if scope is set to "span", a single span in the trace must match
+		# *all* of the conditions associated with this rule for the rule to
+		# apply to the trace.
+		#
+		# this is especially helpful when sampling a dataset written to
+		# by multiple services that call one another in normal operation –
+		# you can set MatchSpan to true to attribute traces to an origin
+		# service in a way that would be difficult without it.
```
ajvondrak, line 254: Missed this comment when changing from `MatchSpan` to `Scope`. 👀

rules_complete.toml
```diff
@@ -242,6 +242,26 @@ SampleRate = 1
...
 
+	[[dataset4.rule]]
+		name = "sample traces originating from a service"
+		# if scope is set to "span", a single span in the trace must match
+		# *all* of the conditions associated with this rule for the rule to
+		# apply to the trace.
+		#
+		# this is especially helpful when sampling a dataset written to
+		# by multiple services that call one another in normal operation –
+		# you can set MatchSpan to true to attribute traces to an origin
```
MikeGoldsmith, line 253:
```suggestion
		# you can set Scope to 'span' to attribute traces to an origin
```

MikeGoldsmith:
FYI the test failure is coming from [config_test](https://github.com/honeycombio/refinery/blob/main/config/config_test.go#L249) - you just need to update the number of rules in the dataset to be 5.

MikeGoldsmith:
> We're planning to test this PR in our infra this week. Will update with any concerns/follow up tweaks after that, and then it should be good for final review.

Hey @isnotajoke - just wanted to check in to see how your testing went?

isnotajoke:
@MikeGoldsmith still ongoing. Should be complete this week.

sample/rules.go
isnotajoke: Makes sense. Will update.

MikeGoldsmith:
Hey @isnotajoke - just wanted to check in to see if you've completed your internal testing?

We're eager to move forward with this feature and value your feedback.

MikeGoldsmith:
Hey @isnotajoke - I'm going to look to move ahead with this change this week. Let me know if you have any additional feedback.

isnotajoke:
@MikeGoldsmith sorry for the radio silence here. We completed our testing and didn't see any issues. No concerns with moving ahead.

PR #455, opened 2022-05-16
Fix crash bug related to sharding

### LINKED ISSUE #454: Sharding bug can crash refinery
This bug was found from code inspection of a stack trace.

In [WhichShard](https://github.com/honeycombio/refinery/blob/main/sharder/deterministic.go#L218):

```go
	// add in the sharding salt to ensure the sh1sum is spread differently from
	// others that use the same algorithm
	sum := sha1.Sum([]byte(traceID + shardingSalt))
	v := bytesToUint32be(sum[:4])

	portion := math.MaxUint32 / len(d.peers)
	index := v / uint32(portion)

	return d.peers[index]
```

The algorithm is dividing MaxUInt32 by the number of peers, but this doesn't divide evenly unless the number of peers is a power of 2. Consequently, when `v` is within the number of peers of `MaxUInt32`, then the result of index is equal to the number of peers, and then `d.peers[index]` will be out of range.

Example: len(d.peers) == 20, and v is 0xFFFFFFF9.

portion will be 214748364, and then index will be 20, which is out of range.

kentquirk:

- Fixes a bug where the shard index can get out of range if the last 4 bytes of the hash of the trace ID is close to 0xFFFFFFFF and the number of shards doesn't divide that value evenly. The algorithm in WhichShard works correctly for divisors of 2^32-1. The prime factorization of that includes 1, 3, 5, 17 (and products of those).

- If shard would be out of range, assigns it to shard 0. This has the most minimal effect on existing behavior, especially when the number of shards changes.
- Adds a test that failed before the out of range check, but works with it in.

Fixes #454

### DIFF
--- a/sharder/deterministic.go
+++ b/sharder/deterministic.go
@@ -227,6 +227,12 @@ func (d *DeterministicSharder) WhichShard(traceID string) Shard {
 	portion := math.MaxUint32 / len(d.peers)
 	index := v / uint32(portion)
 
+	// #454 -- index can get out of range if v is close to 0xFFFFFFFF and portion would be non-integral.
+	// Consider revisiting this with a different sharding mechanism if we rework our scaling behavior.
+	if index > uint32(len(d.peers)) {
+		index = 0
+	}
+
 	return d.peers[index]
 }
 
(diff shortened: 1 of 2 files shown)


kentquirk:
That sounds like a fine idea. Will do!

PR #459, opened 2022-05-25
Replace hand-rolled binary.BigEndian.Uint32 with the real deal

toshok:
`bytesToUint32be` is exactly `binary.BigEndian.Uint32` ([go source](https://cs.opensource.google/go/go/+/refs/tags/go1.18.2:src/encoding/binary/binary.go;l=111)), and is duplicated in the two spots where we use it. might as well save a few lines of code.

### DIFF
--- a/sample/deterministic.go
+++ b/sample/deterministic.go
@@ -2,6 +2,7 @@ package sample
 
 import (
 	"crypto/sha1"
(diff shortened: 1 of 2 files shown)


MikeGoldsmith (approved):
Thanks @toshok 👍🏻

PR #465, opened 2022-06-10
Validate successful span scoped rules test

MikeGoldsmith:
The test to validate the new span scoped rule sets in #440 only validated the keep return property for unsuccessful sets of spans. This PR updates the test to verify that the keep property is correct by updating the sample rate on the rule set to 1.

- Update sample rate in the test to 1, so we can validate the keep return property

### DIFF
--- a/sample/rules_test.go
+++ b/sample/rules_test.go
@@ -844,7 +844,7 @@ func TestRuleMatchesSpanMatchingSpan(t *testing.T) {
 					{
 						Name:       "Rule to match span",
 						Scope:      "span",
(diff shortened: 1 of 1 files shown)


vreynolds:
So I can change all the span data to *not* match the rule, and the test still passes. I don't know that it's actually testing what we expect?

MikeGoldsmith:
I've updated the test cases to be clearer on what they are testing, including testing both span and trace scopes.

vreynolds (approved):
🎈

PR #474, opened 2022-07-06
maint: change title on release workflow from (unset) var to hard-coded string

JamieDanielson:

- Error in previous workflow that attempted to create an issue states "Input required and not supplied: title".

- While the title does match that of the [example](https://github.com/actions-ecosystem/action-create-issue#example) provided in the actions repo, it is based on a variable that is set in the example that we do not have set. I believe this is causing a blank title which is not allowed, as the title value is required. This PR changes the title from a variable that is unset (`{ steps.date.outputs.today }}`) to a hard-coded string (`Bump Refinery to Latest Version`).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,7 +11,7 @@ jobs:
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           repo: github.com/honeycombio/helm-charts
-          title: ${{ steps.date.outputs.today }}
+          title: Bump Refinery to Latest Version
           body: |
             ## Bump Refinery
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -2,6 +2,7 @@ name: Create helm chart issue on release
 on:
   release:
     types: [published]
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -2,6 +2,7 @@ name: Create helm chart issue on release
 on:
   release:
     types: [published]
+  workflow_dispatch:
```
vreynolds, line 5: is this an intentional change?

JamieDanielson: `workflow_dispatch` is the command to run a workflow on-demand. I was hoping this would give me an option right from this PR to test it, but I'm not seeing it (unless I'm missing it somewhere). So I'm thinking we keep it in so I can try it manually after merging to main to see if it creates an issue properly, and if it doesn't, then we know this workflow still doesn't work (as opposed to waiting for the next release to see that it doesn't work). Open to rejection on that idea though, and just see what happens.

JamieDanielson:
This still did not work 😞  Created new issue #476 for next effort.

Issue #476, opened 2022-07-11
auto-create issue in helm chart for new refinery release


JamieDanielson:
**Is your feature request related to a problem? Please describe.**

follow up from #435 which still has some kinks to be worked out... we want an issue auto-created in helm-charts upon release of a new Refinery.

vreynolds:
https://github.com/honeycombio/refinery/runs/7308731041

PR #477, opened 2022-07-12
maint: update GH action token

vreynolds:

- updates #476 

- GITHUB_TOKEN is a limited token with only enough permissions to modify current repo


JamieDanielson (approved):
oh

PR #491, opened 2022-08-29
Added PeerManagement.Timeout config option

thrawn01:
This PR adds a new config option called `PeerManagement.Timeout` which will be used when communicating with peer discovery systems implementing the `Peers` interface. This PR also add support for `context.Context` when first connecting to the `Peers` implementation.

- `NewPeers()` now accepts a `context.Context`
- Moved `NewPeers()` initialization after logging setup to take advantage of logs during any initial connect attempts. 
- Added `//go:build all || race` which is the new 1.18+ format for build tags (I can revert if not needed, my IDE auto fixed it 😄 )
- Fixed minor memory leak in `Router` where the `cancel()` function is never called.

## Note
I'm noting here that `context.Context` passed to `NewPeer()` is not as useful to redis because the redis dial function already has a retry and timeout loop. However, it will be useful for other concrete implementations of the `Peers` interface. If this PR is accepted, I wish to follow up with a PR that uses https://github.com/hashicorp/memberlist for peer discovery.


internal/peer/redis.go
```diff
@@ -163,11 +172,19 @@ func (p *redisPeers) watchPeers() {
...
 
 	for range tk.C {
-		currentPeers, err := p.store.GetMembers(context.TODO())
+		ctx, cancel := context.WithTimeout(context.Background(), p.c.GetPeerTimeout())
+		currentPeers, err := p.store.GetMembers(ctx)
+		cancel()
+
 		if err != nil {
-			// TODO maybe do something better here?
+			logrus.WithField("name", p.publicAddr).
```
kentquirk, line 180: A nit, bu

internal/peer/redis.go
```diff
@@ -139,15 +139,24 @@ func (p *redisPeers) RegisterUpdatedPeersCallback(cb func()) {
...
 
 func (p *redisPeers) updatePeerListOnce() {
-	currentPeers, err := p.store.GetMembers(context.TODO())
+	ctx, cancel := context.WithTimeout(context.Background(), p.c.GetPeerTimeout())
+	defer cancel()
+
+	currentPeers, err := p.store.GetMembers(ctx)
 	if err != nil {
-		// TODO maybe do something better here?
+		logrus.WithField("name", p.publicAddr).
```
kentquirk, line 156: A nit, but could you please do this with logrus.WithFields rather than a succession of WithField calls?

app/app_test.go
```diff
@@ -1,3 +1,4 @@
+//go:build all || race
```
kentquirk, line 1: We're on 1.18+, so this is a good change. Thank you.

kentquirk (approved):
I like where this is headed -- I'll likely have more comments on the next one. But the stuff in this one is useful, so please just use AddFields and we can run with this.

internal/peer/redis.go
kentquirk: @thrawn01 -- to clarify, although I approved this PR, I would like to see that change before we merge it. Are you able to do that, please?

MikeGoldsmith (changes requested):
As per @kentquirk's request, please can you update your logs to use WithFields 👍🏻

MikeGoldsmith:
This will also need a rebase.

Baliedge:
> This will also need a rebase to resolve conflicts.

Done.

Baliedge:
> @Baliedge please could you update the log to use WithFields as @kentquirk suggestion too?

Also done.

internal/peer/redis.go
Baliedge: Done.

MikeGoldsmith (approved):
Thanks @thrawn01 @Baliedge 👍🏻

PR #498, opened 2022-09-03
Implement grpc-health-probe

### LINKED ISSUE #414: Support grpc Health Check Protocol
Refinery should support the standard [grpc Health Check Protocol](https://github.com/grpc-ecosystem/grpc-health-probe), especially for use within k8s.

abatilo:
Similar to the `/x/alive` endpoint for the http server, this PR
implements the gRPC handlers for the [gRPC Health Checking
Protocol](https://github.com/grpc/grpc/blob/592bfb9c0040fba382e5dc106e2ebbf0c96465fb/doc/health-checking.md).

I was able to verify that this works by running refinery with the
complete config to set `GRPCListenAddr = "0.0.0.0:9090"`:
```
⇒  go run ./cmd/refinery -c config_complete.toml -r rules_complete.toml
...
```

And then I could run the
[grpc-health-probe](https://github.com/grpc-ecosystem/grpc-health-probe)
binary:
```
⇒  grpc-health-probe -addr=localhost:9090
status: SERVING
```

Closes #414

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -27,8 +27,9 @@ require (
 	github.com/tidwall/gjson v1.14.3
 	github.com/vmihailenco/msgpack/v4 v4.3.11
 	go.opentelemetry.io/proto/otlp v0.11.0
+	golang.org/x/net v0.0.0-20220826154423-83b083e8dc8b // indirect
 	google.golang.org/grpc v1.49.0
-	google.golang.org/protobuf v1.28.0
+	google.golang.org/protobuf v1.28.1
 	gopkg.in/alexcesaro/statsd.v2 v2.0.0
 	gopkg.in/yaml.v2 v2.4.0
 )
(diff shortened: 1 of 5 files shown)


abatilo:
Tagging @kentquirk also since that's who had opened the original issue!

kentquirk:
@abatilo -- Thank you for this! We are really trying not to add new code to things without having a test for it. It verifies the functionality, but also provides a backstop against regressions later as well as documentation of expected behavior. 

Any chance you could write tests to verify this? Ideally they'd operate by an endpoint request, but even a test at the Route level would help.

MikeGoldsmith (changes requested):
As @kentquirk requested, please can you add a test to verify the new endpoint works as expected 👍🏻

abatilo:
> @abatilo -- Thank you for this! We are really trying not to add new code to things without having a test for it. It verifies the functionality, but also provides a backstop against regressions later as well as documentation of expected behavior.
> 
> Any chance you could write tests to verify this? Ideally they'd operate by an endpoint request, but even a test at the Route level would help.

Thanks for the feedback, @kentquirk. I noticed that there weren't tests covering endpoints like `/alive` so I wasn't 100% sure what the preference was going to be. I'll think about how I might implement this and get back to you!

abatilo:
Hi @MikeGoldsmith @kentquirk.

How do we feel about [these tests](https://github.com/honeycombio/refinery/pull/498/files#diff-2e9ae5a252a7f32fe100e1e6c89103596d3c07999b8c010357aa626fe25d9e2fR524-R559)?

I took inspiration from [otlp_trace_test.go](https://github.com/abatilo/refinery/blob/ce4bf6a8df76462369cdd75965220efcd929483c/route/otlp_trace_test.go#L68-L80).

Thanks again for helping me contribute.

MikeGoldsmith:
Yep, these look great - thanks @abatilo 👍🏻

MikeGoldsmith (approved):
Thanks for your contribution @abatilo 🎉

PR #499, opened 2022-09-03
Make gRPC ServerParameters configurable

### LINKED ISSUE #263: Make grpc ServerParameters configurable
gRPC has a set of [ServerParameters](https://pkg.go.dev/google.golang.org/grpc/keepalive#ServerParameters) that we should make configurable.

Specifically, the MaxConnectionAge and MaxConnectionAgeGrace are of interest to help keep gRPC connections balanced across a pool of refinery hosts.

An issue can arise when a host dies that all the connections will failover to the other hosts and remain there due to the long-lived nature of gRPC connections. The new host that spins up to replace the dead host will never receive any incoming connections and is then underutilized while the other hosts are overburdened.

abatilo:
Which problem is this PR solving
--------------------------------

At the moment, there is no way to override the ServerParameters used in
the `keepalive` configuration of the gRPC server. As to @bdarfler's
comments in #263, issues can arise where new hosts will never be
connected to because clients will reconnect to already existing hosts
and by the time replacement hosts are available, all clients will have
reconnected to existing hosts with a
`MaxConnectionAge`/`MaxConnectionAgeGrace` of
[infinity](https://github.com/grpc/grpc-go/blob/60a3a7e969c401ca16dbcd0108ad544fb35aa61c/internal/transport/defaults.go#L36-L37).

Short description of the changes
--------------------------------

The changes made are to plumb through the configuration of the GRPC
ServerParameters from the viper configuration and into the instantiation
of the actual gRPC server. These changes kept the previous default
configuration of refinery as the defaults of the configurable settings,
so anyone who doesn't explicitly override any of these configurations
should have exactly the same behavior.

Closes #263

### DIFF
--- a/config_complete.toml
+++ b/config_complete.toml
@@ -323,3 +323,44 @@ MetricsReportingInterval = 3
 # listener.
 # Not eligible for live reload.
 # MetricsListenAddr = "localhost:2112"
+
+###########################
+## gRPC ServerParameters ##
+###########################
+
+# Reflects: https://pkg.go.dev/google.golang.org/grpc/keepalive#ServerParameters
+
+[GRPCServerParameters]
+
+# MaxConnectionIdle is a duration for the amount of time after which an
+# idle connection would be closed by sending a GoAway. Idleness duration is
+# defined since the most recent time the number of outstanding RPCs became
+# zero or the connection establishment.
+# Not eligible for live reload.
+# MaxConnectionIdle = "1m"
+
+# MaxConnectionAge is a duration for the maximum amount of time a
+# connection may exist before it will be closed by sending a GoAway. A
+# random jitter of +/-10% will be added to MaxConnectionAge to spread out
+# connection storms.
+# 0s sets duration to infinity which is the default.
+# Not eligible for live reload.
+# MaxConnectionAge = "0s"
+
+# MaxConnectionAgeGrace is an additive period after MaxConnectionAge after
+# which the connection will be forcibly closed.
+# 0s sets duration to infinity which is the default.
+# Not eligible for live reload.
+# MaxConnectionAgeGrace = "0s"
+
+# After a duration of this time if the server doesn't see any activity it
+# pings the client to see if the transport is still alive.
+# If set below 1s, a minimum value of 1s will be used instead.
+# Not eligible for live reload.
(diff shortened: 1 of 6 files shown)


config_complete.toml
```diff
@@ -315,3 +315,44 @@ MetricsReportingInterval = 3
...
+
+# MaxConnectionAgeGrace is an additive period after MaxConnectionAge after
+# which the connection will be forcibly closed.
+# 0s sets duration to infinity which is the default.
+# Not eligible for live reload.
+# MaxConnectionAgeGrace = "0s"
+
+# After a duration of this time if the server doesn't see any activity it
+# pings the client to see if the transport is still alive.
+# If set below 1s, a minimum value of 1s will be used instead.
```
kentquirk, line 350: According to the [docs here](https://pkg.go.dev/google.golang.org/grpc@v1.49.0/keepalive#ClientParameters), the lower limit is 10s, not 1s, unless I'm reading the wrong thing. Can you check please, and also note the meaning of 0s?

kentquirk (commented):
I like this a lot. Please check the default thing, but otherwise I'm ready to approve this.

MikeGoldsmith (changes requested):
As per @kentquirk's request. Please check min value and update docs 👍🏻

config_complete.toml
abatilo: Thanks for pointing that out @kentquirk! Although, it looks like what you linked to is for the `ClientParameters`.

I grabbed the comments here for the [ServerParameters](https://github.com/grpc/grpc-go/blob/v1.49.0/keepalive/keepalive.go#L66).

I've also pushed a commit for adding notes about the meaning of 0s!

abatilo:
I think there might be a bug with GitHub. I tried to re-request reviews from both @kentquirk and @MikeGoldsmith but it seems to have removed @emilyashley and then says I need to re-request review from either Kent or Mike. And if I re-request one, the other gets shown as needing a re-request.

image

Whatever GitHub did, I updated the docs and commented about the min value in question!

kentquirk (approved):
Thanks for the updates!

MikeGoldsmith (approved):
Looks great, thanks @abatilo 👍🏻

PR #500, opened 2022-09-04
Add endpoints to help debug refinery rules

kentquirk:

There have been multiple reports of people not realizing their configs were not parsing the way they were intended. This adds two endpoints:
- `/debug/allrules/$FORMAT` will retrieve the entire rules configuration
- `/debug/rules/$FORMAT/$DATASET` will retrieve the rule set that refinery will use for the specified dataset. It comes back as a map of the sampler type to its rule set.

- Add the endpoints
- Add a feature to config to return the full rule set
- Extend existing config feature to return the sampler name along with its rules struct
- Add/expand tests for it all

### DIFF
--- a/route/route.go
+++ b/route/route.go
@@ -276,6 +276,7 @@ func (r *Router) getSamplerRules(w http.ResponseWriter, req *http.Request) {
 	if err != nil {
 		w.Write([]byte(fmt.Sprintf("got error %v trying to fetch config for dataset %s\n", err, dataset)))
 		w.WriteHeader(http.StatusBadRequest)
+		return
 	}
 	r.marshalToFormat(w, map[string]interface{}{name: cfg}, format)
 }
@@ -286,6 +287,7 @@ func (r *Router) getAllSamplerRules(w http.ResponseWriter, req *http.Request) {
 	if err != nil {
 		w.Write([]byte(fmt.Sprintf("got error %v trying to fetch configs", err)))
 		w.WriteHeader(http.StatusBadRequest)
+		return
 	}
 	r.marshalToFormat(w, cfgs, format)
 }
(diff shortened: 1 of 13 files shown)


route/route.go
```diff
@@ -264,6 +269,60 @@ func (r *Router) debugTrace(w http.ResponseWriter, req *http.Request) {
 	w.Write([]byte(fmt.Sprintf(`{"traceID":"%s","node":"%s"}`, traceID, shard.GetAddress())))
 }
 
+func (r *Router) getSamplerRules(w http.ResponseWriter, req *http.Request) {
+	format := strings.ToLower(mux.Vars(req)["format"])
+	dataset := mux.Vars(req)["dataset"]
+	cfg, name, err := r.Config.GetSamplerConfigForDataset(dataset)
+	if err != nil {
+		w.Write([]byte(fmt.Sprintf("got error %v trying to fetch config for dataset %s\n", err, dataset)))
+		w.WriteHeader(http.StatusBadRequest)
```
MikeGoldsmith, line 278: Should return after writing body & header.

route/route.go
```diff
@@ -264,6 +269,60 @@ func (r *Router) debugTrace(w http.ResponseWriter, req *http.Request) {
...
+	}
+	r.marshalToFormat(w, map[string]interface{}{name: cfg}, format)
+}
+
+func (r *Router) getAllSamplerRules(w http.ResponseWriter, req *http.Request) {
+	format := strings.ToLower(mux.Vars(req)["format"])
+	cfgs, err := r.Config.GetAllSamplerRules()
+	if err != nil {
+		w.Write([]byte(fmt.Sprintf("got error %v trying to fetch configs", err)))
+		w.WriteHeader(http.StatusBadRequest)
```
MikeGoldsmith, line 288: Should return after writing body & header.

MikeGoldsmith (approved):
Looks good - suggested small fix to return early when handling bad requests

PR #502, opened 2022-09-05
Rename /debug endpoints to /query and add auth

kentquirk:

- Adds a local query token to allow the changes in #500 to be protected by a locally-defined token in a header.

- Before #500, there was a single /debug/trace endpoint, but there is also an optional /debug endpoint intended for use with a debugger and profiler. So I renamed /debug to /query for the trace and config.
- I added to these new /query endpoints a local configuration value `QueryAuthToken` that can be specified in the config file or in the environment. If specified, the `X-Honeycomb-Refinery-Query` header must be specified on a query request.
- If the QueryAuthToken is not specified in the configuration, the /query endpoints are not accessible. 

Note that because the `/debug/trace` request has been renamed and is now protected by a token, this is technically a breaking change (although `/debug/trace` isn't intended for regular operation and wasn't documented outside of the source code).

This PR is currently based against #500, and could be merged to that one or wait until that one merges.


route/middleware.go
```diff
@@ -17,6 +17,28 @@ func init() {
 	rand.Seed(time.Now().UnixNano())
 }
 
+func (r *Router) queryTokenChecker(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+		requiredToken := r.Config.GetQueryAuthToken()
+		token := req.Header.Get(types.QueryTokenHeader)
+		fmt.Printf("required: %s  specified: %s\n", requiredToken, token)
```
MikeGoldsmith, line 24: Should we use logrus here instead?

route/route.go
```diff
@@ -155,9 +155,14 @@ func (r *Router) LnS(incomingOrPeer string) {
 	muxxer.HandleFunc("/alive", r.alive).Name("local health")
 	muxxer.HandleFunc("/panic", r.panic).Name("intentional panic")
 	muxxer.HandleFunc("/version", r.version).Name("report version info")
-	muxxer.HandleFunc("/debug/trace/{traceID}", r.debugTrace).Name("get debug information for given trace ID")
-	muxxer.HandleFunc("/debug/rules/{format}/{dataset}", r.getSamplerRules).Name("get formatted sampler rules for given dataset")
-	muxxer.HandleFunc("/debug/allrules/{format}", r.getAllSamplerRules).Name("get formatted sampler rules for all datasets")
+
+	// require a local auth for query usage
```
MikeGoldsmith, line 159: Should we keep the old /debug paths and return an error? Removing these would be a breaking change for anyone who's using them, albeit a feature that's only used to debug.

MikeGoldsmith (changes requested):
Looks like a good change. I've left a query regarding removing the old endpoints.

MikeGoldsmith (approved):

route/middleware.go
kentquirk: Nah, we should delete the printf I left in by mistake. :)

route/route.go
kentquirk: Yeah, they were implemented for the first time in that one; there are no users of this feature.

PR #503, opened 2022-09-06
Optionally record why a sample decision was made

### LINKED ISSUE #473: Add option to rules to add rule name to span when matched
I want to know which rule was applied to a span.

When a rule is matched and applied, it would be nice to have a field to understand which rule got applied. Since rules can use dynamic samplers, the key should be slightly different than what is used with the dynamic samplers to understand when both a rule & dynamic sampler are used.

Usage mode in Honeycomb is the best bet to figure out if maybe something was sent through a rules sampler.

**Additional context**

kentquirk:

- Fixes #473 

- Adds a config flag, `AddRuleReasonToTrace` that adds a new field to traces called `meta.refinery.reason`, which is a string field that contains information about the sampler(s) that made a trace decision. If it was a rule-based sampler, it includes the rule name.
- Also removes unnecessary build flags that were causing tests not to run.
- Updates README and config file

### DIFF
--- a/collect/collect.go
+++ b/collect/collect.go
@@ -343,12 +343,13 @@ func (i *InMemCollector) processSpan(sp *types.Span) {
 		}
 
 		trace = &types.Trace{
-			APIHost:   sp.APIHost,
-			APIKey:    sp.APIKey,
-			Dataset:   sp.Dataset,
-			TraceID:   sp.TraceID,
-			StartTime: time.Now(),
-			SendBy:    time.Now().Add(timeout),
+			APIHost:    sp.APIHost,
+			APIKey:     sp.APIKey,
+			Dataset:    sp.Dataset,
+			TraceID:    sp.TraceID,
+			StartTime:  time.Now(),
+			SendBy:     time.Now().Add(timeout),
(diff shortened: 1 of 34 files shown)


collect/collect.go
```diff
@@ -456,7 +456,7 @@ func (i *InMemCollector) send(trace *types.Trace) {
 	}
 
 	// make sampling decision and update the trace
-	rate, shouldSend := sampler.GetSampleRate(trace)
+	rate, shouldSend, why := sampler.GetSampleRate(trace)
```
MikeGoldsmith, line 459: nit: for consistency, could we also call the result parameter reason?

MikeGoldsmith (approved):
Overall, looks good. I've left some minor feedback 👍🏻 

What do you think about a consistent template for reason strings?

eg
"deterministic/always"
"deterministic/chance"
"totalthroughput/{key}"
"dynamic/{key}"
"emadynamic/{key}"
"rules/span/{rule_name}"
"rules/trace/no_rule"
"rules/invalid_scope/{rule_name}

kentquirk:
I like *all* of your suggestions. Thanks, will implement.

MikeGoldsmith (approved):

PR #504, opened 2022-09-06
Fix sample rate for late spans

### LINKED ISSUE #468: Sample rate not being set correctly for spans that show up late
It looks like [dealWithSentTrace](https://github.com/honeycombio/refinery/blob/main/collect/collect.go#L384) is not [setting 0 sample rates to 1](https://github.com/honeycombio/refinery/blob/c71e4f6f3fa8d131663c05f6bb87afc9cd25b753/collect/collect.go#L484) before doing its multiplication. The result is that some spans show up at Honeycomb without having SampleRate set (because the sample rate is the zero value so therefore not set).

**Additional context**

There is an internal slack thread at Honeycomb where this problem was discovered.

kentquirk:

- Fixes #468

- If sample rate < 1 on a late span, set it to 1 before doing any further sample rate math.

### DIFF
--- a/collect/collect.go
+++ b/collect/collect.go
@@ -358,7 +358,7 @@ func (i *InMemCollector) processSpan(sp *types.Span) {
 	}
 	// if the trace we got back from the cache has already been sent, deal with the
 	// span.
-	if trace.Sent == true {
+	if trace.Sent {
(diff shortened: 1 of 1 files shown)


MikeGoldsmith (approved):
tenor-59491240

Issue #490, opened 2022-08-04
Use event sample rate in calculation for refinery's sample rate value


jwilm:
On traces that already have a positive `sample_rate` attribute like `10`, Refinery will overwrite it with whatever rate Refinery sampled it at rather than the product of the two rates. Eg, if Refinery samples that trace at a rate of 2, the SampleRate should be set to 20 rather than 2.

This results in Honeycomb under-reporting total event volume and makes it more difficult to reason about span volume.

**Versions**

- Go: any
- Refinery: latest

**Steps to reproduce**

1. Send a span with the `sample_rate` attribute already set to a number greater than 1
2. Observe Refinery overwriting it with whatever rate Refinery samples it at

**Additional context**

We have tail sampling implemented in some of our high volume applications as to not burden our already large Refinery cluster with a bunch of extra work.

MikeGoldsmith:
Hey @jwilm - thanks for raising the issue.

Refinery is not designed to take into account sample rate set on spans/events by the sending app / an intermediate process. I don' think this isn't a bug but an enhancement we could consider adding if required.

I've updated title / label to better represent the ask. Let me know if you think otherwise 👍🏻

jwilm:
Sounds good; I didn't have the context to know which way this one went :wink:.

This would be a very welcome enhancement. As we continue to ramp up our observability efforts, we are finding ourselves needing to do more and more sampling in the applications as to avoid large spend on our Refinery cluster. However, this currently has the drawback of not being able to rely on Honeycomb to know the volume of events since they end up being (sometimes severely) underreported.

MikeGoldsmith:
Sure, I understand your use case and think your proposed solution makes sense. We'd likely want to make this an opt-in feature, so not to affect current users who may expect this current calculations to work as they are now.

I'm unsure when we'll get to this but we're always welcoming of community PRs if you feel you'd like to contribute 👍🏻

robbkidd:
An addendum to this and in the spirit of #473, the implementation of this calculation of a final sample rate ought to include some further (optional?) annotations on the sampled spans about original sample rate (`meta.refinery.original_sample_rate`? `meta.refinery.upsampled_from_rate`? 🤔) so as to leave a bread crumb trail to investigate weird sampling results.

kentquirk:
This was fixed by #508.

PR #508, opened 2022-09-08
Add 'meta.refinery.original_sample_rate'

epvanhouten:

Refinery re-writing the sample rate was causing some concern. Writing down the sample rate sent to refinery prior to updating the sample rate should allow for more/better debug.

This is implementing the suggestion of @robbkidd in #490 to add annotations to the spans of their original sample rate. This will allow for better understanding of what is happening when there are multiple tiers of sampling happening in the telemetry processing.

At the places where the Span SampleRate was combined with the Trace SampleRate, the original span SampleRate is copied down into a meta.refinery annotation.

### DIFF
--- a/collect/collect.go
+++ b/collect/collect.go
@@ -411,7 +411,11 @@ func mergeTraceAndSpanSampleRates(sp *types.Span, traceSampleRate uint) {
 	}
 
 	if sp.SampleRate < 1 {
-		// Don't pass along negative/zero sample rates
+		// See https://docs.honeycomb.io/manage-data-volume/sampling/
+		// SampleRate is the denominator of the ratio of sampled spans
+		// HoneyComb treats a missing or 0 SampleRate the same as 1, but
+		// behaves better/more consistently if the SampleRate is explicitly
+		// set instead of inferred
 		sp.SampleRate = 1
 	}
 
--- a/collect/collect.go
+++ b/collect/collect.go
@@ -403,13 +403,18 @@ func (i *InMemCollector) dealWithSentTrace(keep bool, sampleRate uint, sp *types
 }
 
 func mergeTraceAndSpanSampleRates(sp *types.Span, traceSampleRate uint) {
-	if traceSampleRate < 1 {
+	if traceSampleRate != 1 {
 		// When the sample rate from the trace is not 1 that means we are
 		// going to mangle the span sample rate. Write down the original sample
 		// rate so that that information is more easily recovered
 		sp.Data["meta.refinery.original_sample_rate"] = sp.SampleRate
+	}
+
+	if sp.SampleRate < 1 {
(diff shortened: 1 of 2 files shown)


MikeGoldsmith (changes requested):
Thanks for the PR @epvanhouten 👍🏻 

I think we still need to update the original sample rate if unset or negative, otherwise looks good 👍🏻

collect/collect.go
```diff
@@ -395,16 +395,25 @@ func (i *InMemCollector) dealWithSentTrace(keep bool, sampleRate uint, sp *types
...
-		sp.SampleRate *= sampleRate
+		mergeTraceAndSpanSampleRates(sp, sampleRate)
 		i.Transmission.EnqueueSpan(sp)
 		return
 	}
 	i.Logger.Debug().WithField("trace_id", sp.TraceID).Logf("Dropping span because of previous decision to drop trace")
 }
 
+func mergeTraceAndSpanSampleRates(sp *types.Span, traceSampleRate uint) {
+	if traceSampleRate != 1 {
```
MikeGoldsmith, line 406: Missed adding it initial review, this is where I think we should default sample rate if negative or 0.

```suggestion
	if sp.SampleRate < 1 {
		sp.SampleRate = 1
	}
	sp.SampleRate *= sampleRate	if traceSampleRate != 1 {
	if traceSampleRate != 1 {
```

epvanhouten: I updated `mergeTraceAndSpanSampleRates` to set to 1 when something less than 1 and added a unit test to capture that behavior. I very briefly tried to find mention of the range of allowed values for SampleRate in documentation for OpenTelemetry but came up empty. Is there a source for that requirement so we can leave a breadcrumb of why?

kentquirk: Honeycomb's use and interpretation of SampleRate predates OTel. OTel uses a p-value, which is effectively the mathematical inverse of Honeycomb's SampleRate. 

Honeycomb's backend data system interprets a missing value for SampleRate as a value of 0, so therefore converts those to 1 automatically so the math works correclty. We have been trying to push that idea upstream where we can; it benefits everyone to have it more explicit.

epvanhouten:
Gave me enough breadcrumbs to find...

https://opentelemetry.io/docs/reference/specification/trace/tracestate-probability-sampling/
https://docs.honeycomb.io/getting-data-in/opentelemetry/dotnet-distro/#sampling

So SampleRate is explicitly not an OTEL construct, but honeycombs
SampleRate which says probability of sample = 1/SampleRate where SampleRate
is 1 to max(uint)?

kentquirk (approved):
This is excellent. Thank you for your contribution!

MikeGoldsmith (approved):

PR #509, opened 2022-09-09
allow BatchTimeout to be overriden on the libhoney Transmission

leviwilson:

We have an issue where our ALB traffic has a `~5m` latency because of the cadence in which the CloudWatch logs are sent. Because the ALB represents our top-level traces, this means that _all_ web traffic is delayed by that same amount.

Decreasing the `TraceTimeout` value assists with this by spreading out the traffic, however this makes our rules a moot point as things we are intentionally trying to downsample do not get downsampled.

What we _really_ need is a manner to control how frequently events are sent upstream to Honeycomb so we can control how frequently we send things. https://github.com/honeycombio/libhoney-go/blob/8af08e456cfe1f786247ffdc755855999add2d76/libhoney.go#L136 indicates that `SendFrequency` can override how often batches are sent.

This PR allows for the `BatchTimeout` to be configured for refinery rather than explicitly using `DefaultBatchTimeout`. I realize `SendFrequency` is what was mentioned above, but refinery calls `NewClient` rather than `Init` and you can see that all `SendFrequency` does is drive `BatchTimeout` on [the `Transmission` anyway](https://github.com/honeycombio/libhoney-go/blob/8af08e456cfe1f786247ffdc755855999add2d76/libhoney.go#L236).

### DIFF
--- a/cmd/refinery/main.go
+++ b/cmd/refinery/main.go
@@ -138,7 +138,7 @@ func main() {
 	upstreamClient, err := libhoney.NewClient(libhoney.ClientConfig{
 		Transmission: &transmission.Honeycomb{
 			MaxBatchSize:          c.GetMaxBatchSize(),
-			BatchTimeout:          libhoney.DefaultBatchTimeout,
+			BatchTimeout:          c.GetBatchTimeout(),
 			MaxConcurrentBatches:  libhoney.DefaultMaxConcurrentBatches,
 			PendingWorkCapacity:   uint(c.GetUpstreamBufferSize()),
 			UserAgentAddition:     userAgentAddition,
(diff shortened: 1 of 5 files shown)


MikeGoldsmith (changes requested):
Looks great, thanks @leviwilson 🎉 

Only thing missing is adding an example to the [config_config.toml](https://github.com/honeycombio/refinery/blob/main/config_complete.toml#L59) to show it being overriden.

leviwilson:
> Looks great, thanks @leviwilson 🎉
> 
> Only thing missing is adding an example to the [config_config.toml](https://github.com/honeycombio/refinery/blob/main/config_complete.toml#L59) to show it being overriden.

done

jessitron:
Thanks @leviwilson !

@MikeGoldsmith fyi someday there's a more interesting problem to solve here, of refinery noticing Honeycomb's rate-limiting errors and adapting.

This change will go a long way toward letting people solve this problem themselves now.

MikeGoldsmith (approved):
Thanks @leviwilson 🎉

PR #511, opened 2022-09-12
Consolidate honeycomb metrics to use single lock & fix concurrent read/write

### LINKED ISSUE #512: Honeycomb metrics panics with concurrent read/write error
**Versions**

- Go: 1.18
- Refinery: 1.15.0

**Steps to reproduce**
Running refinery for an extended time period using Honeycomb metrics can run into this issue. The problem is a mutex lock is used to register and collect metrics but the functions used to update metrics values doesn't use the generic locks, which can lead to a concurrent read / write error.

For example, the [Counter](https://github.com/honeycombio/refinery/blob/main/metrics/honeycomb.go#L323) func here doesn't acquire the lock before attempting to retrieve a metric from the registers counters map.

MikeGoldsmith:
Honeycomb metrics currently utilises 3 locks, one for each map of counters, gauges and histograms and is unnecessary. In addition, the locks were only used when registering each metric and collecting metrics values - they were missing the lock to retrieve an entry for updating which risks a read / write error.

This PR updates honeycomb metrics to utilise a single lock that is used to control all three types of metrics and is also used when attempting to retrieve a metric from a map.

- Fixes #512 

- Consolidate the 3 locks into a single lock
- Use lock when attempting to retrieve an existing metric from one of the maps to prevent concurrent read / write panics

### DIFF
--- a/metrics/honeycomb.go
+++ b/metrics/honeycomb.go
@@ -295,7 +295,7 @@ func getOrAdd[T *counter | *gauge | *histogram](lock *sync.RWMutex, name string,
 	// attempt to get metric by name using read lock
 	lock.RLock()
 	metric, ok := metrics[name]
-	lock.Unlock()
+	lock.RUnlock()
 
 	// if found, return existing metric
 	if ok {
--- a/metrics/honeycomb.go
+++ b/metrics/honeycomb.go
@@ -276,59 +276,71 @@ func average(vals []float64) float64 {
 }
 
 func (h *HoneycombMetrics) Register(name string, metricType string) {
-	h.lock.Lock()
-	defer h.lock.Unlock()
-
 	switch metricType {
 	case "counter":
-		// inside the lock, let's not race to create the counter
-		_, ok := h.counters[name]
-		if !ok {
-			newCounter := &counter{
-				name: name,
-			}
-			h.counters[name] = newCounter
-		}
+		getOrAdd(&h.lock, name, h.counters, createCounter)
 	case "gauge":
-		_, ok := h.gauges[name]
-		if !ok {
-			newGauge := &gauge{
-				name: name,
-			}
-			h.gauges[name] = newGauge
-		}
+		getOrAdd(&h.lock, name, h.gauges, createGague)
 	case "histogram":
-		_, ok := h.histograms[name]
-		if !ok {
-			newGauge := &histogram{
-				name: name,
-				vals: make([]float64, 0),
-			}
-			h.histograms[name] = newGauge
-		}
+		getOrAdd(&h.lock, name, h.histograms, createHistogram)
 	default:
 		h.Logger.Debug().Logf("unspported metric type %s", metricType)
 	}
 }
 
-func (h *HoneycombMetrics) Count(name string, n interface{}) {
-	// attempt to retrieve existing counter, with read lock
-	h.lock.RLock()
-	counter, ok := h.counters[name]
-	h.lock.RUnlock()
+// getOrAdd attempts to retrieve a (generic) metric from the provided map by name, wrapping the read operation
+// with a read lock (RLock). If the metric is not present in the map, it aquire a write lock and executes
+// create function and add it's to the map.
+func getOrAdd[T *counter | *gauge | *histogram](lock *sync.RWMutex, name string, metrics map[string]T, createMetric func(name string) T) T {
+	// attempt to get metric by name using read lock
+	lock.RLock()
+	metric, ok := metrics[name]
+	lock.Unlock()
 
-	if !ok {
-		// execute register outside of read lock so it can take write lock
-		h.Register(name, "counter")
-
-		// attempt to get counter again, with read lock
-		h.lock.RLock()
-		counter = h.counters[name]
-		h.lock.RUnlock()
+	// if found, return existing metric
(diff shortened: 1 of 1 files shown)


metrics/honeycomb.go
```diff
@@ -228,26 +226,23 @@ func (h *HoneycombMetrics) reportToHoneycommb(ctx context.Context) {
 				"api_host": ev.APIHost,
 				"dataset":  ev.Dataset,
 			}
-			h.countersLock.Lock()
```
vreynolds, line 231: removing individual locks would make the entire function block for the whole reporting loop -- would that slow things down?

metrics/honeycomb.go
```diff
@@ -23,12 +23,10 @@ type HoneycombMetrics struct {
 	UpstreamTransport *http.Transport `inject:"upstreamTransport"`
 	Version           string          `inject:"version"`
 
-	countersLock   sync.Mutex
-	counters       map[string]*counter
-	gaugesLock     sync.Mutex
-	gauges         map[string]*gauge
-	histogramsLock sync.Mutex
-	histograms     map[string]*histogram
+	lock       sync.Mutex
```
vreynolds, line 26: this should be a RWLock

metrics/honeycomb.go
```diff
@@ -281,10 +276,11 @@ func average(vals []float64) float64 {
 }
 
 func (h *HoneycombMetrics) Register(name string, metricType string) {
+	h.lock.Lock()
```
vreynolds, line 279: there is a lock acquired in the parent function. I _think_ if parent does a Read lock and here it does a Write lock, it will not deadlock, but needs to be verified

vreynolds: this was helpful in my past journey with deadlocks: https://github.com/sasha-s/go-deadlock

metrics/honeycomb.go
MikeGoldsmith: I don't think it will have much of an affect. Most of the time we expect to be using read locks only (collect / updating existing metric values).

We use a write lock when adding new metrics (register).

metrics/honeycomb.go
MikeGoldsmith: Yup, you're right. I've refactored to use a RWLock and made sure we release the read lock before attempting to get a write lock.

metrics/honeycomb.go
```diff
@@ -283,75 +278,89 @@ func average(vals []float64) float64 {
...
-func (h *HoneycombMetrics) Count(name string, n interface{}) {
-	count, ok := h.counters[name]
+// getOrAdd attempts to retrieve a (generic) metric from the provided map by name, wrapping the read operation
+// with a read lock (RLock). If the metric is not present in the map, it aquire a write lock and executes
+// create function and add it's to the map.
+func getOrAdd[T *counter | *gauge | *histogram](lock *sync.RWMutex, name string, metrics map[string]T, createMetric func(name string) T) T {
+	// attempt to get metric by name using read lock
+	lock.RLock()
+	metric, ok := metrics[name]
+	lock.Unlock()
```
vreynolds, line 298:
```suggestion
	lock.RUnlock()
```

vreynolds (changes requested):
LGTM, one small change for the read unlock

vreynolds (approved):
🔒 🎇

metrics/honeycomb.go
```diff
@@ -283,75 +278,89 @@ func average(vals []float64) float64 {
...
+	return metric
+}
+
+func createCounter(name string) *counter {
+	return &counter{
+		name: name,
+	}
+}
+
+func createGague(name string) *gauge {
```
JamieDanielson, line 324:
```suggestion
func createGauge(name string) *gauge {
```
drive-by sorry! Suggest Rename Symbol

metrics/honeycomb.go
```diff
@@ -283,75 +278,89 @@ func average(vals []float64) float64 {
...
 	h.Count(name, 1)
 }
 
 func (h *HoneycombMetrics) Gauge(name string, val interface{}) {
-	gauge, ok := h.gauges[name]
-	if !ok {
-		h.Register(name, "gauge")
-		gauge = h.gauges[name]
-	}
+	gauge := getOrAdd(&h.lock, name, h.gauges, createGague)
```
JamieDanielson, line 351:
```suggestion
	gauge := getOrAdd(&h.lock, name, h.gauges, createGauge)
```
(with function rename)

metrics/honeycomb.go
```diff
@@ -283,75 +278,89 @@ func average(vals []float64) float64 {
...
-		h.gaugesLock.Lock()
-		defer h.gaugesLock.Unlock()
-		_, ok := h.gauges[name]
-		if !ok {
-			newGauge := &gauge{
-				name: name,
-			}
-			h.gauges[name] = newGauge
-		}
+		getOrAdd(&h.lock, name, h.gauges, createGague)
```
JamieDanielson, line 283:
```suggestion
		getOrAdd(&h.lock, name, h.gauges, createGauge)
```
(with function rename)

JamieDanielson (approved):
This looks great! I left a few tiny suggestions as I went through, mostly just for clarification 👍

PR #514, opened 2022-09-15
Allow adding extra fields to error logs

### LINKED ISSUE #513: Attach additional metadata to make ingestion error logs more informative

If refinery is sent a span that exceeds 100K bytes, it is rejected before being sent, as Honeycomb will not accept spans that big. 

Unfortunately, there is no easy way to figure out which span caused the problem. 

This is because refinery moves spans around -- the point at which a trace is rejected for being too big is not directly tied to the time and place that it is sent.

Furthermore, it sends them with libhoney, which is enqueueing them for sending later, so even the enqueueing doesn't know whether it was acceptable.

Finally, the data is never transmitted to Honeycomb (since libhoney knew it was too big), so we never even see it in our telemetry.

However, libhoney does decorate the error message with "metadata" that was attached to the event by refinery when it was created. If we add certain fields from the span that caused the error to the error logs, we could have better ability to identify it. 

There could be a refinery configuration to control which fields to use (an array of field names), and I think the most sensible default for that would be just the single field `trace.span_id` (if it exists). 

if you get one of these errors, it would include the span id that caused the error, and you could at least go searching through logs to try to find it.

But if, say, you always decorated your spans with source code information, you could use the source filename instead.

I have a rough proof of concept of this (hard-coded span names, no error checking) working on my local box.

kentquirk:

As noted in #513, errors that occur while Refinery is trying to send data to Honeycomb happen asynchronously. It's hard to relate any errors that show up in the logs with the spans that caused them to occur. However, there is a way provided by libhoney ([metadata](https://github.com/honeycombio/libhoney-go/blob/25068939fe8240ccc45e4c86271b6f057c5f833d/transmission/response.go#L11)) that makes it possible to attach data to the error. Before now, Refinery has attached the dataset, api host, and environment to the metadata. This helps but is not enough.

This PR attempts to leverage that technique further by adding a new configuration value called `AdditionalErrorFields`. It allows the user to specify a list of field names. In the event of a transmission error, these fields (if they exist) will be copied from the failing span into the metadata, and will therefore show up as identified fields in the logs. 

The default value is `trace.span_id`.

This also removes the remaining instances of the obsolete standard library `ioutil`.

Closes #513

### DIFF
--- a/app/app_test.go
+++ b/app/app_test.go
@@ -362,7 +362,7 @@ func TestPeerRouting(t *testing.T) {
 			"long":            "this is a test of the emergency broadcast system",
 			"foo":             "bar",
 		},
-		Metadata: map[string]string{
+		Metadata: map[string]any{
 			"api_host":    "http://api.honeycomb.io",
 			"dataset":     "dataset",
 			"environment": "",
@@ -486,7 +486,7 @@ func TestEventsEndpoint(t *testing.T) {
 				"trace.trace_id": "1",
 				"foo":            "bar",
 			},
-			Metadata: map[string]string{
+			Metadata: map[string]any{
 				"api_host":    "http://api.honeycomb.io",
 				"dataset":     "dataset",
 				"environment": "",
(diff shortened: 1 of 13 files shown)


MikeGoldsmith (approved):
Looks good 👍🏻

MikeGoldsmith:
PS This needs a title update and labels applying.

PR #521, opened 2022-09-19
Cloud support

Baliedge:
When hosting Refinery in a cloud environment, certain network assumptions are no longer true. 
 Issues arise from these assumptions while hosting in our cloud environment using [Nomad](https://www.nomadproject.io/) as container orchestration.

Terms, as used in Nomad:
- Task: Single deployed container.
- Group: Group of tasks running as a scalable service (e.g. Refinery).

This assumption is not necessarily true under Nomad:
- The IP assigned to a task's `eth0` can be accessed by other tasks in the group.

Redis peering expects peers to talk to each other using their `eth0` IP.  This behavior can be overridden by specifying `RedisIdentifier` in config.  This value will be advertised as the peer's IP in the peer list.

However, on startup, peering uses its `eth0` IP to cross-reference in the peers list found in Redis.  Since the internal `eth0` IP is not the true public IP (set by `RedisIdentifier`), this check will always fail and the container exits in error.

A workaround is to treat `RedisIdentifier` as the true public IP address of the task.  Thus, it should be one of the "self" IPs when checking the peers list.

This behavior is reinforced in existing code in [`internal/peer/redis.go`](https://github.com/honeycombio/refinery/blob/main/internal/peer/redis.go#L291-L298) where `RedisIdentifier` is used to create a URL with the value being the host.

### DIFF
--- a/sharder/deterministic.go
+++ b/sharder/deterministic.go
@@ -128,12 +128,11 @@ func (d *DeterministicSharder) Start() error {
 			localAddrCidrs[i] = addr.String()
 		}
 
-		// If RedisIdentifier is an IP, add it to localAddrs.
+		// If RedisIdentifier is an IP, add it to localAddrCidrs.
 		redisIdentifier, err := d.Config.GetRedisIdentifier()
-		if err == nil {
-			d.Logger.Debug().Logf("Using RedisIdentifier as self: %s", redisIdentifier)
+		if err == nil && redisIdentifier != "" {
 			if ip := net.ParseIP(redisIdentifier); ip != nil {
-				d.Logger.Debug().Logf("RedisIdentifier is an IP")
+				d.Logger.Debug().Logf("Using RedisIdentifier as public IP: %s", redisIdentifier)
 				localAddrCidrs = append(localAddrCidrs, fmt.Sprintf("%s/32", redisIdentifier))
 			}
 		}
--- a/sharder/deterministic.go
+++ b/sharder/deterministic.go
@@ -118,26 +118,30 @@ func (d *DeterministicSharder) Start() error {
 		}
 		d.Logger.Debug().Logf("picked up local peer port of %s", localPort)
 
-		// get my local interfaces' IPs.
-		localAddrs, err := net.InterfaceAddrs()
-		if err != nil {
-			return errors.Wrap(err, "failed to get local interface list to initialize sharder")
-		}
-		localIPs := make([]string, len(localAddrs))
-		for i, addr := range localAddrs {
-			addrStr := addr.String()
-			ip, _, err := net.ParseCIDR(addrStr)
-			if err != nil {
-				return errors.Wrap(err, fmt.Sprintf("failed to parse CIDR for local IP %s", addrStr))
-			}
-			localIPs[i] = ip.String()
-		}
+		var localIPs []string
 
-		// If RedisIdentifier is an IP, add it to localIps.
+		// If RedisIdentifier is an IP, use as localIPs value.
 		if redisIdentifier, err := d.Config.GetRedisIdentifier(); err == nil && redisIdentifier != "" {
 			if ip := net.ParseIP(redisIdentifier); ip != nil {
 				d.Logger.Debug().Logf("Using RedisIdentifier as public IP: %s", redisIdentifier)
-				localIPs = append(localIPs, redisIdentifier)
+				localIPs = []string{redisIdentifier}
+			}
+		}
+
+		// Otherwise, get my local interfaces' IPs.
+		if len(localIPs) == 0 {
+			localAddrs, err := net.InterfaceAddrs()
+			if err != nil {
+				return errors.Wrap(err, "failed to get local interface list to initialize sharder")
+			}
+			localIPs = make([]string, len(localAddrs))
+			for i, addr := range localAddrs {
+				addrStr := addr.String()
+				ip, _, err := net.ParseCIDR(addrStr)
+				if err != nil {
+					return errors.Wrap(err, fmt.Sprintf("failed to parse CIDR for local IP %s", addrStr))
+				}
+				localIPs[i] = ip.String()
 			}
 		}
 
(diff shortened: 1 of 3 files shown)


sharder/deterministic.go
```diff
@@ -122,28 +123,50 @@ func (d *DeterministicSharder) Start() error {
 		if err != nil {
 			return errors.Wrap(err, "failed to get local interface list to initialize sharder")
 		}
+		localAddrCidrs := make([]string, len(localAddrs))
+		for i, addr := range localAddrs {
+			localAddrCidrs[i] = addr.String()
+		}
+
+		// If RedisIdentifier is an IP, add it to localAddrs.
```
kentquirk, line 131: Does it actually make sense to add it to `localCidrAddrs` (typo there), or would it be better to simply replace the contents of `localCidrAddrs` with a single value -- the one specified by a non-empty `GetRedisIdentifier()`?

sharder/deterministic.go
```diff
@@ -122,28 +123,50 @@ func (d *DeterministicSharder) Start() error {
...
+		redisIdentifier, err := d.Config.GetRedisIdentifier()
+		if err == nil {
```
kentquirk, line 133: It's clearer, I think, if we don't even try if this wasn't specified.
```suggestion
		if err == nil && redisIdentifier != "" {
```

sharder/deterministic.go
```diff
@@ -122,28 +123,50 @@ func (d *DeterministicSharder) Start() error {
...
+			d.Logger.Debug().Logf("Using RedisIdentifier as self: %s", redisIdentifier)
+			if ip := net.ParseIP(redisIdentifier); ip != nil {
+				d.Logger.Debug().Logf("RedisIdentifier is an IP")
+				localAddrCidrs = append(localAddrCidrs, fmt.Sprintf("%s/32", redisIdentifier))
```
kentquirk, line 137: Assumes that RedisIdentifier is IPv4. That's maybe not a valid assumption? Should this Sprintf be conditional? And should it be smarter than just using `/32`?

kentquirk (commented):
Still a draft, I know, but I thought I'd weigh in since you posted on Pollinators.

sharder/deterministic.go
Baliedge: I didn't know if I could assume `err == nil` means there exists a non-empty value.  If that's so then I will undo.

Baliedge: Agreed.  I made this change.

sharder/deterministic.go
Baliedge: Fixed.

sharder/deterministic.go
Baliedge: The network portion isn't used in the following logic.  It's just extraneous data from `net.InterfaceAddrs()` that has to be replicated by this logic.  Maybe the solution is to strip the network portion earlier?  I'll give it a shot.

Baliedge: There, take a look now.

sharder/deterministic.go
kentquirk: Thanks -- I should have split this into two questions. If we use RedisIdentifier, I don't think we need to search the entire list of IPs we get by querying the interface. We should *only* find RedisIdentifier anyway -- so instead of appending it to the list, we could replace the whole thing with a list containing one element.

Or am I missing something?

Baliedge: I took the minimal code change approach to accomplish the desired functionality.  I admit not understanding why this self lookup is necessary.  Rather than break whatever those use cases are, I just expanded on it.

If it is decided to replace the list, then I'll make that change.  Just please confirm.

sharder/deterministic.go
```diff
@@ -117,33 +118,54 @@ func (d *DeterministicSharder) Start() error {
...
+			addrStr := addr.String()
+			ip, _, err := net.ParseCIDR(addrStr)
+			if err != nil {
+				return errors.Wrap(err, fmt.Sprintf("failed to parse CIDR for local IP %s", addrStr))
+			}
+			localIPs[i] = ip.String()
+		}
+
+		// If RedisIdentifier is an IP, add it to localIps.
+		if redisIdentifier, err := d.Config.GetRedisIdentifier(); err == nil && redisIdentifier != "" {
```
kentquirk, line 137: Yes, as can be seen in `redis.go`, if `RedisIdentifier` is specified then it will be the peer address shown, and you can safely replace the entire contents of the list with the single item. Please do that. 

I'd also look favorably on inverting the logic within the publicAddr function so that if redisIdentifier is specified, it doesn't go through all of the other logic. If you felt up to adding that here, it would make sense. Let me know if not.

sharder/deterministic.go
```diff
@@ -117,33 +118,54 @@ func (d *DeterministicSharder) Start() error {
...
-						return errors.Wrap(err, fmt.Sprintf("failed to parse CIDR for local IP %s", localIP.String()))
-					}
-					if peerIP == ipAddr.String() {
+				for _, ipAddr := range localIPs {
+					if peerIP == ipAddr {
 						if peerShard.port == localPort {
 							d.Logger.Debug().WithField("peer", peerShard).Logf("Found myself in peer list")
 							found = true
 							selfIndexIntoPeerList = i
+						} else {
```
kentquirk, line 164: Did you find a case where this happens? I'm curious as to why you added this.

kentquirk (commented):
Almost there.

sharder/deterministic.go
Baliedge: Yes, and it was ultimately a misconfiguration.  The `RedisIdentifier` port and `PeerListenAddr` port must be identical.  This required configuring static ports in Nomad, where randomly generated port mappings are the default.

sharder/deterministic.go
Baliedge: Done, both parts.

kentquirk (approved):
I apologize for the delay. I was out sick for several days.

This looks good! Thank you.

MikeGoldsmith:
@Baliedge does this supersede the the etcd PR?
- #506

Baliedge:
> @Baliedge does this supersede the the etcd PR?
> 
> * [Added PeerManagement.Type = 'etcd' #506](https://github.com/honeycombio/refinery/pull/506)

No.  I would say that's a separate effort.  We decided to stick to Redis mode for now.  Moving to etcd was determined to be a nice to have.  We may decide to develop that later if necessary.

TylerHelmuth (approved):
LGTM

PR #523, opened 2022-09-27
Properly set metadata to values that will work.

kentquirk:

- A previous PR (#514) changed the type of the metadata object from `map[string]string` to `map[string]any`, but I must have forgotten to hit save at some point, and this part of it (4 lines) didn't get committed. This bug actually will cause that feature to omit fields it knows about.

- Change the type appropriately in processResponses


TylerHelmuth (approved):
LGTM

MikeGoldsmith (approved):

PR #532, opened 2022-10-03
Track span count and optionally add it to root

### LINKED ISSUE #483: Apply sample rates based on span count
Related to this pollinators [slack thread](https://honeycombpollinators.slack.com/archives/C01BZ3J7RMJ/p1659047633507579).
> It would be really cool if we could...weight traces by how many spans they have rather than using span length as a key in refinery. Like, if a trace has over 1k spans, it now counts as if it were two traces for determining sample rate.
> I bring this up because we have some just-infrequent enough web requests to not get sampled by Refinery that are...sadly a couple thousand spans and it'd be nice to be able to force refinery's hand a bit there.

A way to specify a sample rate based on the # of spans in a trace.

**Additional context**
A special type of rule or rule condition could be created based on the # of spans in the trace.  Rule scope would affect how this would get evaluated.

### LINKED ISSUE #524: Add meta.span_count as a meta value, and allow it to be queried in rules
There have been several requests for refinery rules that would include span count. We do allow the EMA sampler to include span count as a key, but there are use cases for more subtle rules as well. 

For example, a rule could take very large traces and sample them more aggressively.

Keep `meta.span_count` on root spans and make sure it's accessible to Rules for conditions.

Sometimes people want to do this, and there's no great alternative right now. 

**Additional context**

Closes #483

kentquirk:

Adds the ability to add the trace's span count to the root span of the trace, and does it in a way that allows the rules engine to query it as `meta.span_count`. This would allow rules that depend on span count -- for example, a rule that adjusts sample rate so that larger traces are kept less often.

Note that this value is not available if the root span hasn't arrived when the TraceTimeout completes; at this point, the rules will be evaluated without a `span_count` because there is no root span to put it on. 

If the root span arrives late, it will be updated with the total number of spans that have arrived at that point, so that the value stored in Honeycomb is close to correct. If more spans arrive later, they will not update the value in the root span.

Closes #524  
Closes #483

### DIFF
--- a/collect/collect_test.go
+++ b/collect/collect_test.go
@@ -723,3 +723,72 @@ func TestDependencyInjection(t *testing.T) {
 		t.Error(err)
 	}
 }
+
+// TestAddSpanCount tests that adding a root span winds up with a trace object in
+// the cache and that that trace gets span count added to it
+func TestAddSpanCount(t *testing.T) {
+	transmission := &transmit.MockTransmission{}
+	transmission.Start()
+	conf := &config.MockConfig{
+		GetSendDelayVal:    0,
+		GetTraceTimeoutVal: 60 * time.Second,
+		GetSamplerTypeVal:  &config.DeterministicSamplerConfig{SampleRate: 1},
+		SendTickerVal:      2 * time.Millisecond,
+		AddSpanCountToRoot: true,
+	}
+	coll := &InMemCollector{
+		Config:       conf,
+		Logger:       &logger.NullLogger{},
+		Transmission: transmission,
+		Metrics:      &metrics.NullMetrics{},
+		SamplerFactory: &sample.SamplerFactory{
+			Config: conf,
+			Logger: &logger.NullLogger{},
+		},
+	}
+	c := cache.NewInMemCache(3, &metrics.NullMetrics{}, &logger.NullLogger{})
+	coll.cache = c
+	stc, err := lru.New(15)
+	assert.NoError(t, err, "lru cache should start")
+	coll.sentTraceCache = stc
+
+	coll.incoming = make(chan *types.Span, 5)
+	coll.fromPeer = make(chan *types.Span, 5)
+	coll.datasetSamplers = make(map[string]sample.Sampler)
+	go coll.collect()
+	defer coll.Stop()
+
+	var traceID = "mytrace"
+
+	span := &types.Span{
+		TraceID: traceID,
+		Event: types.Event{
+			Dataset: "aoeu",
+			Data: map[string]interface{}{
+				"trace.parent_id": "unused",
+			},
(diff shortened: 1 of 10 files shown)


sample/rules.go
```diff
@@ -154,21 +154,19 @@ func ruleMatchesSpanInTrace(trace *types.Trace, rule *config.RulesBasedSamplerRu
...
-				matchCount++
+			if !conditionMatchesValue(condition, value, exists) {
+				ruleMatched = false
+				break // if any condition fails, we can't possibly succeed, so exit inner loop early
 			}
 		}
 		// If this span was matched by every condition, then the rule as a whole
 		// matches (and we can return)
-		if matchCount == len(rule.Condition) {
+		if ruleMatched {
```
TylerHelmuth, line 169: Are these changes related to span count or are they efficiencies for this loop?

collect/collect_test.go
```diff
@@ -723,3 +723,72 @@ func TestDependencyInjection(t *testing.T) {
 		t.Error(err)
 	}
 }
+
+// TestAddSpanCount tests that adding a root span winds up with a trace object in
+// the cache and that that trace gets span count added to it
+func TestAddSpanCount(t *testing.T) {
```
TylerHelmuth, line 729: Is it possible to add test cases for late spans?

sample/rules.go
kentquirk: I realized that the loop was pretty unclear as written and that it could be made a lot simpler. It's not about performance (although it's probably marginally faster) but about comprehension.

It didn't feel big enough to me to justify its own PR, but we could pull it out if it bugs you.

collect/collect_test.go
kentquirk: Done.

sample/rules_test.go
```diff
@@ -496,6 +496,48 @@ func TestRules(t *testing.T) {
...
+							"trace.span_id":   "654321",
+							"trace.parent_id": "54321",
+							"test":            int64(2),
+						},
+					},
+				},
+			},
+			ExpectedName: "Check root span for span count",
+			ExpectedKeep: true,
+			ExpectedRate: 1,
```
TylerHelmuth, line 539: Is 1 also the default sample rate?  If something went wrong with the rule and the default value was used instead (incorrectly) would the test pass?

TylerHelmuth (approved):
LGTM

types/event.go
```diff
@@ -68,6 +68,11 @@ func (t *Trace) GetSpans() []*Span {
 	return t.spans
 }
 
+// SpanCount gets the number of spans currently in this trace as int64
+func (t *Trace) SpanCount() int64 {
+	return int64(len(t.spans))
```
seh, line 73: See #555 for correcting the reported value.

PR #535, opened 2022-10-03
Add support for metrics api key env var

### LINKED ISSUE #520: Allow configuration of metrics api key via environment variable

The current code assumes the user is going to use the same API key for both the events and the metrics since they share an environment variable string here: 

https://github.com/honeycombio/refinery/blob/9ef4ff8d5c6225a2867c5e32b9e2bfb32a5d503f/config/file_config.go#L121-L122

create a separate environment variable for the metrics api key so it can go to a different team. Fallback to the prior behavior (same key) if not set. 

setting the variable in a file is less secure. 

**Additional context**

TylerHelmuth:

Allows setting `HoneycombMetrics.MetricsAPIKey` with a specific environment variable.

- Fixes #520

- Utilizes viper's BindEnv() function to handle env var precedence for `HoneycombMetrics.MetricsAPIKey`.
- If `REFINERY_HONEYCOMB_METRICS_API_KEY` is set then it will be used.  Otherwise `REFINERY_HONEYCOMB_API_KEY` will be used.  As a result existing customer configs will not be affected.
- Added new unit tests for this scenario.


config/file_config.go
```diff
@@ -119,7 +119,7 @@ func NewConfig(config, rules string, errorCallback func(error)) (Config, error)
 	c.BindEnv("PeerManagement.RedisUsername", "REFINERY_REDIS_USERNAME")
 	c.BindEnv("PeerManagement.RedisPassword", "REFINERY_REDIS_PASSWORD")
 	c.BindEnv("HoneycombLogger.LoggerAPIKey", "REFINERY_HONEYCOMB_API_KEY")
-	c.BindEnv("HoneycombMetrics.MetricsAPIKey", "REFINERY_HONEYCOMB_API_KEY")
+	c.BindEnv("HoneycombMetrics.MetricsAPIKey", "REFINERY_HONEYCOMB_METRICS_API_KEY", "REFINERY_HONEYCOMB_API_KEY")
```
kentquirk, line 122: dang, today I learned that BindEnv takes multiple arguments. Now I need to go look at other uses.

kentquirk (approved):
I love the thorough test cases.

PR #542, opened 2022-10-25
Add query endpoint that retrieves the running configuration

mterhar:

- The query endpoint can retrieve the rules but does not include the configuration

- This adds a new `/query` endpoint called `config` which pulls the running configuration the same way the rules endpoint pulls the rules out of the rules file.

### DIFF
--- a/config/file_config.go
+++ b/config/file_config.go
@@ -554,13 +554,20 @@ func (f *fileConfig) GetRunningConfig() (map[string]interface{}, error) {
 	configs := make(map[string]interface{})
 	keys := f.config.AllKeys()
 	for _, key := range keys {
-		configs[key] = f.config.GetString(key)
-		// mask sensitive strings
-		maskKeys := []string{"key", "password", "token"}
-		for _, s := range maskKeys {
-			if(strings.Contains(strings.ToLower(key), s)) {
-				configs[key] = strings.Repeat("x", len(f.config.GetString(key)))
-			}	
+		parts := strings.Split(key, ".")
+		if len(parts) == 1 {
+			// lacking a period means it's a root level config
+			configs[key] = f.config.GetString(key)
+		} else {
+			// having a period means it's a multi-level config
+			m := make(map[string]interface{})
+			if sub := f.config.Sub(parts[0]); sub != nil {
+				err := sub.Unmarshal(&m)
+				if err != nil {
+					return nil, fmt.Errorf("failed to unmarshal config section %s: %w", parts[0], err)
+				}
+			}
+			configs[parts[0]] = m
 		}
 	}
 	return configs, nil
--- a/config/file_config.go
+++ b/config/file_config.go
@@ -550,6 +550,22 @@ func (f *fileConfig) GetCollectorType() (string, error) {
 	return f.conf.Collector, nil
 }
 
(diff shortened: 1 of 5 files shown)


config/file_config.go
```diff
@@ -550,6 +550,22 @@ func (f *fileConfig) GetCollectorType() (string, error) {
 	return f.conf.Collector, nil
 }
 
+func (f *fileConfig) GetRunningConfig() (map[string]interface{}, error) {
+	configs := make(map[string]interface{})
+	keys := f.config.AllKeys()
+	for _, key := range keys {
+		configs[key] = f.config.GetString(key)
+		// mask sensitive strings
+		maskKeys := []string{"key", "password", "token"}
```
TylerHelmuth, line 559: Are we sure this is exhaustive?  Are there any free-form config options a user could use that might contain sensitive information?

mterhar: These are all the ones that will be unmarshalled by the file_config reader. 

I'm iffy on whether we want to bother masking them since the endpoint is authenticated and making sure a list of keys was properly updated may be a valid use case.

mterhar: If someone adds an arbitrary config line like

```toml
SecretFormula = "add cajun at the end"
```

Refinery errors out like this: 

```console
unable to load config: 1 error(s) decoding:

* '' has invalid keys: secretformula
Process 102407 has exited with status 1
```

MikeGoldsmith, line 559: I agree, as the endpoint is authenticated and requires direct access to the cluster, I think it would be okay to not mask those values.

However, if you do decide to mask these values, you could use a `switch key.ToLower()` in place of creating an array and looping over it.

mterhar: I wrote it that way first but it isn't a perfect match so I'm not sure what goes in the 'case' to handle "contains".

kentquirk:
Hi, Mike.

I considered this feature when I added the original /config endpoints and left it out deliberately, as it's a pretty big security hole, so I'm reluctant to include it.

Is there a particular problem you are trying to solve with it?

mterhar:
> Hi, @mterhar.
> 
> I considered this feature when I added the original /config endpoints and left it out deliberately, as it's a pretty big security hole, so I'm reluctant to include it.
> 
> Is there a particular problem you are trying to solve with it?

Same problem as with the rules file changes. Different deployment variables make it very difficult to know when a configuration change has been accepted. The config reloads are supposed to happen frequently but somehow seem not to. This will allow people to confirm the running config is what it's supposed to be. It's also important if a refinery operator is to exist, so the operator pod can inspect the refinery pods to ensure configurations are updated and kill any misbehaving pods.

kentquirk:
The config reloads happen whenever the file changes, although this is mediated by file change notifications built into the configuration management library, and when you get k8s involved, this might get complicated.

However, dumping configuration to an unauthenticated request seems like too big a security problem to allow, and attempting an authentication model seems fraught with slippery slope warnings. 

All reload events are logged with Debug level, so you can see them happen in the log if you're not sure.

If it helps, I'd be willing to consider adding a timestamp for when the configuration was last loaded, and making that available through a separate endpoint.

mterhar:
Moving the reload timestamp on an info-level log event (only when the reload detects a difference) might be enough for most operators. Including a hash of the updated file or some other metadata like last write time would allow people to confirm what they think happened.

PR #547, opened 2022-10-28
New cache management strategy

kentquirk:
Implements a new cache management strategy that ejects "large" items from the cache rather than resizing the cache.

Currently, when Refinery is under memory pressure and exceeds the configured memory maximum, it attempts to resize the trace cache to 90% of its previous size, and ejects the oldest traces. But as the trace cache is sized by trace count, when a very large trace arrives, it can cause the cache to shrink repeatedly, discarding the smaller traces, which doesn't help much. The result is that the cache can be resized to a tiny fraction of its original size to very little benefit.

Furthermore, it never recovers until configuration is manually reloaded. 

This PR implements a different strategy:
* The memory size of individual spans is calculated when they are placed into the cache
* Spans also track their arrival time
* The "cacheImpact" of a span is a measure of how long the span has been in the cache, multiplied by the size of the span
* Traces keep track of the total impact of all the spans in the trace

When a memory overrun occurs, the system sorts the traces by cache impact and ejects (makes a sampling decision and drops or sends) those traces with the largest impact until memory usage falls below the maximum. The cache is not resized.

This strategy leads to more stable memory usage and fewer overruns.

* The memory size of individual spans is calculated when they are placed into the cache
* Spans also track their arrival time
* The "cacheImpact" of a span is a measure of how long the span has been in the cache, multiplied by the size of the span
* Traces keep track of the total impact of all the spans in the trace
* There's some new telemetry
* There's a config value to control switching between the two modes dynamically and it shows up in the sample config
* There are tests for some of the algorithmic calculations as well as the strategy modes

Note to reviewers -- this is on the large side, but I didn't see a great way to break it up.

### DIFF
--- a/collect/cache/cache.go
+++ b/collect/cache/cache.go
@@ -145,9 +145,9 @@ func (d *DefaultInMemCache) TakeExpiredTraces(now time.Time) []*types.Trace {
 	return res
 }
 
-// RemoveSentTraces accepts a set of trace IDs and removes any matching ones from
+// RemoveTraces accepts a set of trace IDs and removes any matching ones from
 // the insertion list. This is used in the case of a cache overrun.
-func (d *DefaultInMemCache) RemoveSentTraces(toDelete map[string]struct{}) {
+func (d *DefaultInMemCache) RemoveTraces(toDelete map[string]struct{}) {
 	d.Metrics.Gauge("collect_cache_capacity", float64(len(d.insertionOrder)))
 	d.Metrics.Histogram("collect_cache_entries", float64(len(d.cache)))
 
--- a/collect/cache/cache.go
+++ b/collect/cache/cache.go
@@ -128,6 +128,8 @@ func (d *DefaultInMemCache) GetAll() []*types.Trace {
 	return tmp
 }
 
+// TakeExpiredTraces should be called to decide which traces are past their expiration time;
+// It removes and returns them.
 func (d *DefaultInMemCache) TakeExpiredTraces(now time.Time) []*types.Trace {
 	d.Metrics.Gauge("collect_cache_capacity", float64(len(d.insertionOrder)))
 	d.Metrics.Histogram("collect_cache_entries", float64(len(d.cache)))
@@ -141,4 +143,20 @@ func (d *DefaultInMemCache) TakeExpiredTraces(now time.Time) []*types.Trace {
 		}
 	}
 	return res
+}
+
+// RemoveSentTraces accepts a set of trace IDs and removes any matching ones from
+// the insertion list. This is used in the case of a cache overrun.
+func (d *DefaultInMemCache) RemoveSentTraces(toDelete map[string]struct{}) {
+	d.Metrics.Gauge("collect_cache_capacity", float64(len(d.insertionOrder)))
+	d.Metrics.Histogram("collect_cache_entries", float64(len(d.cache)))
+
+	for i, t := range d.insertionOrder {
+		if t != nil {
+			if _, ok := toDelete[t.TraceID]; ok {
+				d.insertionOrder[i] = nil
+				delete(d.cache, t.TraceID)
+			}
+		}
+	}
 }
--- a/collect/collect.go
+++ b/collect/collect.go
@@ -47,6 +47,14 @@ func GetCollectorImplementation(c config.Config) Collector {
 	return collector
 }
 
+// These are the name of the metric to increment
+const (
+	TraceSendGotRoot        = "trace_send_got_root"
+	TraceSendExpired        = "trace_send_expired"
+	TraceSendEjectedFull    = "trace_send_ejected_full"
+	TraceSendEjectedMemsize = "trace_send_ejected_memsize"
+)
+
 // InMemCollector is a single threaded collector
 type InMemCollector struct {
 	Config         config.Config          `inject:""`
@@ -100,12 +108,18 @@ func (i *InMemCollector) Start() error {
 	i.Metrics.Register("collector_tosend_queue", "histogram")
 	i.Metrics.Register("collector_incoming_queue", "histogram")
 	i.Metrics.Register("collector_peer_queue", "histogram")
+	i.Metrics.Register("collector_cache_size", "gauge")
+	i.Metrics.Register("memory_heap_allocation", "gauge")
 	i.Metrics.Register("trace_sent_cache_hit", "counter")
 	i.Metrics.Register("trace_accepted", "counter")
 	i.Metrics.Register("trace_send_kept", "counter")
 	i.Metrics.Register("trace_send_dropped", "counter")
 	i.Metrics.Register("trace_send_has_root", "counter")
 	i.Metrics.Register("trace_send_no_root", "counter")
+	i.Metrics.Register(TraceSendGotRoot, "counter")
+	i.Metrics.Register(TraceSendExpired, "counter")
+	i.Metrics.Register(TraceSendEjectedFull, "counter")
+	i.Metrics.Register(TraceSendEjectedMemsize, "counter")
 
 	stc, err := lru.New(imcConfig.CacheCapacity * 5) // keep 5x ring buffer size
 	if err != nil {
@@ -155,7 +169,7 @@ func (i *InMemCollector) reloadConfigs() {
 			// pull the old cache contents into the new cache
 			for j, trace := range existingCache.GetAll() {
 				if j >= imcConfig.CacheCapacity {
-					i.send(trace)
+					i.send(trace, TraceSendEjectedFull)
 					continue
 				}
 				c.Set(trace)
@@ -174,17 +188,21 @@ func (i *InMemCollector) reloadConfigs() {
 	// TODO add resizing the LRU sent trace cache on config reload
 }
 
-func (i *InMemCollector) checkAlloc() {
+func (i *InMemCollector) oldCheckAlloc() {
 	inMemConfig, err := i.Config.GetInMemCollectorCacheCapacity()
 
 	var mem runtime.MemStats
 	runtime.ReadMemStats(&mem)
+	i.Metrics.Gauge("memory_heap_allocation", int64(mem.Alloc))
 	if err != nil || inMemConfig.MaxAlloc == 0 || mem.Alloc < inMemConfig.MaxAlloc {
 		return
 	}
 
 	existingCache, ok := i.cache.(*cache.DefaultInMemCache)
-	if !ok || existingCache.GetCacheSize() < 100 {
+	existingSize := existingCache.GetCacheSize()
+	i.Metrics.Gauge("collector_cache_size", existingSize)
+
+	if !ok || existingSize < 100 {
 		i.Logger.Error().WithField("alloc", mem.Alloc).Logf(
 			"total allocation exceeds limit, but unable to shrink cache",
 		)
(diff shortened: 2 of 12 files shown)


collect/collect.go
```diff
@@ -47,6 +47,14 @@ func GetCollectorImplementation(c config.Config) Collector {
 	return collector
 }
 
+// These are the name of the metric to increment
```
MikeGoldsmith, line 50:
```suggestion
// These are the names of the metrics to increment
```

rules_complete.toml
```diff
@@ -242,6 +242,22 @@ SampleRate = 1
 			AddSampleRateKeyToTrace = true
 			AddSampleRateKeyToTraceField = "meta.refinery.dynsampler_key"
 
+	# Note that Refinery comparisons are type-dependent. If you are operating in an environment where different
```
MikeGoldsmith, line 245: 🎉

types/event.go
```diff
@@ -92,7 +114,46 @@ func (t *Trace) GetSamplerKey() (string, bool) {
...
+		default:
+			total += 8 // catchall
+		}
+	}
+	return total
+}
+
+// cacheImpactFactor controls how much we overweight older spans compared to newer ones;
+// setting this to 1 means they're not weighted by duration
+const cacheImpactFactor = 4
```
MikeGoldsmith, line 146: Should this be a config option?

config_complete.toml
```diff
@@ -143,6 +143,16 @@ AdditionalErrorFields = [
...
 
+# UseStableCacheManagement controls the use of a non-resizing cache management strategy.
+# It defaults to false, but is recommended to set to true, as the old strategy does not recover from
+# cache overruns.
+# In the previous strategy, under memory pressure the cache is shrunk and never grows again.
+# In the stable strategy, under memory pressure the items having the most impact on the cache size are
+# ejected from the cache early and the cache is not resized.
+# This only applies if MaxAlloc is nonzero.
+# Eligible for live reload.
+# UseStableCacheManagement = true
```
MikeGoldsmith, line 154: I feel like having a boolean here isn't too future proof - if we have to explore other strategies in the future, it means deprecating this option. An alternative config option would be something like "CacheEvictionStrategy" with the default being the existing "Resizing" and this one being "CacheImpact" (names are just suggestions).

MikeGoldsmith (commented):
I think this look really good.

I've left some suggestions, with the main one being how we expose the cache strategy via configuration.

config_complete.toml
kentquirk: That's a really good point, and I will implement it. Thank you.

collect/cache/cache.go
```diff
@@ -142,3 +144,19 @@ func (d *DefaultInMemCache) TakeExpiredTraces(now time.Time) []*types.Trace {
 	}
 	return res
 }
+
+// RemoveSentTraces accepts a set of trace IDs and removes any matching ones from
+// the insertion list. This is used in the case of a cache overrun.
+func (d *DefaultInMemCache) RemoveSentTraces(toDelete map[string]struct{}) {
```
TylerHelmuth, line 150: The name of this method implies it checks if the trace was sent or not, but it doesn't look like thats the case.  Should we go with `RemoveTraces` since the only criteria is that the trace is present in the `toDelete` map?

collect/collect.go
```diff
@@ -228,6 +245,76 @@ func (i *InMemCollector) checkAlloc() {
...
+	i.Metrics.Gauge("memory_heap_allocation", int64(mem.Alloc))
+	if err != nil || inMemConfig.MaxAlloc == 0 || mem.Alloc < inMemConfig.MaxAlloc {
+		return
+	}
+
+	// Figure out what fraction of the total cache we should remove. We'd like it to be
+	// enough to get us below the max capacity, but not TOO much below.
+	// Because our impact numbers are only the data size, we shoot for 90% of max
+	// capacity and we should end up comfortably below that.
+	totalToRemove := mem.Alloc - inMemConfig.MaxAlloc //*9/10
```
TylerHelmuth, line 262: What is `//*9/10` implying?

collect/collect.go
```diff
@@ -228,6 +245,76 @@ func (i *InMemCollector) checkAlloc() {
...
+	// until the total size is less than 90% of the intended max.
+
+	existingCache, ok := i.cache.(*cache.DefaultInMemCache)
+	if !ok {
+		i.Logger.Error().WithField("alloc", mem.Alloc).Logf(
+			"total allocation exceeds limit, but unable to control cache",
+		)
+		return
+	}
+	allTraces := existingCache.GetAll()
```
TylerHelmuth, line 276: If this function is getting used in an OOM risk scenario, will retrieving the entire cache put extra strain on refinery?

collect/collect.go
```diff
@@ -228,6 +245,76 @@ func (i *InMemCollector) checkAlloc() {
...
+	timeout, err := i.Config.GetTraceTimeout()
+	if err != nil {
+		timeout = 60 * time.Second
```
TylerHelmuth, line 279: Is this refinery's default timeout?

TylerHelmuth (commented):
Great stuff.  Left a couple questions here and there

collect/cache/cache.go
kentquirk: Good catch; previously, it wasn't a bad name, but this is better.

collect/collect.go
kentquirk: It is. This is only used in an error condition, which only happens if you specify an unparseable value for a timeout.

collect/collect.go
kentquirk: Heh, that I had a 90% in there but decided it was too dependent on cache details so removed it. I've cleaned up the comment.

config/file_config.go
```diff
@@ -295,6 +295,15 @@ func (f *fileConfig) validateConditionalConfigs() error {
...
 		}
 	}
+
+	// validate cache strategy
+	st := f.GetCacheOverrunStrategy()
+	switch st {
+	case "resize", "impact":
+		break
+	default:
+		return fmt.Errorf("invalid CacheOverrunStrategy: '%s'", st)
```
MikeGoldsmith, line 305: An empty value `""` would cause this to error.

What happens when an error is returned here? Does it stop refinery from starting?

MikeGoldsmith (commented):
Updated config looks good - I've left a query regarding config validation and empty values.

TylerHelmuth (approved):
Looks good to me as long as @MikeGoldsmith question is addressed

config/file_config.go
kentquirk: Because a default of "resize" is baked into the code, an empty value is only possible if someone explicitly overrides the configuration and forces it to empty, or some other value we don't recognize. 

In these cases, if it occurs on initial config load, refinery logs it and fails to start -- if it occurs on an edit to a running config, it logs it and doesn't change the loaded config.

collect/collect.go
kentquirk: In a conversation I demonstrated that it only duplicates the trace records, which are relatively small, and this is exactly what the "resize" code does as well, so it's no worse.

MikeGoldsmith (approved):

collect/collect.go
```diff
@@ -228,6 +245,77 @@ func (i *InMemCollector) checkAlloc() {
...
+		return
+	}
+
+	// Figure out what fraction of the total cache we should remove. We'd like it to be
+	// enough to get us below the max capacity, but not TOO much below.
+	// Because our impact numbers are only the data size, reducing by enough to reach
+	// max alloc will actually do more than that.
+	totalToRemove := mem.Alloc - inMemConfig.MaxAlloc
+
+	// the size of the cache exceeds the user's intended allocation, so we're going to
```
seh, line 264:
```suggestion
	// The size of the cache exceeds the user's intended allocation, so we're going to
```

collect/collect.go
```diff
@@ -228,6 +245,77 @@ func (i *InMemCollector) checkAlloc() {
...
+	// remove the traces from the cache that have had the most impact on allocation.
+	// To do this, we sort the traces by their CacheImpact value and then remove traces
+	// until the total size is less than 90% of the intended max.
```
seh, line 267:
```suggestion
	// until the size of the cache shrinks below 90% of the intended maximum size.
```

types/event.go
```diff
@@ -92,7 +114,46 @@ func (t *Trace) GetSamplerKey() (string, bool) {
...
+			total += len(v.(string))
+		default:
+			total += 8 // catchall
+		}
+	}
+	return total
+}
+
+// cacheImpactFactor controls how much we overweight older spans compared to newer ones;
+// setting this to 1 means they're not weighted by duration
```
seh, line 145:
```suggestion
// cacheImpactFactor controls how much more we weigh older spans compared to newer ones;
// setting this to 1 means they're not weighted by duration.
```

types/event.go
```diff
@@ -92,7 +114,46 @@ func (t *Trace) GetSamplerKey() (string, bool) {
...
+const cacheImpactFactor = 4
+
+// CacheImpact calculates an abstract value for something we're calling cache impact, which is
+// the product of the size of the span and a factor related to the amount of time the span
+// has been stored in the cache, based on the TraceTimeout value.
+func (sp *Span) CacheImpact(traceTimeout time.Duration) int {
+	// multiplier will be a value from 1-cacheImpactFactor, depending on how long the
+	// span has been in the cache compared to the traceTimeout.
+	multiplier := int(cacheImpactFactor*time.Since(sp.ArrivalTime)/traceTimeout) + 1
+	// we can assume DataSize was set when the span was added
```
seh, line 155:
```suggestion
	// We can assume DataSize was set when the span was added.
```

types/event.go
```diff
@@ -92,7 +114,46 @@ func (t *Trace) GetSamplerKey() (string, bool) {
...
+// setting this to 1 means they're not weighted by duration
+const cacheImpactFactor = 4
+
+// CacheImpact calculates an abstract value for something we're calling cache impact, which is
+// the product of the size of the span and a factor related to the amount of time the span
+// has been stored in the cache, based on the TraceTimeout value.
+func (sp *Span) CacheImpact(traceTimeout time.Duration) int {
+	// multiplier will be a value from 1-cacheImpactFactor, depending on how long the
+	// span has been in the cache compared to the traceTimeout.
+	multiplier := int(cacheImpactFactor*time.Since(sp.ArrivalTime)/traceTimeout) + 1
```
seh, line 154: Are we assuming that the time since arrival can never exceed the trace timeout?

collect/collect.go
```diff
@@ -348,18 +447,18 @@ func (i *InMemCollector) processSpan(sp *types.Span) {
...
-			TraceID:    sp.TraceID,
-			StartTime:  time.Now(),
-			SendBy:     time.Now().Add(timeout),
-			SampleRate: sp.SampleRate, // if it had a sample rate, we want to keep it
+			APIHost:     sp.APIHost,
+			APIKey:      sp.APIKey,
+			Dataset:     sp.Dataset,
+			TraceID:     sp.TraceID,
+			ArrivalTime: time.Now(),
+			SendBy:      time.Now().Add(timeout),
```
seh, line 455: For clarity in the resulting values, consider capturing `time.Now` once above line 449:
```go
now := time.Now()
```
Then, use `now` whenever we're trying to refer to the current time.

collect/collect.go
```diff
@@ -228,6 +245,77 @@ func (i *InMemCollector) checkAlloc() {
...
+		)
+		return
+	}
+	allTraces := existingCache.GetAll()
+	timeout, err := i.Config.GetTraceTimeout()
+	if err != nil {
+		timeout = 60 * time.Second
+	} // Sort traces by CacheImpact, heaviest first
+	sort.Slice(allTraces, func(i, j int) bool {
+		return allTraces[i].CacheImpact(timeout) > allTraces[j].CacheImpact(timeout)
```
seh, line 282: Since `CacheImpact` may be called multiple times for some traces, it _may_ be worth memoizing. It would take a benchmark to be more sure.

types/event.go
kentquirk: Not necessarily, and this won't break if it does, but I've documented a bit better. Thanks for the question.

collect/collect.go
kentquirk: Good idea.

collect/collect.go
kentquirk: Good catch! I did a little abstract benchmark of Sort() and it calls the comparator `O(Nlog(N))` times! So even without testing here, it seemed well worth memoizing this.

PR #556, opened 2022-11-09
Add command to query config metadata

kentquirk:

- Operators would like to know that their refinery is running current configurations, and to be able to verify that their configuration file in use is the one they believe it to be.  For security reasons, we can't simply dump the configuration information, so instead we can calculate a hash and indicate the time when the configuration file was last loaded. This should be especially helpful in a cluster to make sure that different instances have the same configuration in use.

- Closes #542 

## Changes made

- Add a `/query/configmetadata` endpoint that has the same query header rule as the other query endpoints
- Add a function to calculate the md5 of a file (verified to return the same results as the `md5sum` command)
- Use that function to generate the sums of the config files currently in use
- Record the time whenever config is loaded
- Report all of this in JSON format when requested
- Update README.md

Will require docs update.

Sample of the generated data:

```json
[
  {
    "type": "config",
    "id": "tools/loadtest/config.toml",
    "hash": "1047bb6140b487ecdb0745f3335b6bc3",
    "loaded_at": "2022-11-08T22:24:18-05:00"
  },
  {
    "type": "rules",
    "id": "tools/loadtest/rules.toml",
    "hash": "2d88389e1ff6530fba53466973e591e0",
    "loaded_at": "2022-11-08T22:24:18-05:00"
  }
]
```


config/config.go
```diff
@@ -170,4 +170,13 @@ type Config interface {
 	GetAddSpanCountToRoot() bool
 
 	GetCacheOverrunStrategy() string
+
+	GetConfigMetadata() []ConfigMetadata
+}
+
+type ConfigMetadata struct {
+	Type     string `json:"type"`
+	ID       string `json:"id"`
```
MikeGoldsmith, line 179: I\m not sure ID is clear that it's going to return the path to the file, should we use `path` instead?

MikeGoldsmith (approved):
Looks good, good work @kentquirk. I've left one suggestion for updating one of the fields in the new struct.

config/config.go
kentquirk: I chose not to name it path or filename because I have longterm plans to allow configuration in other ways besides files that have a discernable path (for example, a shared configuration in Redis).

PR #558, opened 2022-11-10
change to environment to correctly represent

sgsharma:
The naming in the rules file uses terminology (`dataset`) that no longer reflects the Honeycomb experience. Wondering if it makes sense to change that to reflect the new world of E&S

-

Updated dataset rule blocks to environment nomenclature 

-

### DIFF
--- a/rules_complete.toml
+++ b/rules_complete.toml
@@ -22,7 +22,7 @@ Sampler = "DeterministicSampler"
 # Eligible for live reload.
 SampleRate = 1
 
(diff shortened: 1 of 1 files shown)


kentquirk:
@sgsharma, thank you for this. Unfortunately, we are not *quite* ready to make this change, and when we do it will be accompanied by a more sweeping documentation change. I'm going to close it for now, but I've written #560 to keep track of the task.

PR #561, opened 2022-11-13
Extract Sent Cache to an interface for future expansion

kentquirk:

This is a prep PR for further work on the sent trace cache. For improved scalability, we want to be able to store trace decision records for a longer time. The best way to implement this in a backwards-compatible way is to pull the mechanisms for managing decision records into an interface, and then implement the interface with the legacy logic. That's what this does. There are no expected changes in behavior, and all tests still pass.

- Define interfaces for a TraceSentCache and a TraceSentRecord
- Implement code for those that duplicates the existing legacy logic
- Refactor the places the code is used to use the new interfaces
- Tweak span count data type so that it's not an int64 any more
- Rename SpanCount to DescendantCount because now that we have links and events, they're not all Spans anymore, and future PRs will add additional counting functions.  I haven't renamed the corresponding Get and Add functions because that's pretty messy.

### DIFF
--- a/collect/cache/traceSentCache.go
+++ b/collect/cache/traceSentCache.go
@@ -9,8 +9,8 @@ type TraceSentRecord interface {
 	Kept() bool
 	// Rate() returns the sample rate for the trace
 	Rate() uint
-	// Descendants returns the count of items associated with the trace, including all types of children like span links and span events.
-	Descendants() uint
+	// DescendantCount returns the count of items associated with the trace, including all types of children like span links and span events.
+	DescendantCount() uint
 	Add(*types.Span)
 }
 
--- a/collect/cache/traceSentCache.go
+++ b/collect/cache/traceSentCache.go
@@ -1,0 +1,23 @@
+package cache
+
+import (
+	"github.com/honeycombio/refinery/types"
+)
+
+type TraceSentRecord interface {
+	// Kept returns whether the trace was kept (sampled and sent to honeycomb) or dropped.
+	Kept() bool
+	// Rate() returns the sample rate for the trace
+	Rate() uint
+	// Descendants returns the count of items associated with the trace, including all types of children like span links and span events.
+	Descendants() uint
+	Add(*types.Span)
+}
+
+type TraceSentCache interface {
+	// Record preserves the record of a trace being sent or not.
+	Record(trace *types.Trace, keep bool)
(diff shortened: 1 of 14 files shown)


collect/cache/traceSentCache.go
```diff
@@ -0,0 +1,24 @@
+package cache
+
+import (
+	"github.com/honeycombio/refinery/types"
+)
+
+type TraceSentRecord interface {
+	// Kept returns whether the trace was kept (sampled and sent to honeycomb) or dropped.
+	Kept() bool
```
MikeGoldsmith, line 9: As we're introducing an interface, can we consolidate the wording to match the usage / sampler? eg we use kept for the record, and should send from the sampler. Would `Sent` be clearer?

collect/cache/traceSentCache.go
```diff
@@ -0,0 +1,24 @@
...
+	// Count records additional spans in the totals
+	Count(*types.Span)
+}
+
+type TraceSentCache interface {
+	// Record preserves the record of a trace being sent or not.
+	Record(trace *types.Trace, keep bool)
+	// Check tests if a trace corresponding to the span is in the cache; if found, it returns the appropriate TraceSentRecord and true,
+	// else nil and false.
+	Check(span *types.Span) (TraceSentRecord, bool)
```
MikeGoldsmith, line 23: How do you feel about using a TraceID instead of a span. Something like:

```suggestion
	Get(traceID string) (TraceSentRecord, bool)
```

MikeGoldsmith (approved):
Looks great @kentquirk 👍🏻 

I've left a couple of suggestions for updating the new interface but none are essential as they are for internal use only.

collect/cache/traceSentCache.go
kentquirk: I did consider that, but there's code in refinery that uses "send" to mean, effectively, "do something with this" and the result is either keep or drop. So I think it's clearer to see this from the cache's point of view.

collect/cache/traceSentCache.go
kentquirk: I did that originally, but this is partly prep for a future change where we keep more information in the kept cache regarding not just total number of descendants, but also keep separate counts of other things like span events, links, etc. To do that I need to inspect the whole span.

PR #567, opened 2022-11-18
Add cuckoo-based drop cache

kentquirk:

Currently, Refinery uses a circular LRU cache to retain a record for every trace; this cache is hardcoded to 5x the configured cache size, and does not change when the cache is resized. 

This is a relatively small number, and for long-lived traces, it might mean that late spans look like new traces, and therefore might get a different sampling decision -- which would result in missing spans in Honeycomb.

#561 abstracted the SampleCache interface to prepare for other implementations. This uses it to provide a new cache design.

This design implements a "cuckoo filter" cache for dropped traces, which can store the dropped trace information much more efficiently (about 4 bytes per trace as compared to about 200 bytes for kept traces). 

- Adds a CuckooTraceChecker type that implements a 2-stage cuckoo filter for tracking recently-used trace IDs over time.
- Implements the SampleCache interface with a CuckooSentCache, which uses the existing LRU for kept traces, and a CuckooTraceChecker for dropped traces.
- Implements a new configuration block for caches to allow users to opt into the cuckoo cache and control it for their needs, but is still backwards compatible.
- Adds documentation to the config_complete file.

### DIFF
--- a/collect/cache/cuckoo.go
+++ b/collect/cache/cuckoo.go
@@ -51,18 +51,17 @@ func (c *CuckooTraceChecker) Maintain() {
 	c.mut.RLock()
 	dropFactor := c.current.LoadFactor()
 	c.mut.RUnlock()
+
 	// once the current one is half full, we can drop the firstTime check
-	if c.firstTime && dropFactor > 0.5 {
+	if (c.firstTime && dropFactor > 0.5) || dropFactor > 0.99 {
 		c.mut.Lock()
 		defer c.mut.Unlock()
 		c.firstTime = false
-	}
-	// if the current one is full, cycle the filters
-	if dropFactor > 0.99 {
-		c.mut.Lock()
-		defer c.mut.Unlock()
-		c.current = c.future
-		c.future = cuckoo.NewFilter(c.capacity)
+		// if the current one is full, cycle the filters
+		if dropFactor > 0.99 {
+			c.current = c.future
+			c.future = cuckoo.NewFilter(c.capacity)
+		}
 	}
 }
 
--- a/collect/cache/cuckoo.go
+++ b/collect/cache/cuckoo.go
@@ -3,29 +3,42 @@ package cache
 import (
 	"sync"
 
+	"github.com/honeycombio/refinery/metrics"
 	cuckoo "github.com/panmari/cuckoofilter"
+)
+
+// These are the names of metrics tracked for the cuckoo filter
+const (
+	CurrentLoadFactor = "cuckoo_current_load_factor"
+	FutureLoadFactor  = "cuckoo_future_load_factor"
+	CurrentCapacity   = "cuckoo_current_capacity"
 )
 
 // This wraps a cuckoo filter implementation in a way that lets us keep it running forever
 // without filling up.
-// You must call Maintain() periodically, most likely from a goroutine.
-// We maintain two filters that are out of sync; when current is full, future is half full
-// and we then discard current and replace it with the future, and then create a new, empty future.
-// This means that we need to use the firstTime flag to avoid putting anything into future
-// until after current reaches .5.
+// A cuckoo filter can't be emptied (you can delete individual items if you know what they are,
+// but you can't get their names from the filter. Consequently, what we do is keep *two* filters,
+// current and future. The current one is the one we use to check against, and when we add, we
+// add to both. But the future one is started *after* the current one, so that when the current
+// gets too full, we can discard it, replace it with future, and then start a new, empty future.
+// This is why the future filter is nil until the current filter reaches .5.
+// You must call Maintain() periodically, most likely from a goroutine. The call is cheap,
+// and the timing isn't very critical. The effect of going above "capacity" is an increased
+// false positive rate, but the filter continues to function.
 type CuckooTraceChecker struct {
-	current   *cuckoo.Filter
-	future    *cuckoo.Filter
-	mut       sync.RWMutex
-	capacity  uint
-	firstTime bool
+	current  *cuckoo.Filter
+	future   *cuckoo.Filter
+	mut      sync.RWMutex
+	capacity uint
+	met      metrics.Metrics
 }
 
-func NewCuckooTraceChecker(capacity uint) *CuckooTraceChecker {
+func NewCuckooTraceChecker(capacity uint, m metrics.Metrics) *CuckooTraceChecker {
 	return &CuckooTraceChecker{
 		capacity: capacity,
 		current:  cuckoo.NewFilter(capacity),
-		future:   cuckoo.NewFilter(capacity),
+		future:   nil,
+		met:      m,
 	}
 }
 
@@ -34,8 +47,8 @@ func (c *CuckooTraceChecker) Add(traceID string) {
 	c.mut.Lock()
 	defer c.mut.Unlock()
 	c.current.Insert([]byte(traceID))
-	// don't add anything to future until we're no longer in the 'firstTime' section.
-	if !c.firstTime {
+	// don't add anything to future if it doesn't exist yet
+	if c.future != nil {
 		c.future.Insert([]byte(traceID))
 	}
 }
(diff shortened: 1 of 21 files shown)


collect/cache/cuckooSentCache.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+	// if we're not keeping it, save it in the dropped trace filter
+	c.dropped.Add(trace.TraceID)
+}
+
+func (c *cuckooSentCache) Check(span *types.Span) (TraceSentRecord, bool) {
+	// was it dropped?
+	if c.dropped.Check(span.TraceID) {
+		// we recognize it as dropped, so just say so; there's nothing else to do
+		return &cuckooDroppedRecord{}, false
```
MikeGoldsmith, line 136: Would it be worth creating a local const for the dropped record we always return instead of creating a new one each time?

collect/cache/cuckooSentCache.go
```diff
@@ -0,0 +1,179 @@
...
+	stc, err := lru.New(int(cfg.KeptSize))
+	if err != nil {
+		return err
+	}
+
+	// grab all the items in the current cache; if it's larger than
+	// what will fit in the new one, discard the oldest ones
+	// (we don't have to do anything with the ones we discard, this is
+	// the trace decisions cache).
+	keys := c.kept.Keys()
```
MikeGoldsmith, line 160: Do we need a lock here to protect the current set of keys and prevent adding new ones during the resize?

collect/collect.go
```diff
@@ -111,9 +111,20 @@ func (i *InMemCollector) Start() error {
...
+		if err != nil {
+			return err
+		}
+	case "cuckoo":
+		i.sampleTraceCache, err = cache.NewCuckooSentCache(sampleCacheConfig)
+		if err != nil {
+			return err
+		}
+	default:
+		return fmt.Errorf("sampleTraceCache had impossible config type %s", sampleCacheConfig.Type)
```
MikeGoldsmith, line 127:
```suggestion
		return fmt.Errorf("sampleTraceCache type '%s' not recognized", sampleCacheConfig.Type)
```

collect/cache/cuckooSentCache.go
```diff
@@ -0,0 +1,179 @@
...
+
+// Make sure it implements TraceSentRecord
+var _ TraceSentRecord = (*cuckooKeptRecord)(nil)
+
+// cuckooSentRecord is what a TraceSentRecord we return when the trace was dropped.
+// It's always the same one.
+type cuckooDroppedRecord struct{}
+
+func (t *cuckooDroppedRecord) Kept() bool {
+	return true
```
MikeGoldsmith, line 49: Dropped record should return false, right?

```suggestion
	return false
```

MikeGoldsmith (commented):
Also needs PR title updating 😄

collect/cache/cuckoo.go
```diff
@@ -0,0 +1,74 @@
...
+		c.future.Insert([]byte(traceID))
+	}
+}
+
+// Check tests if a traceID is (very probably) in the filter.
+func (c *CuckooTraceChecker) Check(traceID string) bool {
+	b := []byte(traceID)
+	c.mut.RLock()
+	defer c.mut.RUnlock()
+	return c.current.Lookup(b)
```
TylerHelmuth, line 45: I'm no expert in caching strategy, but when reading about cuckoo cache for this PR it sounds like cuckoo can give false positives, but never false negatives.  Do we care about false positives here?  What will be the impact of a false positive?

collect/cache/cuckoo.go
```diff
@@ -0,0 +1,74 @@
...
+	// once the current one is half full, we can drop the firstTime check
+	if c.firstTime && dropFactor > 0.5 {
+		c.mut.Lock()
+		defer c.mut.Unlock()
+		c.firstTime = false
+	}
+	// if the current one is full, cycle the filters
+	if dropFactor > 0.99 {
+		c.mut.Lock()
+		defer c.mut.Unlock()
```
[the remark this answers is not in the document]
kentquirk: Fixed in latest update, please re-review.

collect/cache/cuckooSentCache.go
kentquirk: Fixed in latest

collect/collect.go
kentquirk: I adjusted it some.

collect/cache/cuckoo.go
kentquirk: This is true -- a positive says "we saw this trace before and dropped it". The effect of this means that some trace that might not have been dropped might instead be dropped. But since the decision will be consistent, it's no worse than making a negative random choice a few nanoseconds later.

collect/cache/cuckoo.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+// Maintain should be called periodically; if the current filter is full, it replaces
+// it with the future filter and creates a new future filter.
+func (c *CuckooTraceChecker) Maintain() {
+	c.mut.RLock()
+	dropFactor := c.current.LoadFactor()
+	c.mut.RUnlock()
+	// once the current one is half full, we can drop the firstTime check
+	if c.firstTime && dropFactor > 0.5 {
```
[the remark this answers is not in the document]
kentquirk: Basically, we keep two filters. In the long term, we want `current` to be more than .5 full and `future` to be less than .5 full. But to get there, we need to start filling `current` without also filling `future`.  Hence the firstTime flag.

collect/cache/cuckoo.go
```diff
@@ -0,0 +1,76 @@
...
+// We maintain two filters that are out of sync; when current is full, future is half full
+// and we then discard current and replace it with the future, and then create a new, empty future.
+// This means that we need to use the firstTime flag to avoid putting anything into future
+// until after current reaches .5.
+type CuckooTraceChecker struct {
+	current   *cuckoo.Filter
+	future    *cuckoo.Filter
+	mut       sync.RWMutex
+	capacity  uint
+	firstTime bool
```
TylerHelmuth, line 21: nit: I'd prefer a name like like `halfEmpty` or `insertIntoBoth` (that one may change some of the if statements).  I'm still struggling with `firstTime` because it feels like it is tracking something that only happens 1 time, but it reality `c.current.Insert([]byte(traceID))` could happen many times before `firstTime` is set to false.

collect/cache/cuckoo.go
```diff
@@ -0,0 +1,76 @@
...
+	dropFactor := c.current.LoadFactor()
+	c.mut.RUnlock()
+
+	// once the current one is half full, we can drop the firstTime check
+	if (c.firstTime && dropFactor > 0.5) || dropFactor > 0.99 {
+		c.mut.Lock()
+		defer c.mut.Unlock()
+		c.firstTime = false
+		// if the current one is full, cycle the filters
+		if dropFactor > 0.99 {
```
TylerHelmuth, line 64: when we cycle the filters do we need to set `firstTime` back to true?

collect/cache/cuckooSentCache.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+	dropped := NewCuckooTraceChecker(cfg.DroppedSize)
+
+	cache := &cuckooSentCache{
+		kept:    stc,
+		dropped: dropped,
+		cfg:     cfg,
+		done:    make(chan struct{}),
+	}
+	// TODO: metrics for when this puppy gets cycled
```
TylerHelmuth, line 99: Do we have an issue to track this TODO?

collect/cache/cuckoo.go
kentquirk: We don't -- the point is that we need that flag (I'll rename it) to get them out of sync, and then the cycling ensures that one is always more than half full while the next one is filling up. The point of this is that after the initial fillup, there the cache is always at least half full, which means we maximize the amount of items we are storing.

I'm going to rework this a little bit to make it more obvious.

TylerHelmuth: Ah ok, so it's only going into the first cache when it first starts up, but one that cache gets halfway full for the first time that kicks off the "dual fill up" approach, which will never stop.

In that case, I am missing where `firstTime` is ever set to true.  I think it'll be false when we call `NewCuckooTraceChecker`, so the desyncing will only occur when `dropFactor > 0.99` right?  Or have I missed where it gets set to true initially?

collect/cache/cuckoo.go
kentquirk: I've reworked this to eliminate this entirely and just leave future as nil until it's time to use it. Also has a lot more comments.

collect/cache/cuckooSentCache.go
kentquirk: I've added the metrics instead.

collect/cache/cuckoo.go
```diff
@@ -0,0 +1,97 @@
...
+const (
+	CurrentLoadFactor = "cuckoo_current_load_factor"
+	FutureLoadFactor  = "cuckoo_future_load_factor"
+	CurrentCapacity   = "cuckoo_current_capacity"
+)
+
+// This wraps a cuckoo filter implementation in a way that lets us keep it running forever
+// without filling up.
+// A cuckoo filter can't be emptied (you can delete individual items if you know what they are,
+// but you can't get their names from the filter. Consequently, what we do is keep *two* filters,
```
TylerHelmuth, line 20:
```suggestion
// but you can't get their names from the filter). Consequently, what we do is keep *two* filters,
```

collect/cache/cuckoo.go
```diff
@@ -0,0 +1,97 @@
...
+// current and future. The current one is the one we use to check against, and when we add, we
+// add to both. But the future one is started *after* the current one, so that when the current
+// gets too full, we can discard it, replace it with future, and then start a new, empty future.
+// This is why the future filter is nil until the current filter reaches .5.
+// You must call Maintain() periodically, most likely from a goroutine. The call is cheap,
+// and the timing isn't very critical. The effect of going above "capacity" is an increased
+// false positive rate, but the filter continues to function.
```
TylerHelmuth, line 27: Great thing to callout 👍

TylerHelmuth (approved):
LGTM!

collect/cache/cuckoo.go
kentquirk: Thanks. Just so you don't think I'm being a jerk, I don't generally press the "commit suggestion" button because then I don't have the commit on my local branch except after it's been squashed, which means that I can't autodelete the branches.

PR #569, opened 2022-11-23
On shutdown, remove ourself from the peers list

### LINKED ISSUE #393: Refinery nodes should remove themselves from peer management on shutdown
When using a peer management coordinator (eg Redis) to determine which nodes nodes are active and how to distribute traffic across the cluster, when a node goes down the coordinator has to wait 30 seconds for it to be removed from the cache of active nodes.

A refinery node on shutdown should remove itself from the cache so that the cache is updated quicker.

**Additional context**

kentquirk:

- In a redis-managed pool of peers, the peer group is updated only after a timeout. This adds explicit unregistration so that in the case of a deliberately terminated instance, redis is notified immediately that the peer has dropped.
- Doesn't change the behavior of refreshing the peer list, which is still a pull-based timeout.
- Fixes #393 

- Add an Unregister method to redimem
- Add a done channel to the goroutines that keep redis updated
- Unregister when the done channel is closed
- Propagate the done channel through the app initialization
- Close the done channel on shutdown
- Update tests
- Add a new test to make sure that unregistration happens
- Also adds a new test case to rules; I wrote it to verify we didn't have a bug and thought it should probably stay in.

### DIFF
--- a/app/app_test.go
+++ b/app/app_test.go
@@ -1,5 +1,3 @@
-//go:build all || race
-
 package app
 
 import (
@@ -123,7 +121,7 @@ func newStartedApp(
 
 	var err error
 	if peers == nil {
-		peers, err = peer.NewPeers(context.Background(), c)
+		peers, err = peer.NewPeers(context.Background(), c, make(chan struct{}))
 		assert.NoError(t, err)
 	}
 
--- a/cmd/refinery/main.go
+++ b/cmd/refinery/main.go
@@ -106,7 +106,8 @@ func main() {
 
 	ctx, cancel := context.WithTimeout(context.Background(), c.GetPeerTimeout())
 	defer cancel()
-	peers, err := peer.NewPeers(ctx, c)
+	done := make(chan struct{})
+	peers, err := peer.NewPeers(ctx, c, done)
 
 	if err != nil {
 		fmt.Printf("unable to load peers: %+v\n", err)
@@ -226,5 +227,8 @@ func main() {
 
 	// block on our signal handler to exit
 	sig := <-sigsToExit
+	// unregister ourselves before we go
+	close(done)
+	time.Sleep(100 * time.Millisecond)
 	a.Logger.Error().Logf("Caught signal \"%s\"", sig)
 }
--- a/internal/peer/redis.go
+++ b/internal/peer/redis.go
@@ -45,7 +45,7 @@ type redisPeers struct {
 }
 
 // NewRedisPeers returns a peers collection backed by redis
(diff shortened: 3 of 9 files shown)


app/app_test.go
```diff
@@ -1,5 +1,3 @@
-//go:build all || race
```
kentquirk, line 1: We don't need these build flags any more so I removed them.

cmd/refinery/main.go
```diff
@@ -226,5 +227,8 @@ func main() {
 
 	// block on our signal handler to exit
 	sig := <-sigsToExit
+	// unregister ourselves before we go
+	close(done)
+	time.Sleep(100 * time.Millisecond)
```
TylerHelmuth, line 232: Do we sleep here to give time for the unregistering to occur?  Is it possible that this amount of time would ever not be enough?  Would be great if we would block until we are 100% sure we've unregistered and then continue on.

internal/peer/redis.go
```diff
@@ -135,15 +135,24 @@ func (p *redisPeers) RegisterUpdatedPeersCallback(cb func()) {
...
+		case <-tk.C:
+			ctx, cancel := context.WithTimeout(context.Background(), p.c.GetPeerTimeout())
+			// every interval, insert a timeout record. we ignore the error
+			// here since Register() logs the error for us.
+			p.store.Register(ctx, p.publicAddr, peerEntryTimeout)
+			cancel()
+		case <-done:
+			// unregister ourselves
+			ctx, cancel := context.WithTimeout(context.Background(), p.c.GetPeerTimeout())
+			p.store.Unregister(ctx, p.publicAddr)
```
TylerHelmuth, line 152: It looks like `Unregister` can return an error with logging it itself, so we care about handling those errors here?

TylerHelmuth, line 152: It looks like `Unregister` can return an error with logging it itself, so we care about handling those errors here?

kentquirk: We're shutting down, so I chose to define it in the interface for more general use, but ignore it in this case. I can't really think of a situation where this error matters on shutdown.

cmd/refinery/main.go
kentquirk: Yes, that's why the sleep. There's a way to make it safe, but it's really complicated to make it perfectly correct (channels of channels with complicated deadlock prevention) -- and the difference here is between other nodes seeing that you shut down in N seconds if it worked, and 3N seconds if it didn't work. I've actually implemented this before and I don't think the complexity is worth it.

TylerHelmuth (approved):

PR #570, opened 2022-11-28
feat: Implement alternative sharding using rendezvous hash to improve dynamic scalability

kentquirk:

The legacy code in Refinery does a deterministic sharding calculation; for every span that comes in, we hash the traceID with SHA1, truncate to generate a 32-bit number, then divide the range by N if there are N shards. If there are 4 shards, the lowest values go to shard 0, the next block goes to shard 1, etc. 

The problem is that if the number of shards changes from 4 to 5, 1/5 of the shards that would have gone to 0 now go to 1. 2/5 of shard 1 now goes to shard 2; 3/5 of shard 2 (more than half!) goes to shard 3, and so forth. On average, half the shards move.

The practical result of this is that when a span arrives just after a change in shard count, there’s a chance that it will end up on a different shard from other spans in the same trace that arrived earlier. The two shards might make different sampling decisions, and therefore we could end up with missing spans in Honeycomb. 

A technique called "[Rendezvous Hashing](https://en.wikipedia.org/wiki/Rendezvous_hashing)" uses a double hash — both the traceID and the shardID are hashed, and then we do a calculation to find which trace should go with which shard. 

There are several algorithms for doing this; all the algorithms seem to be roughly similar in terms of their ability to balance the load among shards. The randomness of traceIDs can lead to variation anyway, so it’s mainly a probabalistic effect.

With one of these algorithms, adding or dropping a new shard usually affects approximately 1/N of the existing traces, where N is the number of shards. This is far better than half! Sometimes, random chance will be unlucky, and it might be closer to 2/N, but in general it's still quite a bit better than the legacy algorithm.

Implementing the new algorithm reduces the probability of being on a different shard from 1/2 to 1/N, which is most meaningful for large installations where N might be 15, 40, or even something like 120. This seems worthwhile.

Note that the point of these algorithms is to allow shard assignment decisions to be made independently by different shards yet generate the same results without requiring any synchronization between shards.

It should be noted that while the legacy sharder is O(1) in its operation, the hash sharder is O(N) where N is the number of partitions (which is either 50 or the number of shards, whichever is greater). It still runs in under a microsecond even with 128 shards.

- Add configuration to choose legacy or hash sharding strategy (default to legacy)
- Update example config to explain it
- Add new sharding strategy
- Add mechanism to select a sharding strategy based on the config
- Add tests to show that it works 
- Add benchmarks to show that it's performant enough

### DIFF
--- a/sharder/deterministic.go
+++ b/sharder/deterministic.go
@@ -24,7 +24,6 @@ import (
 const (
 	shardingSalt        = "gf4LqTwcJ6PEj2vO"
 	peerSeed     uint64 = 6789531204236
-	traceSeed    uint64 = 5498616120353
 )
 
 // DetShard implements Shard
--- a/sharder/deterministic.go
+++ b/sharder/deterministic.go
@@ -267,21 +267,31 @@ func (d *DeterministicSharder) loadPeerList() error {
 	// make sure the list is in a stable, comparable order
 	sort.Sort(SortableShardList(newPeers))
 
-	// PartitionCount controls the minimum number of partitions used to control node assignment
-	// when we use the "hash" strategy.
-	// Ideally, this value should be 3-5x the maximum expected number of nodes.
-	// If necessary, it will (effectively) be increased to equal the number of nodes.
-	// A larger value will achieve more even trace distribution, at the cost of a minor performance
-	// hit once per span (roughly 1 microsecond per 50 partitions). Since the distribution isn't
-	// that even anyway (yay randomness) 50 seems to be a good balance; it helps with small
-	// installations and doesn't hurt for big ones.
+	// In general, the variation in the traffic assigned to a randomly partitioned space is
+	// controlled by the number of partitions. PartitionCount controls the minimum number
+	// of partitions used to control node assignment when we use the "hash" strategy.
+	// When there's a small number of partitions, the two-layer hash strategy can end up giving
+	// one partition a disproportionate fraction of the traffic. So we create a large number of
+	// random partitions and then assign (potentially) multiple partitions to individual nodes.
+	// We're asserting that if we randomly divide the space among at this many partitions, the variation
+	// between them is likely to be acceptable. (As this is random, there might be exceptions.)
+	// The reason not to make this value much larger, say 1000, is that finding the right partition
+	// is linear -- O(number of partitions) and so we want it to be as small as possible
+	// while still being big enough.
+	// PartitionCount, therefore, is the smallest value that we believe will yield reasonable
+	// distribution between nodes. We divide it by the number of nodes using integer division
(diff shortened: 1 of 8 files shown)


sharder/deterministic.go
```diff
@@ -11,16 +11,21 @@ import (
...
 
-// shardingSalt is a random bit to make sure we don't shard the same as any
-// other sharding that uses the trace ID (eg deterministic sampling)
-const shardingSalt = "gf4LqTwcJ6PEj2vO"
+// These are random bits to make sure we differentiate between different
+// hash cases even if we use the same value (traceID).
+const (
+	shardingSalt        = "gf4LqTwcJ6PEj2vO"
+	peerSeed     uint64 = 6789531204236
+	traceSeed    uint64 = 5498616120353
```
TylerHelmuth, line 27: Is this used anywhere?

sharder/deterministic.go
```diff
@@ -217,20 +262,41 @@ func (d *DeterministicSharder) loadPeerList() error {
...
+	// If necessary, it will (effectively) be increased to equal the number of nodes.
+	// A larger value will achieve more even trace distribution, at the cost of a minor performance
+	// hit once per span (roughly 1 microsecond per 50 partitions). Since the distribution isn't
+	// that even anyway (yay randomness) 50 seems to be a good balance; it helps with small
+	// installations and doesn't hurt for big ones.
+	const partitionCount = 50
+	// now build the hash list;
+	// We make a list of hash value and an index to a peer.
+	hashes := make([]hashShard, 0)
+	hashesPerPeer := partitionCount/len(peerList) + 1
```
TylerHelmuth, line 283: Does it matter that we are doing integer devision?

sharder/deterministic.go
```diff
@@ -217,20 +262,41 @@ func (d *DeterministicSharder) loadPeerList() error {
...
+	for ix := range newPeers {
+		hashes = append(hashes, newPeers[ix].GetHashesFor(ix, hashesPerPeer, peerSeed)...)
```
TylerHelmuth, line 285: Do we end up with `50 * len(newPeers)` unique hashes?  I admit it is my lack of knowledge about rendezvous hashing, but I am missing the reason why we need so many.  What would happen if `hashesPerPeer` was 1?

sharder/deterministic.go
kentquirk: Good catch; it was for a different variant of the algorithm but is not any longer.

sharder/deterministic.go
kentquirk: It's specifically the point to use integer division. I'll add some text to explain in more detail.

TylerHelmuth: Alright this makes a lot more sense now, thanks for the added comments.

TylerHelmuth (approved):
Super exceited for this change

PR #580, opened 2022-12-06
add early check for span matching rule

### LINKED ISSUE #434: optimize rule checks - break out of loop once match is found

_Posting from internal discussion:_

Looking at this [code](https://github.com/honeycombio/refinery/blob/153a398d13636722ccd03a2e65ed58d128a6cfbb/sample/rules.go#L67-L72) it appears to evaluate matches:

- for each rule
- for each condition in the rule
- for each span in the trace

and at the end it checks that the number of matches == the number of conditions
So conditions are evaluated independently across all spans in the trace, and then ANDed together.
Rules are ORed in that the first rule for which all conditions are matched will be the rule that is evaluated.

This could be optimized - instead of checking against every span, even after a match is made, break out if a match has been found.

add this below the `span:` label on [line 71](https://github.com/honeycombio/refinery/blob/main/sample/rules.go#L71). 

```go
if matched == len(rule.Condition) {
    break
}
```

**Additional context**

This code adjustment cannot be added as-is, so using this issue as a placeholder to work on it. (Error: `label span declared but not used.`)

fchikwekwe:

- Rules would continue to check every span every after a matching span was found. 
- Fixes #434

- Checks if matching span has been found and stops checking if it has.

### DIFF
--- a/sample/rules.go
+++ b/sample/rules.go
@@ -67,11 +67,20 @@ func (s *RulesBasedSampler) GetSampleRate(trace *types.Trace) (rate uint, keep b
 	for _, rule := range s.Config.Rule {
 		var matched bool
 		var reason string
+		var condition bool
 
 		switch rule.Scope {
 		case "span":
 			matched = ruleMatchesSpanInTrace(trace, rule, s.Config.CheckNestedFields)
 			reason = "rules/span/"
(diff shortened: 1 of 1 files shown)


kentquirk:
Turns out my idea was poorly thought through; we shouldn't do this.

Closes #434.

PR #584, opened 2022-12-23
feat: stress relief

kentquirk:
# Stress Relief

The problem this is designed to address is that refinery can become unstable under heavy load, and it goes from stable to unstable very quickly. A larger refinery cluster can be very hard to bring up -- we've seen what are effectively contagious restarts, where a given node crashes, wakes up, and then another node crashes, etc. 

The general idea here is "let's make refinery be more stable by allowing it to shed load quickly under stress". This PR implements a "StressRelief" system that, when activated, deterministically samples each span based only on its trace ID, without attempting to direct traces to the appropriate shard first. The configuration determines the sample rate used.

There is configuration to set this up, and it allows the "mode" to be configured as "never" (it's not active), "monitor" (it's active when the stress indicator is on), or "always" (always active). 

I tested this system with mode set to never and established a baseline instability level for a cluster of 3 refineries at a given "load index" (an arbitrary value indicating the amount of traffic). I found that nodes began restarting once the load index went over 600. When I set the mode to "always", the cluster was quite happy with a load of 1500, and there was no indication of stress -- it felt like it could go much higher, but I haven't yet tried higher numbers. 

The next problem was to get an accurate measurement of stress so that we can automatically turn this system on. Many hours of watching refinery crash led to the realization that the internal queues seem to be critical, as does memory usage. 

The desired strategy is to measure some key metrics and compare them with some defined threshold values on an appropriately adjusted scale from 0 to 1. We can then watch to see when any of them starts to rise too close to the threshold, and turn on the stress relief (and also give the user some indication of which metric is the biggest problem)

The StressRelief system uses a hysteresis model where it turns on at a much higher level than it turns off -- the current defaults are 50 and 20 (user-adjustable, of course). 

The particular metrics to watch is not currently user-adjustable, although they're coded to make that possible without too much difficulty -- but it's a lot of added complexity in the config for probably little value. Most people won't have a good idea which things to worry about.

## Memory

The existing cache system tries to manage memory, particularly with the newer "impact" cache overrun strategy. But we can also use excessive RAM usage as a signal for stress.

## Queues
There are multiple queues that are tracked in metrics:

* "collector_peer_queue_length"
* "collector_incoming_queue_length"
* "libhoney_peer_queue_length"
* "libhoney_upstream_queue_length"

Once any of these queue lengths rise above zero, it means that refinery is not processing data as quickly as it is arriving, and in a high-volume situation, it doesn't take long for that to overwhelm refinery and crash it. But the length of the queues is definitely a factor, and queue length is not part part of current refinery metrics.

The problem is that the libhoney peer queue lengths were not available to refinery in its current form -- they are metrics reported through the Metrics object given to libhoney. Furthermore, none of the queue sizes were currently reported, and most of them are defined through some mathematical expression based on a configuration value.

The options considered were:
* modifying libhoney to allow querying this information, which we'd rather not do -- libhoney is a low-level library that we'd rather not touch.
* modifying metrics to add features that allow reporting all this stuff

Some features were added to metrics:
* `Get(name) float64` which returns the current value of a non-histogram metric as a float
* `Store(name, value)` that stores a value under a name (this allows metrics to be used as a key-value store). Values stored this way can be retrieved by Get. It's like a gauge but isn't reported as part of the metrics reporting.

This was all made more complicated by the fact that we don't use a single Metrics object for all of refinery, but in fact instantiate at least 3 different ones. The metrics object had to be wrapped (with MetricsNamer) in order to unify it.


TylerHelmuth (commented):
@kentquirk I'll do an indepth review tomorrow, but is there any possibility of doing 1 PR specifically to add the new metrics and 1 PR to use them for stress relief?

kentquirk:
Closing in favor of two separate PRs.

Issue #541, opened 2022-10-18
Give rules a way to coerce input data to match the rule condition


kentquirk:
**Is your feature request related to a problem? Please describe.**

Refinery's rules engine has conditions that look like this:

```
    [[production.rule.condition]]
      field = "http.status_code"
      operator = "="
      value = 200
```

The problem is that different backends can and do send status code as "200" instead of 200. This will cause this condition to fail in refinery, but then, assuming the data is sent to Honeycomb, the Honeycomb backend would then coerce "200" to 200 (assuming the field is defined as an integer in the dataset schema). 

A team using a mixed technology stack can (and has!) run into this problem, and looking at the data in Honeycomb yields no clue, since the different datatypes have been coerced on ingestion. 

**Describe the solution you'd like**

Expand the conditions to add another element:
```
    [[production.rule.condition]]
      field = "http.status_code"
      operator = "="
      value = 200
      datatype = "int"
```

If the datatype is specified, then the value will be stored as the given datatype, and all input data will be coerced to the specified datatype before the comparison.

Like Honeycomb's backend:
* the datatypes will be limited to `int`, `float`, `string`, and `bool`.
* conversion failure results in the compare function returning false (no matter what operator is used)
* existence tests are not affected by the datatype
* bonus feature: we could add a new operator `hastype` that would compare the type of the incoming field to one of the values above.

For backwards compatibility, if datatype is not specified, no change to existing behavior will occur. 

**Describe alternatives you've considered**

Currently, if this problem is occurring, debugging is difficult, and the only solution is to add additional rules to handle the other datatypes. 

We could enable additional config that would generate debugging info about incoming datatypes, but this expands an already complex configuration in addition to the additional rule complexity.

**Additional context**

JamieDanielson:
Does this supersede #452 ? If so we can close that one as this has much more detail.

kentquirk:
addressed by #585

PR #585, opened 2023-01-10
feat: allow user to convert datatype if valid

fchikwekwe:

- When datatype is specified, attempt to convert span data to given data type. #541 

- This change allows a data check to the rules condition to see if input span/trace data matches the type specified on the data condition. If it is not, then attempt to convert the data to specified type (e.g. int to float, string to int, etc). If it cannot be converted, then do not convert the data.


config/sampler_config.go
```diff
@@ -45,12 +47,379 @@ type RulesBasedSamplerCondition struct {
...
+		case "starts-with":
+			return strings.HasPrefix(s, conditionValue)
+		case "contains":
+			return strings.Contains(s, conditionValue)
+		case "does-not-contain":
+			return !strings.Contains(s, conditionValue)
+		}
+		return false
+	}
+	return nil
```
TylerHelmuth, line 420: nit: similar to the CompareOperators can we do this switch in `setMatchStringBasedOperators` and split each case into its own match function?

config/sampler_config.go
```diff
@@ -45,12 +47,379 @@ type RulesBasedSamplerCondition struct {
...
+					return n <= conditionValue
+				}
+				return false
+			}
+			return nil
+		}
+	case bool:
+		// check if conditionValue and spanValue are not equal
+		switch condition {
+		case "!=":
```
TylerHelmuth, line 375: Similar to strings we should handle invalid conditions.

config/sampler_config.go
```diff
@@ -45,12 +47,379 @@ type RulesBasedSamplerCondition struct {
...
+					return n == conditionValue
+				}
+				return false
+			}
+			return nil
+		}
+	default:
+		return fmt.Errorf("%s value's type must be either string, int, float or bool, but was '%T'", conditionValue, conditionValue)
+	}
+	return nil // what return was missing?
```
TylerHelmuth, line 395: I think if we made it here there is a problem with a type's inner switch statement.

config/sampler_config.go
```diff
@@ -45,12 +47,379 @@ type RulesBasedSamplerCondition struct {
...
+		}
+	case "does-not-contain":
+		err := setMatchStringBasedOperators(r, "does-not-contain")
+		if err != nil {
+			return err
+		}
+	default:
+		return fmt.Errorf("unknown operator '%s'", r.Operator)
+	}
+	return nil
```
TylerHelmuth, line 102: nit: first time reading through I got confused how this return could be hit but then I saw that `exists` and `not exists` don't return.  Since we have a default case I'd opt for `exists` and `not exists` returning nil so that the default case is the guaranteed return statement.

config/sampler_config.go
```diff
@@ -45,12 +47,379 @@ type RulesBasedSamplerCondition struct {
...
+			}
+			return nil
+		}
+	case int:
+		// check if conditionValue and spanValue are not equal
+		switch condition {
+		case "!=":
+			r.Matches = func(spanValue any, exists bool) bool {
+				if n, ok := tryConvertToInt(spanValue); exists && ok {
+					return n != conditionValue
```
TylerHelmuth, line 222: All the match functions end up being the same except for the comparator.  I think we may be able to take advantage of some generics/extra functions to reduce the duplicated code.  Take a look at what @kentquirk did with a [similar comparitor problem in OTTL](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/ottl/compare.go).

config/sampler_config.go
```diff
@@ -45,12 +47,379 @@ type RulesBasedSamplerCondition struct {
...
+	return r.setMatchesFunction()
 }
 
 func (r *RulesBasedSamplerCondition) String() string {
 	return fmt.Sprintf("%+v", *r)
 }
 
+func (r *RulesBasedSamplerCondition) setMatchesFunction() error {
+	switch r.Operator {
+	case "exists":
```
TylerHelmuth, line 64: nit: The strings for these operators now live here and in https://github.com/honeycombio/refinery/blob/e7f9bf393fc884e0549c11c1ea73bd30249910c6/sample/rules.go#L195. Would be nice to start using constants/enums

sample/rules.go
```diff
@@ -126,40 +135,53 @@ func (s *RulesBasedSampler) GetSampleRate(trace *types.Trace) (rate uint, keep b
...
 	var matched int
 
-	for _, condition := range rule.Condition {
+	for _, condition := range rule.Conditions {
 	span:
 		for _, span := range t.GetSpans() {
 			value, exists := extractValueFromSpan(span, condition, checkNestedFields)
-
-			if conditionMatchesValue(condition, value, exists) {
+			if condition.Matches == nil {
```
TylerHelmuth, line 148: Under what conditions will `condition.Matches` be nil?  If the user doesn't supply a `Datatype` does `.Matches` get a default match function or is it nil?

config/sampler_config.go
```diff
@@ -45,12 +47,379 @@ type RulesBasedSamplerCondition struct {
 	Field    string
 	Operator string
 	Value    interface{}
+	Datatype string
```
TylerHelmuth, line 50: I expected `DataType` to come into play with the match functions attempted type conversion but I don't see it being used.  Instead it looks like we attempt to convert the incoming data to the Value's data type.  If thats the intention then I think we can remove DataType.

TylerHelmuth (commented):
@fchikwekwe left a couple questions and a couple nits.  Great work!

sample/rules.go
fchikwekwe: Yeah, if there is not a datatype specified, then the matches function is nil.

config/sampler_config.go
fchikwekwe: Yeah, I did talk about doing something like this with @kentquirk before this project ended up dragging a bit. I think we decided to push this change first and then make that change as a refactor in a second PR.

config/sampler_config.go
fchikwekwe: True, this is better.

config/sampler_config.go
fchikwekwe: Right, I think I forgot to remove this comment. Good catch.

config/sampler_config.go
fchikwekwe: Yeah, this makes it more consistent. I added this change in `setMatchStringBasedOperators`.

config/sampler_config.go
fchikwekwe: I pushed a version with DataType removed.

config/sampler_config.go
TylerHelmuth: I meant to say `also` not `now.  Sorry about that.

config/sampler_config.go
kentquirk: We talked at some length yesterday and Datatype is required because without it, this PR would be a breaking change.

config/sampler_config.go
kentquirk: Tyler is right here -- it would be worthwhile to define some constants like:

```
const (
     EQ = "="
     NE = "!="
     LT = "<"
...
```

This helps to avoid silly but common copy-paste errors.

config/sampler_config.go
```diff
@@ -45,12 +47,379 @@ type RulesBasedSamplerCondition struct {
...
+		return false
+	}
+}
+
+func setCompareOperators(r *RulesBasedSamplerCondition, condition string) error {
+	switch conditionValue := r.Value.(type) {
+	case string:
+		// check if conditionValue and spanValue are not equal
+		switch condition {
+		case "!=":
```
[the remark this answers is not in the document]
kentquirk: Inequalities on strings should be permitted (they are in the current refinery, even though you can't do them in the honeycomb UI). Faith has added them below.

config/sampler_config.go
```diff
@@ -45,12 +47,361 @@ type RulesBasedSamplerCondition struct {
...
+		if err == nil {
+			return n, true
+		}
+		return 0, false
+	default:
+		return 0, false
+	}
+}
+
+func tryConvertToInt64(v any) (int64, bool) {
```
kentquirk, line 108: You can get rid of this now.

config/sampler_config.go
kentquirk: As coded in this PR, we say "if it's not exactly equal to 'true' it must be false", and don't have any error cases

Just to complicate matters:
* JSON and TOML accept only true and false, lowercase only
* YAML accepts any case of true and false
* Go's ParseBool accepts 1, t, T, TRUE, true, True, 0, f, F, FALSE, false, False

The config parser we use can read JSON, TOML, and YAML. There's a lot of pressure to move our standard configuration to YAML instead of TOML. I'm considering doing that as part of a 2.0 release, and possibly changing the configuration parser while we're at it.

I recommend that we switch to the Go behavior (use ParseBool) for the cases where Datatype is specified, and don't change the behavior for existing configuration.

config/sampler_config.go
```diff
@@ -45,12 +47,361 @@ type RulesBasedSamplerCondition struct {
...
+
+func setMatchStringBasedOperators(r *RulesBasedSamplerCondition, condition string) error {
+	conditionValue, ok := tryConvertToString(r.Value)
+	if !ok {
+		return fmt.Errorf("%s value must be a string, but was '%s'", condition, r.Value)
+	}
+
+	switch condition {
+	case "starts-with":
+		// datatype := r.datatype
```
kentquirk, line 376: delete spurious comment

sample/rules_test.go
```diff
@@ -1060,3 +1060,463 @@ func TestRuleMatchesSpanMatchingSpan(t *testing.T) {
 		})
 	}
 }
+
+func TestRulesDatatypes(t *testing.T) {
```
kentquirk, line 1064: We basically need an entirely new set of tests that includes Datatype in the conditions, that exercise converting the different input types to the appropriate datatype. 

I suggest writing a couple of helper functions to build Conditions and Spans.

kentquirk (commented):
It's getting close. Needs more testing.

config/sampler_config.go
fchikwekwe: yes, done.

config/sampler_config.go
fchikwekwe: done, ty.

sample/rules_test.go
fchikwekwe: Sorry, I added back the datatype information. Should resolve this.

config/sampler_config.go
TylerHelmuth: covered via https://github.com/honeycombio/refinery/issues/592

sample/rules.go
```diff
@@ -158,10 +173,18 @@ func ruleMatchesSpanInTrace(trace *types.Trace, rule *config.RulesBasedSamplerRu
...
+				if !conditionMatchesValue(condition, value, exists) {
+					ruleMatched = false
+					break // if any condition fails, we can't possibly succeed, so exit inner loop early
+				}
+			}
 
-			if !conditionMatchesValue(condition, value, exists) {
-				ruleMatched = false
-				break // if any condition fails, we can't possibly succeed, so exit inner loop early
+			if condition.Matches != nil {
```
TylerHelmuth, line 183: If we make it here we know `condition.Matches != nil` so I believe we replace this if statement with an else if statement.

TylerHelmuth (approved):
Left 1 nit, otherwise LGTM.

sample/rules.go
fchikwekwe: sweet. updated.

PR #587, opened 2023-01-17
chore: Change workflow to use Collections board

kentquirk:
This changes the workflow to use the Collections board instead of the Telemetry team.


vreynolds (approved):
😭 🐥

PR #593, opened 2023-01-31
feat: extend and unify metrics system

kentquirk:

In service of being able to monitor the performance of refinery from within refinery itself, this PR extends refinery's Metrics system to add a few things:

- Use only one metrics object rather than several strewn about the system
- Move the "prefix" feature of metrics to a wrapper (metricsnamer) that applies the prefix on metric creation rather than during metrics reporting.
- Add the ability of metrics to have updown metrics
- Give metrics the ability to store constants that are not reported

The result of these changes is to allow the upcoming stress relief system to accurately calculate stress based on the metrics the system is already (or soon to be) tracking.

This set of changes should be transparent to the operation of refinery -- all existing metrics should work just as they have already.


collect/collect.go
```diff
@@ -483,6 +498,7 @@ func (i *InMemCollector) processSpan(sp *types.Span) {
 		trace.SendBy = time.Now().Add(timeout)
 		trace.RootSpan = sp
 	}
+	i.Metrics.Increment("span_processed")
```
TylerHelmuth, line 501: Because of the deferred func will we be double-incrementing `span_processed`.

metrics/honeycomb.go
```diff
@@ -230,14 +238,21 @@ func (h *HoneycombMetrics) reportToHoneycomb(ctx context.Context) {
...
-				ev.AddField(PrefixMetricName(h.prefix, count.name), count.val)
+				ev.AddField(count.name, count.val)
 				count.val = 0
 				count.lock.Unlock()
 			}
 
+			for _, updown := range h.updowns {
+				updown.lock.Lock()
+				ev.AddField(updown.name, updown.val)
+				// updown.val = 0   // updowns are never reset to 0
```
TylerHelmuth, line 249: Can this be removed?

collect/collect.go
kentquirk: Yup! Good catch.

metrics/honeycomb.go
kentquirk: It's in similar cases for other kinds of metrics, because in some cases we *do* reset the metrics, so I copied the pattern. I'd like to leave it there for clarity.

kentquirk:
Let's wait until #596 lands before rebasing this one.

TylerHelmuth (approved):

PR #594, opened 2023-01-31
feat: Stress Relief system

kentquirk:
# Stress Relief

The problem this is designed to address is that refinery can become unstable under heavy load, and it goes from stable to unstable very quickly. A larger refinery cluster can be very hard to bring up -- we've seen what are effectively contagious restarts, where a given node crashes, wakes up, and then another node crashes, etc. 

The general idea here is "let's make refinery be more stable by allowing it to shed load quickly under stress". This PR implements a "StressRelief" system that, when activated, deterministically samples each span based only on its trace ID, without attempting to direct traces to the appropriate shard first. The configuration determines the sample rate used.

There is configuration to set this up, and it allows the "mode" to be configured as "never" (it's not active), "monitor" (it's active when the stress indicator is on), or "always" (always active). 

I tested this system with mode set to never and established a baseline instability level for a cluster of 3 refineries at a given "load index" (an arbitrary value indicating the amount of traffic). I found that nodes began restarting once the load index went over 600. When I set the mode to "always", the cluster was quite happy with a load of 1500, and there was no indication of stress -- it felt like it could go much higher, but I haven't yet tried higher numbers. 

The next problem was to get an accurate measurement of stress so that we can automatically turn this system on. Many hours of watching refinery crash led to the realization that the internal queues seem to be critical, as does memory usage. 

The desired strategy is to measure some key metrics and compare them with some defined threshold values on an appropriately adjusted scale from 0 to 1. We can then watch to see when any of them starts to rise too close to the threshold, and turn on the stress relief (and also give the user some indication of which metric is the biggest problem)

The StressRelief system uses a hysteresis model where it turns on at a much higher level than it turns off -- the current defaults are 50 and 20 (user-adjustable, of course). 

The set of particular metrics to watch is not currently user-adjustable, although they're coded to make that possible without too much difficulty. But it's a lot of added complexity in the config for probably little value, since most people won't have a good idea which things they should worry about.

## Memory

The existing cache system tries to manage memory, particularly with the newer "impact" cache overrun strategy. But we can also use excessive RAM usage as a signal for stress.

## Queues
There are multiple queues that are tracked in metrics:

* "collector_peer_queue_length"
* "collector_incoming_queue_length"
* "libhoney_peer_queue_length"
* "libhoney_upstream_queue_length"

Once any of these queue lengths rise above zero, it means that refinery is not processing data as quickly as it is arriving, and in a high-volume situation, it doesn't take long for that to overwhelm refinery and crash it. But the configured length of the queues is definitely a factor.

### DIFF
--- a/collect/collect.go
+++ b/collect/collect.go
@@ -525,18 +525,18 @@ func (i *InMemCollector) processSpan(sp *types.Span) {
 		trace.SendBy = time.Now().Add(timeout)
 		trace.RootSpan = sp
 	}
-	i.Metrics.Increment("span_processed")
 }
 
-// ProcessSpanImmediately is an escape hatch used under stressful conditions -- it
-// submits a span for immediate transmission without enqueuing it for normal
+// ProcessSpanImmediately is an escape hatch used under stressful conditions --
+// it submits a span for immediate transmission without enqueuing it for normal
 // processing. This means it ignores dry run mode and doesn't build a complete
 // trace context or cache the trace in the active trace buffer. It only gets
 // called on the first span for a trace under stressful conditions; we got here
 // because the StressRelief system detected that this is a new trace AND that it
 // is being sampled. Therefore, we also put the traceID into the sent traces
 // cache as "kept".
-// It doesn't do any logging either; this is about as minimal as we can make it.
+// It doesn't do any logging and barely touches metrics; this is about as
+// minimal as we can make it.
 func (i *InMemCollector) ProcessSpanImmediately(sp *types.Span, keep bool, sampleRate uint, reason string) {
 	now := time.Now()
 	trace := &types.Trace{
--- a/collect/collect.go
+++ b/collect/collect.go
@@ -27,6 +27,9 @@ type Collector interface {
 	// scheduled for transmission.
 	AddSpan(*types.Span) error
 	AddSpanFromPeer(*types.Span) error
+	Stressed() bool
+	GetStressedSampleRate(traceID string) (rate uint, keep bool, reason string)
+	ProcessSpanImmediately(sp *types.Span, keep bool, sampleRate uint, reason string)
 }
 
 func GetCollectorImplementation(c config.Config) Collector {
@@ -61,6 +64,7 @@ type InMemCollector struct {
 	Transmission   transmit.Transmission  `inject:"upstreamTransmission"`
 	Metrics        metrics.Metrics        `inject:"genericMetrics"`
 	SamplerFactory *sample.SamplerFactory `inject:""`
+	StressRelief   StressReliever         `inject:"stressRelief"`
 
 	// For test use only
 	BlockOnAddSpan bool
@@ -89,6 +93,7 @@ func (i *InMemCollector) Start() error {
 		return err
 	}
 	i.cache = cache.NewInMemCache(imcConfig.CacheCapacity, i.Metrics, i.Logger)
+	i.StressRelief.UpdateFromConfig(i.Config.GetStressReliefConfig())
 
 	// listen for config reloads
 	i.Config.RegisterReloadCallback(i.sendReloadSignal)
@@ -99,6 +104,8 @@ func (i *InMemCollector) Start() error {
 	i.Metrics.Register("collector_peer_queue_length", "gauge")
 	i.Metrics.Register("collector_incoming_queue_length", "gauge")
 	i.Metrics.Register("collector_peer_queue", "histogram")
+	i.Metrics.Register("stress_level", "gauge")
+	i.Metrics.Register("stress_relief_activated", "gauge")
 	i.Metrics.Register("collector_cache_size", "gauge")
 	i.Metrics.Register("memory_heap_allocation", "gauge")
 	i.Metrics.Register("span_received", "counter")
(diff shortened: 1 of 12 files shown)


kentquirk:
I'll resolve conflicts etc after #593 is merged.

collect/collect.go
```diff
@@ -498,6 +525,40 @@ func (i *InMemCollector) processSpan(sp *types.Span) {
 		trace.SendBy = time.Now().Add(timeout)
 		trace.RootSpan = sp
 	}
+	i.Metrics.Increment("span_processed")
```
TylerHelmuth, line 528: Duplicate incrementing again.

collect/collect.go
```diff
@@ -498,6 +525,40 @@ func (i *InMemCollector) processSpan(sp *types.Span) {
...
+// ProcessSpanImmediately is an escape hatch used under stressful conditions -- it
+// submits a span for immediate transmission without enqueuing it for normal
+// processing. This means it ignores dry run mode and doesn't build a complete
+// trace context or cache the trace in the active trace buffer. It only gets
+// called on the first span for a trace under stressful conditions; we got here
+// because the StressRelief system detected that this is a new trace AND that it
+// is being sampled. Therefore, we also put the traceID into the sent traces
+// cache as "kept".
+// It doesn't do any logging either; this is about as minimal as we can make it.
+func (i *InMemCollector) ProcessSpanImmediately(sp *types.Span, keep bool, sampleRate uint, reason string) {
```
TylerHelmuth, line 540: Do we need to do anything with `i.Metrics.Increment("span_processed")` in this function? Or any other metrics?

collect/stressRelief.go
```diff
@@ -0,0 +1,307 @@
...
+	algorithms map[string]func(string, string) float64
+	calcs      []StressReliefCalculation
+	lock       sync.RWMutex
+}
+
+func (s *StressRelief) Start() error {
+	s.Logger.Debug().Logf("Starting StressRelief system")
+	defer func() { s.Logger.Debug().Logf("Finished starting StressRelief system") }()
+
+	s.algorithms = map[string]func(string, string) float64{
```
TylerHelmuth, line 71: Instead of mapping from algorithm name to algorithm function here, can the `calcs` struct be instantiated with the algorithm function itself?

```go
	s.calcs = []StressReliefCalculation{
		{Numerator: "collector_peer_queue_length", Denominator: "PEER_CAP", Algorithm: s.sqrt, Reason: "CacheCapacity (peer)"},
		{Numerator: "collector_incoming_queue_length", Denominator: "INCOMING_CAP", Algorithm: s.sqrt, Reason: "CacheCapacity (incoming)"},
		{Numerator: "libhoney_peer_queue_length", Denominator: "PEER_BUFFER_SIZE", Algorithm: s.sqrt, Reason: "PeerBufferSize"},
		{Numerator: "libhoney_upstream_queue_length", Denominator: "UPSTREAM_BUFFER_SIZE", Algorithm: s.sqrt, Reason: "UpstreamBufferSize"},
		{Numerator: "memory_heap_allocation", Denominator: "MEMORY_MAX_ALLOC", Algorithm: s.sigmoid, Reason: "MaxAlloc"},
	}
```

collect/stressRelief.go
```diff
@@ -0,0 +1,307 @@
...
+		"square":  s.square,
+		"sigmoid": s.sigmoid,
+	}
+
+	s.calcs = []StressReliefCalculation{
+		{Numerator: "collector_peer_queue_length", Denominator: "PEER_CAP", Algorithm: "sqrt", Reason: "CacheCapacity (peer)"},
+		{Numerator: "collector_incoming_queue_length", Denominator: "INCOMING_CAP", Algorithm: "sqrt", Reason: "CacheCapacity (incoming)"},
+		{Numerator: "libhoney_peer_queue_length", Denominator: "PEER_BUFFER_SIZE", Algorithm: "sqrt", Reason: "PeerBufferSize"},
+		{Numerator: "libhoney_upstream_queue_length", Denominator: "UPSTREAM_BUFFER_SIZE", Algorithm: "sqrt", Reason: "UpstreamBufferSize"},
+		{Numerator: "memory_heap_allocation", Denominator: "MEMORY_MAX_ALLOC", Algorithm: "sigmoid", Reason: "MaxAlloc"},
```
TylerHelmuth, line 83: It looks like only sqrt and sigmoid are used, do we need definitions for linear and square?

collect/stressRelief.go
kentquirk: Yes, but I specifically avoided that so that this is easier to make configurable. For now, I don't want to add to the configuration load, but I thought the extra indirection was worth it. 

If you disagree, I could be convinced.

collect/collect.go
kentquirk: Hmm. I hope I didn't screw up a merge.

collect/stressRelief.go
kentquirk: Again, this is intended for future configuration and tuning. I will add some more comments to it.

collect/collect.go
kentquirk: Ah, I reverted it when I split this into two PRs. No big deal. Thanks for catching it again!

collect/stressRelief.go
TylerHelmuth: Are you anticipating something in the config file that allows specifying the algorithm to use?

kentquirk: If we decide to allow configuration at all, it would essentially be filling in that `calcs` array. You'd specify 2 metrics and an algorithm to combine them with.

TylerHelmuth (approved):

PR #596, opened 2023-01-31
fix: Fix flaky tests, clean up logic on rules

kentquirk:

- Some of the tests for rules were dependent on random numbers. This fixes that, and also cleans up some of the datatype conversion rules to work in more cases

- Fix tests to check both the keep and rate values
- Fix logic in some tests
- Change `tryConvertToString` to just use sprintf("%v"), even for strings, for consistency
- fix string operators to use tryConvertToString everywhere

### DIFF
--- a/sample/rules_test.go
+++ b/sample/rules_test.go
@@ -1090,6 +1090,7 @@ func TestRulesDatatypes(t *testing.T) {
 				},
 			},
 			ExpectedKeep: true,
+			ExpectedRate: 10,
 		},
 		{
 			Rules: &config.RulesBasedSamplerConfig{
@@ -1118,6 +1119,7 @@ func TestRulesDatatypes(t *testing.T) {
 				},
 			},
 			ExpectedKeep: true,
+			ExpectedRate: 10,
 		},
 		{
 			Rules: &config.RulesBasedSamplerConfig{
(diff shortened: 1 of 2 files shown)


sample/rules_test.go
```diff
@@ -1392,12 +1403,13 @@ func TestRulesDatatypes(t *testing.T) {
 				{
 					Event: types.Event{
 						Data: map[string]interface{}{
-							"test": 9.3,
+							"test": 9.3, // "9.3" is greater than "10.3" when compared as a string
```
TylerHelmuth, line 1406: 👀

TylerHelmuth (approved):

sample/rules_test.go
cartermp: Huh. Turns out it's consistent with F# too, so this isn't just some "rob pike is a weird dude" behavior:

```
λ ~/scratch/otel-pixie/ dotnet fsi

Microsoft (R) F# Interactive version 12.4.0.0 for F# 7.0
Copyright (c) Microsoft Corporation. All Rights Reserved.

For help type #help;;

> "9.3" > "10.3";;
val it: bool = true
```

kentquirk: Yeah, alphabetically '9' is greater than '1'. :)

PR #604, opened 2023-02-03
perf: Improve performance of stress relief

kentquirk:

In testing stress relief, I found that it could turn off too easily, and it might then cause refinery to crash. 

- Don't let stress relief turn off until stress levels are below the deactivation level for 10 consecutive seconds
- Move the calculation of stress relief into the periodic stress check, rather than doing it on every span. Now each span check needs only a read lock.
- Tweak the math for sigmoid (it had a little slop that was bugging me).

### DIFF
--- a/collect/stressRelief.go
+++ b/collect/stressRelief.go
@@ -224,8 +224,10 @@ func (s *StressRelief) square(num, denom string) float64 {
 // under the presumption that if we're using less than half of RAM,
 func (s *StressRelief) sigmoid(num, denom string) float64 {
 	r := s.ratio(num, denom)
-	// this is an S curve from 0 to 1, centered around 0.5 -- determined
-	// by messing around with a graphing calculator
+	// This is an S curve from 0 to 1, centered around 0.5 -- constants were
+	// empirically determined by messing around with a graphing calculator. The
+	// only reason you might change these is if you want to change the bendiness
+	// of the S curve.
 	stress := 0.400305589*math.Atan(6*(r-0.5)) + 0.5
 	s.Logger.Debug().
 		WithField("algorithm", "sigmoid").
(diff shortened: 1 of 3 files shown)


collect/stressRelief.go
```diff
@@ -226,7 +226,7 @@ func (s *StressRelief) sigmoid(num, denom string) float64 {
 	r := s.ratio(num, denom)
 	// this is an S curve from 0 to 1, centered around 0.5 -- determined
 	// by messing around with a graphing calculator
-	stress := .395*math.Atan(6*(r-0.5)) + 0.5
+	stress := 0.400305589*math.Atan(6*(r-0.5)) + 0.5
```
TylerHelmuth, line 229: Any value in this scalar being defined in a constant somewhere?

kentquirk: I didn't really think so -- it's just designed to be an S-shaped curve from 0 to 1, and symmetric around 0.5. There's no reason to ever change it. Does it need more comments?

collect/stressRelief.go
```diff
@@ -261,22 +261,17 @@ func (s *StressRelief) Recalc() {
...
 	s.reason = reason
-	s.lock.Unlock()
-}
 
-func (s *StressRelief) StressLevel() uint {
-	s.lock.RLock()
-	defer s.lock.RUnlock()
-	return s.stressLevel
-}
+	if s.stressLevel >= s.deactivateLevel {
```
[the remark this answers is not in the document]
kentquirk: I put it out there deliberately, but on second thought, I can move it inside without causing problems, so I will.

collect/stressRelief.go
TylerHelmuth: Nah I just wanted to make sure it wouldn't get used to scale something else similarly in the future.

collect/stressRelief.go
```diff
@@ -266,22 +268,24 @@ func (s *StressRelief) Recalc() {
...
 	case Monitor:
+		// If it's off, should we activate it?
 		if !s.stressed && s.stressLevel >= s.activateLevel {
 			s.stressed = true
 			s.Logger.Info().WithField("stress_level", s.stressLevel).WithField("reason", s.reason).Logf("StressRelief has been activated")
 		}
+		// We want make sure that stress relief is below the deactivate level
+		// for a minimum time after the last time we said it should be, so
+		// whenever it's above that value we push the time out.
+		if s.stressLevel >= s.deactivateLevel {
```
TylerHelmuth, line 285: Do ww need to push out the time whenever `level > deactivateLimit` or only when currently stressed and `level > deactivateLimit`?

kentquirk: I added an extra test for isStressed.

TylerHelmuth (approved):

PR #607, opened 2023-02-10
refactor: Change Router to use TraceServer

TylerHelmuth:

- Updates Refinery to use the gRPC server pattern that Shepherd uses.  This will allow us to add support for metrics and logs proxying in the future.

- Adds new TraceServer that implements the `TraceServiceServer` interface.
- Unimplement `TraceServiceServer` for `Router`.


kentquirk (approved):
Good start. Thanks

PR #609, opened 2023-02-15
Add ability to proxy grpc metrics and logs

TylerHelmuth:

Refinery is able to proxy http traffic but not gRPC traffic

- added MetricServer and LogServer
- added grpcProxyClientConn with hard-coded endpoint to honeycomb
- enabled forwarding of metrics and logs to honeycomb over grpc including incoming headers

## Additional Info

This was started during Swarm 2023 hackathon.  Care needs to be taken when reviewing the new grpc client implementation.


kentquirk:
This is superceded by current work on logs.

PR #611, opened 2023-02-23
fix: do not send sample rate in dry run

### LINKED ISSUE #415: Don't send SampleRate in dry run mode
It seems quite wrong to send the sample rate when refinery is running in dry run mode. For example, it could mess up your `COUNT`s even though the data isn't actually being sampled.

fchikwekwe:
Fixes #415 

This issues prevents sending the actual sample rate in dry run mode, instead adding a metric called `meta.dryrun.sample_rate` so that a user can know what their sample rate in dry run mode would have been.


kentquirk (approved):
Code changes look good; let's fix up the PR title and description, please.

TylerHelmuth (approved):

fchikwekwe:
Thanks for the reviews. Updated PR title.

cartermp:
Nice! Love to see it

PR #614, opened 2023-02-24
feat: Add ability to set Redis database and prefix in config

### LINKED ISSUE #544: Redis peer prefix/namespace are not configurable
This is sort of already implemented as 
https://github.com/honeycombio/refinery/blob/main/internal/redimem/redimem.go#L39
```
type RedisMembership struct {
	// Prefix is a way of namespacing your group membership
	Prefix string
	Pool *redis.Pool
	RepeatCount int
}
```
but the value of the prefix is hardcoded to  `refinery` 
https://github.com/honeycombio/refinery/blob/main/internal/peer/redis.go#L94
```
peers := &redisPeers{
		store: &redimem.RedisMembership{
			Prefix: "refinery",
			Pool:   pool,
		},
		...
	}
```

Expose the prefix as a config under PeerManagement

```
[PeerManagement]
Type = "redis"
RedisPrefix = "refinery"
```

kentquirk:

In a blue/green deployment model, two refinery clusters might want to share the same instance of redis. It's also possible that multiple apps want to share one redis. This permits that by:
- allowing the redis "database" (an integer from 0-15) to be specified in the configuration
- allowing the redis "prefix" (a string used in the name of the keys stored in redis) to be set as well. It already existed but was hardcoded to "refinery".

- Add RedisDatabase and RedisPrefix to the PeerManagement section of config.
- Add tests
- Add documentation

Fixes #544

### DIFF
--- a/config_complete.toml
+++ b/config_complete.toml
@@ -219,6 +219,14 @@ Metrics = "honeycomb"
 # Not eligible for live reload.
 # RedisPassword = ""
 
+# RedisDatabase is an integer indicating the database number to use for the Redis instance storing
+# the peer membership. It might be useful to set this in any situation where multiple refinery clusters
+# or multiple applications want to share a single Redis instance.
+# If the environment variable 'REFINERY_REDIS_DATABASE' is set it takes
+# precedence and this value is ignored.
+# Default = 0
+# RedisDatabase = 1
+
 # UseTLS enables TLS when connecting to redis for peer cluster membership management, and sets the MinVersion to 1.2.
 # Not eligible for live reload.
 # UseTLS = false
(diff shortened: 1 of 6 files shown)


config_complete.toml
```diff
@@ -219,6 +219,14 @@ Metrics = "honeycomb"
 # Not eligible for live reload.
 # RedisPassword = ""
 
+# RedisDatabase is an integer indicating the database number to use for the Redis instance storing
```
TylerHelmuth, line 222: Can we add somewhere in this comment that the value must be between 0-15.  I didn't actually know that redis worked like this haha

TylerHelmuth (approved):

config_complete.toml
kentquirk: Done.

Issue #615, opened 2023-02-27
Feature: Allow easily identifying late spans


TylerHelmuth:
**Is your feature request related to a problem? Please describe.**

When Refinery decides to sample a trace the reason is added to the `meta.refinery.reason` field  for all spans and the trace is sent on.  The fact that this trace was sampled is remembered, but the reason is not (on purpose, for performance reasons).

Spans for that trace that arrive after the sampling decision is made are forwarded on, but not `meta.refinery.reason` is added.  We call these late spans, and they can be the source of confusion and frustration when looking at a trace.  It would be nice to be able to quickly identify late spans. 

**Describe the solution you'd like**

Add `meta.refinery.reason=late` to late spans that are sampled.

**Describe alternatives you've considered**

Add a new field specifically for lateness.

PR #623, opened 2023-03-03
feat: Add meta.refinery.reason to late spans

TylerHelmuth:

- Closes https://github.com/honeycombio/refinery/issues/615

- Adds a `meta.refinery.reason=late` to all late spans that are kept.


kentquirk (approved):
Thanks!

PR #630, opened 2023-03-05
feat: Add configuration for trace, span, and parent ID field names

### LINKED ISSUE #626: Make trace fields configurable
At least one customer has configured their dataset schema via honeycomb to use alternative fields for trace ID, parent ID, and span ID. But refinery has these fields hardcoded to specific values.

Make these fields configurable in Refinery -- ideally with an ordered list of possibilities.

Existing behavior -- but Refinery is inconsistent with Honeycomb at large.

**Additional context**
Proposed config model:

TraceIdFieldNames: ["trace.trace_id", "trace_id", "traceId"]
ParentIdFieldNames: ["trace.parent_id", "parent_id", "parentId"]
SpanIdFieldNames: ["trace.span_id", "span_id", "spanId"]

All these fields would default to just the "trace.*" form. Helper functions would walk the name list and return the value of the first matching field.

tdarwin:

#626 - Make trace fields configurable

This should make trace, span, and parent ID fields configurable, which will allow refinery to work where a client uses a non-standard trace ID field name.

Fixes #626.


app/app_test.go
```diff
@@ -117,6 +117,9 @@ func newStartedApp(
 		GetHoneycombAPIVal:                   "http://api.honeycomb.io",
 		GetInMemoryCollectorCacheCapacityVal: config.InMemoryCollectorCacheCapacity{CacheCapacity: 10000},
 		AddHostMetadataToTrace:               enableHostMetadata,
+		TraceIdFieldNames:								 		[]string{"trace.trace_id"},
+		SpanIdFieldNames:								 			[]string{"trace.span_id"},
```
kentquirk, line 121: Since refinery doesn't currently use the span_id name, let's take it out of this PR. If we do something later, we can add it then with no harm.

collect/collect.go
```diff
@@ -515,7 +515,7 @@ func (i *InMemCollector) processSpan(sp *types.Span) {
 	trace.AddSpan(sp)
 
 	// if this is a root span, send the trace
-	if isRootSpan(sp) {
+	if isRootSpan(sp, i.Config) {
```
kentquirk, line 518: instead of passing config in as a parameter, which feels inappropriate here, please make `isRootSpan` into a member function of InMemCollector and simply reference i.Config inside the function.

collect/collect.go
```diff
@@ -617,15 +617,18 @@ func mergeTraceAndSpanSampleRates(sp *types.Span, traceSampleRate uint, dryRunMo
...
-	if parentID == nil {
-		parentID = sp.Data["parentId"]
-		if parentID == nil {
-			// no parent ID present; it's a root span
-			return true
+func isRootSpan(sp *types.Span, c config.Config ) bool {
+	var parentID string
+	for _, ParentSpanId := range c.GetParentIdFieldNames() {
+		if pID, ok := sp.Data[ParentSpanId]; ok {
+			parentID = pID.(string)
```
kentquirk, line 624: Why are you asserting it as a string? Seems like the nil test here should work -- if not, why not? This is a semantic change so I'd want to be sure it's correct.

config_complete.toml
```diff
@@ -153,6 +153,33 @@ AdditionalErrorFields = [
 # Eligible for live reload.
 # CacheOverrunStrategy = "impact"
 
+############################
+## Event ID Field Names   ##
+############################
+
+# Set the field names to use for the event ID fields. These fields are used to
+# identify events that are part of the same trace. The default values are
+# "trace.trace_id", "trace.span_id", "trace.parent_id". These values are
```
kentquirk, line 162: You set the defaults properly above, but this explanation is incomplete; please add traceId and parentId to the defaults.

kentquirk (changes requested):
This is close! A few changes, please. Although I originally specified spanID, I think we should leave it off for now since it's unused; we can always add it later.

I had also thought about putting these configurations down a level under a parent key, but decided that too should wait for the config revamp. It's fine the way it is.

collect/collect.go
tdarwin: The new code seems to be working and no longer forces the string type in the function

kentquirk (approved):
This is great. Thank you.

vthacker:
🎉 Thank you all! This is useful to us

PR #638, opened 2023-03-16
maint: Add LICENSES dir

TylerHelmuth:

- Adds LICENSES folder with all required licenses/notices

- Adds licenses/notices to LICENSES folder
- Adds a check to the build job that ensures the LICENSES folder is up-to-date.


kentquirk:
Additional requests:
* Can you add an "update-licenses" Make target also, please, and maybe have the verify task say how to run it?
* Is just stuffing these into our repo enough, or do we also need to attach them to the built artifacts somehow?

TylerHelmuth:
> Can you add an "update-licenses" Make target also, please, and maybe have the verify task say how to run it?

Will do

> Is just stuffing these into our repo enough, or do we also need to attach them to the built artifacts somehow?

As I understand it, we do need to update some stuff about our build, mainly some adding some important labels to the image that points back to this folder.  I plan to do that in a separate PR.

PR #639, opened 2023-03-17
fix: correct syntax error in config_complete.toml

hazelweakly:

This PR fixes a small syntax error in the `config_complete.toml` file.
It's not a huge deal, but I tend to uncomment lines and go from there and didn't read the line closely enough before uncommenting it; I suspect another tired person would make a similar mistake :)

It makes the strings stringier

### DIFF
--- a/config_complete.toml
+++ b/config_complete.toml
@@ -573,7 +573,7 @@ MetricsReportingInterval = 3
 # enabled, once activated. This prevents oscillations.
 # Default = 10s
 # Eligible for live reload.
-# MinimumActivationDuration = 10s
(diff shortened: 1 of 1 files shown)


TylerHelmuth (commented):
@hazelweakly Thanks for taking care of this! I'm surprised it caused an error tho bc the original format should still be interpreted as a string. What error were you seeing?

kentquirk:
@hazelweakly is correct -- YAML will interpret this correctly, but TOML is more rigid.

`'toml: expected newline but got U+0073 's'' loading config from config_complete.toml with filetype `

Thanks for the PR!

PR #643, opened 2023-03-23
ci: add permissions

jsoref:

Workflows that require write permissions need to declare them, otherwise they will fail in repositories with more conservative default permissions (`read`)

- specify `pull-requests: write`


kentquirk:
Thanks, but we prefer a manual approve step.

jsoref:
This isn't about manual approval. It's about the fact that it's broken in forks w/ paranoid settings.

PR #653, opened 2023-04-02
chore: Update MetricsReportingInterval in config_complete.toml

tdarwin:

- Some folks are sending metrics to Honeycomb much more frequently than necessary.  30 seconds should give plenty of data to understand things happening with Refinery and reduce the number of events being sent to be less overwhelming

-  Changes the example `MetricsReportingInterval` in the Honeycomb Metrics config section to be `30` instead of `3`.


kentquirk (approved):
This change is fine in general, but I'll note that a long reporting interval is a problem during setup when you're trying to right-size memory and queue sizes; issues can and will arise and completely vanish between reporting periods.

When I've been working with customers lately, I ask them to set it to 1 second while we debug.

fchikwekwe (approved):

PR #654, opened 2023-04-02
feat: [V2] Rewrite config code without Viper

kentquirk:
This is bigger than I wanted it to be, but there were a lot of tests that needed to be updated. I did rework a number of tests to simplify the creation of sample configurations, and eliminating Viper required the creation of a couple of new functions.

I feel like this has significantly simplified the config code and made it easier to understand and maintain. Please read the comments below before reviewing the code.

Elements of this project:

- [x] Separate cmdline/env from configuration
- [x] Find/write appropriate libraries to do all the tasks of loading cmdline, applying defaults, applying the cmdline, and validating afterward
- [x] Replace viper with these new libraries within the same config structure
- [x] Do the automatic reload (I wasn't going to do it for this PR but it breaks lots of tests, so I'm adding it).

Out of scope for this PR:
* Change config internal structure and helper functions
* Change sample configs and test code to YAML

How to tell if we're done?
* Passes all its tests
* Existing config files can still be loaded

[Viper](https://github.com/spf13/viper) is a strongly opinionated configuration tool. That can make it easy to start building something, but because in this case "opinionated" means "there's one way to do it and no easy way to change that" it can (and has, in my experience) become more and more of a maintenance problem as a codebase ages.

Refinery needs a significant rework to its configuration model:
* Its many config values should be broken up into more sections of related items and fewer at the root level
* This will allow better unit testing without mocking the entire config in a single object
* The relationship between command line, environment variables, and configuration should be clear and unambiguous
* Default values should be clear and unambiguous
* Since some configuration is reloadable at runtime, how that happens should be clear
* We want to support loading config from a remote source
* We want YAML-first configuration so that it's easier to work in a Kubernetes world
* We want to enable config to be loaded from a URL
* Full validation of configuration should become possible, including extra or mistyped names (this PR will not include this but will enable it). 

Some of these things are possible with viper, but not all of them, and in practice Viper makes balancing these things difficult.

This PR introduces some new tools to enable us to remove Viper:

- Configuration defaults are applied by a new library (creasty/defaults) that simply examines `default` struct tags and applies them when called. 
- A new library (jessevdk/go-flags) is used to handle command line and environment variables and map them onto a struct. This enables clear control of a separate struct (`CmdEnv`) for these things.
- Items in the configuration that are tied to the CmdEnv struct are given the `cmdenv` struct tag in the configuration that identifies them.
- A reflection-based function applies these cmdenv tags to configuration values at the appropriate time (after the configuration is loaded and defaults are applied).
- We continue to use the existing validation library (go-playgroundvalidator) to express simple post-loading validations
- These tools also allow loading a config from a URL instead of a file
- There's a new config value `ConfigReloadInterval` that controls the timing of reloads; if the hash of the contents is the same, no reload occurs. This is a timed reload rather than one that depends on operating system notifications. A future PR will allow a cluster of refineries to share knowledge of a configuration change.

The result is that entries in configuration structs can have up to 3 primary tags:
* `default` sets default values
* `cmdenv` when they need to be configurable on the command line and in the environment
* `validate` expresses validation rules


config/file_config.go
```diff
@@ -92,336 +111,137 @@ type PrometheusMetricsConfig struct {
...
+	DeactivationLevel         uint     `default:"75" validate:"gte=0,lte=100"`
+	StressSamplingRate        uint64   `default:"1000" validate:"gte=1"`
+	MinimumActivationDuration Duration `default:"10s"`
+	StartStressedDuration     Duration `default:"3s"`
+}
+
+// newConfig does the work of creating and loading the start of a config object
+// from the given arguments; if args is nil, it uses the command line arguments.
+// It's used by both the main init as well as the reload code.
+func newConfig(opts *CmdEnv) (*fileConfig, error) {
```
TylerHelmuth, line 166: Should this be named `newFileConfig`?  Looks like its goal is to return a fileConfig struct vs `NewConfig` which returns the Config interface.

config/file_config.go
```diff
@@ -92,336 +111,137 @@ type PrometheusMetricsConfig struct {
...
+				cb()
 			}
+			f.mux.Unlock() // can't defer since the goroutine never ends
 		}
 	}
-	return nil
+}
+
+// Stop halts the monitor goroutine
+func (f *fileConfig) Stop() {
```
TylerHelmuth, line 241: Is this used anywhere?

config/file_config.go
```diff
@@ -435,603 +255,494 @@ func (f *fileConfig) GetListenAddr() (string, error) {
...
+	return f.mainConfig.Collector, nil
 }
 
 func (f *fileConfig) GetAllSamplerRules() (map[string]interface{}, error) {
 	samplers := make(map[string]interface{})
 
-	keys := f.rules.AllKeys()
-	for _, key := range keys {
-		parts := strings.Split(key, ".")
+	_, err := readConfigInto(samplers, f.opts.RulesLocation, f.opts)
```
TylerHelmuth, line 433: Is it necessary to read each time or can it be read once per unique hash like the other rules/config?

config/file_config.go
```diff
@@ -435,603 +255,494 @@ func (f *fileConfig) GetListenAddr() (string, error) {
...
-			i = &TotalThroughputSamplerConfig{}
-		default:
-			return nil, notfound, errors.New("No Sampler found")
-		}
+	config := make(map[string]any)
 
-		if sub := f.rules.Sub(dataset); sub != nil {
-			return i, t, sub.Unmarshal(i)
-		}
+	_, err := readConfigInto(&config, f.opts.RulesLocation, nil)
```
TylerHelmuth, line 466: Same question here, do we have to read each time?

tools/convert/main.go
```diff
@@ -0,0 +1,145 @@
+package main
+
+import (
```
TylerHelmuth, line 3: Does `tools/convert` need to be part of this PR or could it be added separately, either in parallel or after this PR?

cmd/refinery/main.go
```diff
@@ -53,10 +45,9 @@ func (g graphLogger) Debugf(format string, v ...interface{}) {
 }
 
 func main() {
-	var opts Options
-	flagParser := flag.NewParser(&opts, flag.Default)
-	if extraArgs, err := flagParser.Parse(); err != nil || len(extraArgs) != 0 {
-		fmt.Println("command line parsing error - call with --help for usage")
```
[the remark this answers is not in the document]
kentquirk: 👍

config/file_config.go
```diff
@@ -1,89 +1,108 @@
...
-	libhoney "github.com/honeycombio/libhoney-go"
-	"github.com/sirupsen/logrus"
-	viper "github.com/spf13/viper"
 )
 
+// In order to be able to unmarshal "15s" etc. into time.Duration, we need to
+// define a new type and implement MarshalText and UnmarshalText.
+// We want this to be just inside config, though, so the accessors will still
+// return time.Duration.
+type Duration time.Duration
```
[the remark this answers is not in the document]
kentquirk: I'm going to fix the comment, because it was ill-considered. I don't think we want accessors for every field.

config/file_config.go
```diff
@@ -1,89 +1,108 @@
...
+	Collector                 string                         `default:"InMemCollector" validate:"required,oneof= InMemCollector"`
+	Metrics                   string                         `default:"honeycomb" validate:"required,oneof= prometheus honeycomb"`
+	SendDelay                 Duration                       `default:"2s" validate:"required"`
+	BatchTimeout              Duration                       `default:"100ms"`
+	TraceTimeout              Duration                       `default:"60s" validate:"required"`
+	MaxBatchSize              uint                           `default:"500" validate:"required"`
+	SendTicker                Duration                       `default:"100ms" validate:"required"`
+	UpstreamBufferSize        int                            `default:"10_000" validate:"required"`
+	PeerBufferSize            int                            `default:"10_000" validate:"required"`
+	DebugServiceAddr          string                         ``
```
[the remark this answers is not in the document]
kentquirk: I do this because without them, gofmt doesn't align the struct tags across the empty fields. It makes it a lot easier to read the struct tags this way. 

It may be that in the next round, we move them into groups where they can be trailing fields and not break up the struct tags. For now I'd prefer to leave them this way.

config/file_config.go
```diff
@@ -92,336 +111,137 @@ type PrometheusMetricsConfig struct {
...
+	Mode                      string   `default:"never" validate:"required,oneof= always never monitor"`
+	ActivationLevel           uint     `default:"90" validate:"gte=0,lte=100"`
+	DeactivationLevel         uint     `default:"75" validate:"gte=0,lte=100"`
+	StressSamplingRate        uint64   `default:"1000" validate:"gte=1"`
+	MinimumActivationDuration Duration `default:"10s"`
+	StartStressedDuration     Duration `default:"3s"`
+}
+
+// newConfig does the work of creating and loading the start of a config object
+// from the given arguments; if args is nil, it uses the command line arguments.
```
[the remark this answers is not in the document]
kentquirk: No, this is an old comment that's now inaccurate. Thanks.

config/file_config.go
kentquirk: Not yet, but I think it will be as I further refactor things to allow a cleaner shutdown. Whenever I add a channel I want a Stop capability to be available.

config/file_config.go
```diff
@@ -92,336 +111,137 @@ type PrometheusMetricsConfig struct {
...
 	}
 
-	err = f.rules.Unmarshal(f.conf)
+	cfg.callbacks = make([]func(), 0)
+	cfg.errorCallback = errorCallback
 
-	if err != nil {
-		return err
-	}
+	go cfg.monitor()
```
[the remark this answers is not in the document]
kentquirk: I don't think so, because it would imply that hot reloading would become an option, and also means that a lot of tests would have to make explicit calls to Start(). I don't personally feel that cost is justified by the symmetry.

config/file_config.go
kentquirk: I had a justification for doing it this way (we need to read into different data types at different times). But now that you've poked me about it, I think I have a better idea. Thanks.

kentquirk:
Tests broke; will fix.

TylerHelmuth (approved):

TylerHelmuth (approved):

PR #658, opened 2023-04-06
feat: annotate incoming sample rate

### LINKED ISSUE #612: If spans come in with existing Sample Rate > 1, we should annotate them

A customer had traces that were matching a rule, but had different sample rates. We eventually figured out that it was due to upstream telemetry setting sample rate (head sampling) but it took a while to get there, because the field "meta.refinery.original_sample_rate" was blank.

This was because we only populate that field if we're going to change the upstream value -- but here Refinery's desired rate was 1, so we weren't changing it. 

We should either:
* Populate that field anytime it's not 0
* Populate some other field that indicates that it was set upstream

I suggest the first one.

Not being confused.

**Additional context**

fchikwekwe:
Closes #612 

This issue simply annotates all incoming spans that have an original sample rate.


TylerHelmuth (commented):
Is there a test that confirms original_sample_rate is not set if `sp.SampleRate == 0`?

collect/collect.go
```diff
@@ -595,10 +595,9 @@ func (i *InMemCollector) dealWithSentTrace(keep bool, sampleRate uint, spanCount
 
 func mergeTraceAndSpanSampleRates(sp *types.Span, traceSampleRate uint, dryRunMode bool) {
 	tempSampleRate := sp.SampleRate
-	if traceSampleRate != 1 {
-		// When the sample rate from the trace is not 1 that means we are
-		// going to mangle the span sample rate. Write down the original sample
-		// rate so that that information is more easily recovered
+	if sp.SampleRate != 0 {
+		// Write down the original sample rate so that that information
+		//is more easily recovered
```
kentquirk, line 600:
```suggestion
		// is more easily recovered
```

kentquirk (approved):
One silly formatting nit; otherwise looks good!

PR #664, opened 2023-04-13
fix: validate cache overrun strategy for stress relief mode

### LINKED ISSUE #637: StressRelief mode is not compatible with Legacy CacheOverrunStrategy
**Versions**
- Refinery: 1.20.0

**Steps to reproduce**

1. Configure StressRelief mode = `monitor` and configure CacheOverrunStrategy=legacy
2. Stress the system to active StressRelief mode.
3. Observe the cache size being reduced
4. Get stuck in permanent stress relief mode

**Additional context**

Observed this behavior while troubleshooting with a customer.  I believe the fix is to require CacheOverrunStrategy=impact when using stress relief.  We can enforce this in the config validation.

fchikwekwe:
Closes #637 

Validate that the cache overrun strategy is set to "impact" before allowing stress relief mode


config/file_config.go
```diff
@@ -350,6 +353,14 @@ func (f *fileConfig) validateGeneralConfigs() error {
 	default:
 		return fmt.Errorf("invalid CacheOverrunStrategy: '%s'", st)
 	}
+
+	stressReliefConfig := f.GetStressReliefConfig()
+	switch stressReliefConfig.Strategy {
```
TylerHelmuth, line 358: Instead of adding `Strategy` to the stress relief config struct can we use `f.GetCacheOverrunStrategy()` and validate that the result is `impact` when `f.GetStressReliefConfig().mode` is `monitor` or `always`?

TylerHelmuth (commented):
Please add a new test to `config_test.go` that covers the new validation.

config/file_config.go
fchikwekwe: yes, I overcomplicated looking at the cache overrun strategy. Fix incoming.

config/file_config.go
```diff
@@ -350,6 +350,15 @@ func (f *fileConfig) validateGeneralConfigs() error {
...
 		return fmt.Errorf("invalid CacheOverrunStrategy: '%s'", st)
 	}
+
+	switch st {
+	case "impact":
+		if f.GetStressReliefConfig().Mode == "never" || f.GetStressReliefConfig().Mode == "always" {
+			return fmt.Errorf("invalid CacheOverrunStrategy for StressReliefMode: '%s'", st)
+		}
+	default: // return error if cache overrun strategy not "impact" i.e. "legacy"
+		return fmt.Errorf("invalid CacheOverrunStrategy for StressReliefMode: '%s'", st)
```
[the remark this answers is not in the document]
fchikwekwe: Oh, that was an oversight. Fixed now.

config/file_config.go
```diff
@@ -350,6 +350,12 @@ func (f *fileConfig) validateGeneralConfigs() error {
 	default:
 		return fmt.Errorf("invalid CacheOverrunStrategy: '%s'", st)
 	}
+
+	if st != "impact" {
+		if f.GetStressReliefConfig().Mode == "monitor" || f.GetStressReliefConfig().Mode == "always" {
+			return fmt.Errorf("invalid CacheOverrunStrategy for StressReliefMode: '%s'", st)
+		}
+	}
```
TylerHelmuth, line 358:
```suggestion
	if st != "impact" && (f.GetStressReliefConfig().Mode == "monitor" || f.GetStressReliefConfig().Mode == "always"){
		return fmt.Errorf("invalid CacheOverrunStrategy for StressReliefMode: '%s'", st)
	}
```

fchikwekwe: Yeah I wrote it this way at first and did not prefer it, but I will change it if you prefer the longer conditional.

PR #665, opened 2023-04-13
fix: only decorate late spans when configured to do so

### LINKED ISSUE #642: Mark late spans only when AddRuleReasonToTrace is configured
Today it does it unconditionally.

fchikwekwe:
Closes #642 

Check that late arriving spans are configured with `AddRuleReasonToTrace` before marking them late.


config/file_config.go
```diff
@@ -350,6 +353,14 @@ func (f *fileConfig) validateGeneralConfigs() error {
 	default:
 		return fmt.Errorf("invalid CacheOverrunStrategy: '%s'", st)
 	}
+
+	stressReliefConfig := f.GetStressReliefConfig()
```
TylerHelmuth, line 357: Can you remove these changes and save them for the other PR?

TylerHelmuth (commented):
One cleanup request, otherwise LGTM.

config/file_config.go
fchikwekwe: Oh, not sure how this happened. Accidental. Should be fixed now.

TylerHelmuth (approved):

PR #666, opened 2023-04-14
fix: add hostname to span during stress relief mode

### LINKED ISSUE #641: Stress Relief doesn't respect AddHostMetadataToTrace
If Stress Relief activates, it doesn't decorate the trace with meta.refinery.local_hostname.

It should; it's useful in diagnostics.

We should also consider if there are other meta fields that would also be valuable.

fchikwekwe:
Closes #641 

Add hostname to spans where stress relief mode is activated.

### DIFF
--- a/collect/collect.go
+++ b/collect/collect.go
@@ -559,8 +559,8 @@ func (i *InMemCollector) ProcessSpanImmediately(sp *types.Span, keep bool, sampl
 	if i.Config.GetAddRuleReasonToTrace() {
 		sp.Event.Data["meta.refinery.reason"] = reason
 	}
-	if i.Config.GetStressReliefConfig().Mode != "" {
-		sp.Event.Data["meta.refinery.local_hostname"] = i.hostname
+	if i.hostname != "" {
+		sp.Data["meta.refinery.local_hostname"] = i.hostname
 	}
 	i.addAdditionalAttributes(sp)
 	mergeTraceAndSpanSampleRates(sp, sampleRate, i.Config.GetIsDryRun())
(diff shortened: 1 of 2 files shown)


collect/collect.go
```diff
@@ -707,6 +712,9 @@ func (i *InMemCollector) send(trace *types.Trace, reason string) {
 		if i.Config.GetAddRuleReasonToTrace() {
 			sp.Data["meta.refinery.reason"] = reason
 		}
+		if i.Config.GetStressReliefConfig().Mode != "" {
+			sp.Event.Data["meta.refinery.local_hostname"] = i.hostname
```
TylerHelmuth, line 716: It looks like this function is already setting `local_hostname` whenever `i.hostname` is set on line 731.

collect/collect.go
```diff
@@ -559,7 +559,9 @@ func (i *InMemCollector) ProcessSpanImmediately(sp *types.Span, keep bool, sampl
 	if i.Config.GetAddRuleReasonToTrace() {
 		sp.Event.Data["meta.refinery.reason"] = reason
 	}
-
+	if i.Config.GetStressReliefConfig().Mode != "" {
```
TylerHelmuth, line 562: Can we use the same `if i.hostname !=nil` logic that is used in the `send` function?

collect/collect.go
```diff
@@ -571,6 +573,9 @@ func (i *InMemCollector) ProcessSpanImmediately(sp *types.Span, keep bool, sampl
 func (i *InMemCollector) dealWithSentTrace(keep bool, sampleRate uint, spanCount uint, sp *types.Span) {
 	if i.Config.GetAddRuleReasonToTrace() {
 		sp.Data["meta.refinery.reason"] = "late"
+
+	if i.Config.GetStressReliefConfig().Mode != "" {
```
TylerHelmuth, line 577: Can we use the same `if i.hostname !=nil` logic that is used in the `send` function?

collect/collect.go
fchikwekwe: Okay, changed all instances to adopt that pattern.

TylerHelmuth (approved):

Issue #671, opened 2023-04-27
Refinery doesn't seem to be checking APIKeys on OTLP traffic


kentquirk:
It works for http traffic but doesn't seem to for OTLP. 

A customer noticed this, and the effect is as if APIKeys is just `["*"]`. There are currently no test cases to validate the behavior for OTLP.

I have failing test cases on a branch right now.

kentquirk:
Fixed by #672

PR #672, opened 2023-04-27
fix: check apikeys for otlp requests too (+tests)

kentquirk:

- #671 -- Refinery's APIKeys configuration value was not being used for OTel requests

- Refactor middleware to pull out the key-validation logic
- Use it in the http middleware
- Call it in the OTLP handler
- Add a normal (mocked) test for OTLP
- Add an integration test for HTTP

### DIFF
--- a/route/middleware.go
+++ b/route/middleware.go
@@ -37,6 +37,22 @@ func (r *Router) queryTokenChecker(next http.Handler) http.Handler {
 	})
 }
 
+func (r *Router) isKeyAllowed(key string) bool {
+	allowedKeys, err := r.Config.GetAPIKeys()
+	if err != nil {
+		return false
+	}
+	for _, allowedKey := range allowedKeys {
+		if allowedKey == "*" {
+			return true
+		}
+		if allowedKey == key {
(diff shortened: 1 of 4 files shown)


route/middleware.go
```diff
@@ -37,6 +37,25 @@ func (r *Router) queryTokenChecker(next http.Handler) http.Handler {
 	})
 }
 
+func (r *Router) isKeyAllowed(key string) bool {
```
TylerHelmuth, line 40: Since this is called on every request, it could be nice to be able to check the key in a map instead of looping through a slice each time.

kentquirk: Doing that would require changing the data type of the config value to a map, and writing code that parses the current configuration into it both on initial load and on reload. It's a good idea, but it's something that a) we've already lived with for a long time, b) in practice might even be more expensive because most installations just have `["*"]` as the value, and c) and I'd rather do it as [a separate issue](https://github.com/honeycombio/refinery/issues/673) as part of the config rework.

TylerHelmuth (approved):

PR #677, opened 2023-05-09
feat: [v2] Config conversion and validation code from one data file

kentquirk:
This is the beginnings of a multi-pronged approach to data conversion.

We want a tool that can read an existing V1 config file and generate a new one, omitting no-longer-needed values and reformatting others, and generating a properly commented YAML file that is up to date with Refinery's capabilities.

We *also* want a validation library that can check a config completely. The best way to do that is with a JSON Schema validator, which in turn needs a JSON Schema data file that expresses the constraints on the fields.

Since both of these things need a lot of information about all the fields in the config, this PR creates one big data file that will contain enough information to solve both problems, and I'm planning to generate both the template for the converter *and* the JSON schema from the same data file.

In scope for this PR:

- [x] create config converter that uses a template to convert
- [x] create data file containing details of config items
- [x] create tool for generating the template from the data file
- [x] create rough structure of new config
- [x] manually check every field 
- [x] settle on the final design of config structure
- [x] adjust config structure to that design
- [x] finish config converter tool
- [x] document config converter tool

NOT in scope:

- adapt Refinery's config structure to this new design
- do anything with validation schema (this includes additional annotations to support schema -- these will come later)


kentquirk:
The config format and this tool are now in good enough shape to move on to other things; we will revisit before final release.

kentquirk:
There is now a single executable with subcommands(rather than running things as tests) with a Makefile to drive the non-converter commands. I've also done a `go:embed` so that the executable stands alone without need for external template or data files. And I've put summaries back and edited the descriptions so they're not redundant to the summaries.

kentquirk:
I just added docs generation for config; it writes out config.md.

tools/convert/Makefile
```diff
@@ -0,0 +1,35 @@
...
+	@echo
+	go run . sample --output=minimal_config.yaml
+
+.PHONY: doc
+#: generate the documentation
+doc:
+	@echo
+	@echo "+++ generating documentation"
+	@echo
+	go run . doc --output=config.md
```
TylerHelmuth, line 35:
```suggestion
	go run . doc --output=config.md

```

.gitignore
```diff
@@ -20,3 +20,5 @@ dockerize*
 
 # IDE configs
 .idea/
+
+.DS_Store
```
TylerHelmuth, line 24:
```suggestion
.DS_Store

```

tools/convert/README.md
```diff
@@ -0,0 +1,40 @@
...
+        file will list that key under the "LegacyMetrics" section under the "APIKey" name.
+
+        By default, it reads config.toml and writes to stdout. It will try to determine the
+        filetype of the input file based on the extension, but you can override that with
+        the --type flag.
+
+
+Application Options:
+  -i, --input=       the Refinery v1 config file to read (default: config.toml)
+  -o, --output=      the Refinery v2 config file to write (default: -)
```
TylerHelmuth, line 33: what does `default: -` mean in this context?

tools/convert/configData.yaml
```diff
@@ -0,0 +1,1204 @@
+# this contains a list of all of the config items supported by refinery
+# if v1name is missing, it's the same as name
+# if firstversion is missing, it's v1.x (meaning it's part of the legacy config)
+# if lastversion is present, it was valid in that version but not after
+
+# fields
+    # - v1name -- the name of this field in v1 (if empty, same as name)
+    # - v1group -- the group of this field in v1 (if empty, no group)
+    # - name -- the name of this field in v2 (required)
+    # - group -- the group of this field in v2 (if empty, no group)
```
TylerHelmuth, line 10: When we eventually do a v3 release, how with `v1name`, `v1group`, `name` and `group` interact?  Do we need `v2` prefixes here?

tools/convert/configV2.tmpl
```diff
@@ -0,0 +1,846 @@
+######################################
+## Honeycomb Refinery Configuration ##
+######################################
+#
+# created on {{ .Now }} from {{ .Input }}
```
TylerHelmuth, line 5: Smart

tools/convert/templates/configV2.tmpl
TylerHelmuth, line 1: Is this a duplicate of `tools/convert/configV2.tmpl`?

tools/convert/README.md
kentquirk: It's a convention meaning stdout (since defaulting to "stdout" would send it to a file of that name). I've used a lot of tools that do it that way...but maybe it's not conventional enough?

The other way to do this would be to set no default value and detect "" in the code instead of "-".

tools/convert/configData.yaml
kentquirk: It's more my expectation that v1name and v1group would go away after we're not parsing v1 anymore. The point is that name and group are what they say, and the v1* entries are how they used to be. 

In some v3 future, I'd probably add v2name to refer to the past, rather than adding the v2 prefixes to the actual name.

tools/convert/templates/configV2.tmpl
kentquirk: I moved it into templates but apparently failed to delete the parent one.

tools/convert/README.md
kentquirk: I changed the default to empty string and updated the comment and the doc.

tools/convert/templates/configV2.tmpl
kentquirk: Deleted.

TylerHelmuth (approved):

PR #683, opened 2023-05-17
docs: Use correct config section name for sample cache (do not merge)

isnotajoke:

The example configuration given in `config_complete.toml` didn't seem to turn on the cuckoo cache strategy when I uncommented the `Type` value and set it appropriately. After some digging, I found that [the code appears to look for `SampleCache` rather than `SampleCacheConfig`](https://github.com/honeycombio/refinery/blob/main/config/file_config.go#L709) (which is contained in the example file). When I changed my local config to use `SampleCache`, I saw my local installation pick up the config values and start using the new cache strategy. I assume this was a typo (it does match the name of implementation details in refinery itself).

* Change `SampleCacheConfig` to `SampleCache` in `config_complete.toml`
* Change `SampleCacheConfig` to `SampleCache` in the release notes


kentquirk:
@isnotajoke Thank you for this. 

And wow, we've had this problem for 6 months and no one else had noticed. I apologize.

As you might notice if you look at the last few pull requests, we're in the middle of a major rework of configuration. While your change is correct in the context of the current release, it's going to be obsoleted by the work that's been going on that will be released in a few weeks. But it has implications on that work and I *really* appreciate this report.

I've fixed this in the next release of refinery. We can't accept this PR as is, but I'll add you to the credits for the upcoming release! I'm going to leave this PR open until then with the WIP label so it doesn't get merged accidentally.

TylerHelmuth:
@isnotajoke thanks again for this fix, you've been credited in the Refinery 2.0 release notes!

Issue #688, opened 2023-05-18
If stress relief is activated by peers, it will never deactivate


kentquirk:
**Steps to reproduce**

1. Set the peer queue size too small; cause stress relief to trigger because the peer queue fills up. When stress relief activates, it stops sending traffic to peers, and thus the metric will never update and the stress indicator never resets.

That metric is controlled by libhoney. Options are to periodically send *something* to peers, or to change how that metric updates in libhoney.

PR #689, opened 2023-05-18
fix: send probes to peers during stress relief

kentquirk:

- #688 

When stress relief activates, it stops sending peer traffic. But the metrics that update stress relief based on peer volume only update when there's peer traffic. 

This causes any trace that refinery keeps (forwards to Honeycomb) in stress relief mode to *also* be sent as a probe to the appropriate peer (which then drops it).  This ensures that a little bit of traffic also gets to peers, which keeps the system operating normally and updating metrics.

- Make a copy of the incoming event as a probe whenever stress relief is active and would keep a span; if the probe would have been sent to a peer, send it. 
- When a probe is received from a peer, drop it without further examination.
- Log when this happens on both sides.
- Add the stress relief reason to the `meta.refinery.reason` field for kept spans so that it's not only found in the logs.

### DIFF
--- a/route/route.go
+++ b/route/route.go
@@ -497,7 +497,7 @@ func (r *Router) processEvent(ev *types.Event, reqID interface{}) error {
 		WithString("dataset", ev.Dataset).
 		WithString("environment", ev.Environment)
 
-	// check if this is a probe; if so, we should drop it
+	// check if this is a probe from another refinery; if so, we should drop it
 	if ev.Data["meta.refinery.probe"] != nil {
 		debugLog.Logf("dropping probe")
 		return nil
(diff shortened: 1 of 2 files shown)


route/route.go
```diff
@@ -524,19 +530,29 @@ func (r *Router) processEvent(ev *types.Event, reqID interface{}) error {
...
 		rate, keep, reason := r.Collector.GetStressedSampleRate(traceID)
 
 		r.Collector.ProcessSpanImmediately(span, keep, rate, reason)
-		return nil
+
+		if !keep {
+			return nil
+		}
+		// If the span was kept, we want to generate a probe that we'll forward
+		// to a peer IFF this span would have been forwarded.
```
akvanhar, line 543:
```suggestion
		// to a peer IF this span would have been forwarded.
```

kentquirk: Heh, that was an abbreviation my math professors would use to mean "If and Only If" -- but I'll take this as criticism that not everyone knows that and accept your suggestion. :) Thanks for the review!

akvanhar: I wondered if it was an abbreviation because it was in caps.

PR #700, opened 2023-06-04
feat: validation integration, part 1

kentquirk:

The validation system needed to be part of Refinery itself; I tried putting it in its own library, but it made more sense to make it an integral part of config.

- Moves the validation logic into Config
- Rename the validation data file and instances of it to metadata and into a validation object
- Use the object in the convert tool
- Embed the metadata file into the binary

This does not do rule validation yet -- that's next.

This also does not integrate the validation into Refinery's loading yet, although that is just a few lines of code. That will come after rule validation.


tools/convert/Makefile
```diff
@@ -33,3 +33,11 @@ doc:
...
 	@echo
 	go run . doc --output=config.md
+
+.PHONY: validate
+#: validate the sample config
+validate:
+	@echo
+	@echo "+++ validating sample config"
+	@echo
+	go run . validate --input=minimal_config.yaml
```
TylerHelmuth, line 43:
```suggestion
	go run . validate --input=minimal_config.yaml

```

TylerHelmuth (approved):
LGTM

PR #701, opened 2023-06-05
feat: Validation part deux -- add rules metadata and rules validation

kentquirk:
Note: this PR is dependent on #700 -- that one should be merged first.

- Adds complete rules validation to the validation system, plus a few bonus features.

- Add a rulesMeta.yaml file to describe the rules data.
- Extend the validation system to handle a structure nested more deeply than the basic config validation, which was only 2 layers deep. This meant adding objects and arrays of objects.
- Add a few more validation features to support the new semantics.
- Add a "did you mean...?" feature to suggest the proper spelling if someone gets something wrong. This is particularly useful when you get the case wrong.
- Generate a .md file from the rules metadata; it's not yet pretty but it's a good start. Will improve before release.

It's been validated against the converted rules_complete rules file. 

I'm really happy with this -- it makes it almost impossible to make a major error in your rules.


sample/dynamic.go
```diff
@@ -39,6 +39,7 @@ func (d *DynamicSampler) Start() error {
 	d.dynsampler = &dynsampler.AvgSampleRate{
 		GoalSampleRate:         int(d.sampleRate),
 		ClearFrequencyDuration: time.Duration(d.clearFrequency),
+		MaxKeys:                d.Config.MaxKeys,
```
TylerHelmuth, line 42: Is this a separate bug fix?

kentquirk: Kinda? I realized while doing this that some samplers didn't allow this field to be configured, and I wanted all the documentation to be consistent, so I did it along the way.

TylerHelmuth (approved):

PR #709, opened 2023-06-09
feat: Add WindowedThroughput sampler

kentquirk:
Depends on #708 

- Adds the other new sampler from the latest dynsampler-go.

- Add the sampler and its config info
- Update the docs
- Add a basic test (dynsampler has all the sampler testing)

### DIFF
--- a/config/metadata/rulesMeta.yaml
+++ b/config/metadata/rulesMeta.yaml
@@ -402,6 +402,111 @@ groups:
           is a useful indicator of traces you'd like to see in Honeycomb, set
           this to true.
 
+  - name: WindowedThroughputSampler
+    title: Windowed Throughput Sampler
+    description: >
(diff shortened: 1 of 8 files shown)


config/metadata/rulesMeta.yaml
```diff
@@ -402,6 +402,111 @@ groups:
           is a useful indicator of traces you'd like to see in Honeycomb, set
           this to true.
 
+  - name: WindowedThroughputSampler
```
TylerHelmuth, line 405: Do we even need the ThroughputSampler anymore?

TylerHelmuth (approved):

config/metadata/rulesMeta.yaml
kentquirk: Yes, mainly for compatibility. I do hope to deprecate it but don't think we need to do that for 2.0.

PR #711, opened 2023-06-10
docs: Improve generated documentation

kentquirk:

- This improves the documentation templates so the generated documentation is better-formatted, easier to read, and generally more usable. It also moves the generated docs into the root folder.
- It's not really as big as it might look -- the template files got revised and then the files were regenerated and moved. 
- A few tweaks were made to code and the metadata to eliminate redundancy.

- Revise templates to explain more upfront
- Add example to each template
- Add table of contents so the file is more referenceable
- Add $ reference capability to rules data to eliminate redundant descriptions in the rules metadata file


TylerHelmuth (commented):
Can we use this opportunity to throw in a `generated on` timestamp to all the generated files?

tools/convert/helpers.go
```diff
@@ -197,6 +198,11 @@ func nonZero(data map[string]any, key, oldkey string, example string) string {
 	return fmt.Sprintf(`# %s: %v`, key, yamlf(example))
 }
 
+func now() string {
+	t := time.Now()
+	return fmt.Sprintf("on %s at %s", t.Format("2006-01-02"), t.Format("15:04:05 MST"))
```
TylerHelmuth, line 203: Lets use UTC like https://github.com/open-telemetry/opentelemetry-operator/blob/846046ccccf3d2bde63be9cb7f75537af5332638/bundle/manifests/opentelemetry-operator.clusterserviceversion.yaml#L34

kentquirk: OK...I didn't want to use the RFC3339 format because these are supposed to be human-readable. But using UTC is the right thing.

TylerHelmuth (approved):

PR #715, opened 2023-06-13
perf: Create fibcache alternative cache; not used yet

### LINKED ISSUE #545: Cache ejects traces too soon
**Problem**

The trace cache keeps a circular buffer to preserve insertion order, and ejects traces when it overwrites a cache entry in the circular buffer. 

However, when a trace is sent because its root span has arrived, the spot in the circular buffer is just replaced with nil. This means that even though there may be plenty of space (nil slots) in the cache, a trace is ejected automatically whenever the buffer has wrapped around

Example: the buffer has 100 slots. The first trace arrives, and takes a long time to complete. Then 99 fast traces arrive and are completed and sent. The 101st trace to arrive will cause the first to be ejected, even though there are 99 empty slots in the buffer.

This isn't necessarily incorrect behavior, but in systems under heavy load where some traces may take a long time, those traces may get ejected sooner than would otherwise be necessary.

This is, I think, reasonable to characterize as a bug, although it's fairly benign so I'm *also* calling it an enhancement.

A more efficient insertion order algorithm (perhaps a priority queue) so that traces can stay in the cache as long as configuration permits without being artificially limited.

Not changing it.

**Additional context**

Discovered when attempting to create metrics for cache behavior.

kentquirk:

- The existing legacy in-memory cache uses a hand-coded circular buffer to manage cache entries, and will overwrite long-lived traces because it just marches through the buffer in receive order -- even if the buffer wasn't full. This PR replaces that with a Fibonacci Heap, which is a priority heap structure that offers O(1) insertion and O(log n) deletion without keeping old items in the heap. Therefore, we should see many fewer cache overruns.

The code is also quite small, and with a bit of rework to the cache interface, this library would make resizing the cache a lot cheaper.

This implementation preserves the cache size limit.

This is still a draft; needs some benchmarking and a bit more testing.

- Add the FibHeap object
- Implement the Key and Tag functions on Trace so they can be stored in the heap easily
- Add some tests

Closes #545.

### DIFF
--- a/collect/cache/fibcache.go
+++ b/collect/cache/fibcache.go
@@ -1,0 +1,134 @@
+package cache
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/honeycombio/refinery/logger"
+	"github.com/honeycombio/refinery/metrics"
+	"github.com/honeycombio/refinery/types"
+
+	fibheap "github.com/starwander/GoFibonacciHeap"
+)
+
+// // Cache is a non-threadsafe cache. It must not be used for concurrent access.
+// type Cache interface {
+// 	// Set adds the trace to the cache. If it is kicking out a trace from the cache
+// 	// that has not yet been sent, it will return that trace. Otherwise returns nil.
+// 	Set(trace *types.Trace) *types.Trace
+// 	Get(traceID string) *types.Trace
+// 	// GetAll is used during shutdown to get all in-flight traces to flush them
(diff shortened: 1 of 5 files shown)


kentquirk:
Going to leave this here for a little longer but will probably abandon it. The discussion has evolved this a bit -- we may be able to get most of the benefit by just checking more slots before deciding we have an overflow.

PR #717, opened 2023-06-16
fix: Update dynamic samplers to use GetSampleRateMulti

### LINKED ISSUE #636: Refinery's throughput calculations are based on traces, not spans
The dynsampler library that refinery uses assumes it's counting spans -- but refinery uses it to count traces. Thus, throughput calculations that are attempting to limit based on volumes sent to Honeycomb are wrong by a large factor. 

Dynsampler is getting the ability to have the information to do the calculation correctly; this is a tracking issue to make sure that refinery uses it.

I think for the case of throughput calculations, we don't need to worry about #555 -- but I'm noting it here to make sure we consider it.

kentquirk:

- As it says in #636, throughput and dynamic sampler calculations have been based on the wrong numbers for some time. They've been counting traces, not spans, and yet everything about Honeycomb is based on span count. Most people think it works by counting spans and are then surprised when sample rates seem incorrect for their volume, and in fact, our documentation says that it works this way.

- This changes Refinery's dynamic and throughput sampler requests to use the span count, not the trace count, when calculating numbers for the dynamic samplers and throughput samplers. 

This will need to be reflected in updated documentation.

Closes #636.

### DIFF
--- a/sample/dynamic.go
+++ b/sample/dynamic.go
@@ -65,7 +65,8 @@ func (d *DynamicSampler) Start() error {
 
 func (d *DynamicSampler) GetSampleRate(trace *types.Trace) (rate uint, keep bool, reason string, key string) {
 	key = d.key.build(trace)
-	rate = uint(d.dynsampler.GetSampleRate(key))
+	count := int(trace.DescendantCount())
+	rate = uint(d.dynsampler.GetSampleRateMulti(key, count))
 	if rate < 1 { // protect against dynsampler being broken even though it shouldn't be
 		rate = 1
 	}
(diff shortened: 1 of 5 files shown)


TylerHelmuth (approved):
This is an exciting change to help people use the windowedthroughput sampler to enforce their rate limit, but it is a significant breaking change right?  We definitely want to highlight this change in the release. 

Is there anything we can do in the config converter to detect users that are affected by this change and warn them to look at their rules configs closely?

kentquirk:
I believe that I can make the config converter notice that and emit a warning. It's on the list.

PR #719, opened 2023-06-17
feat: allow suffixes on memory size in config

kentquirk:
Based on #718, so ready to review independently.

- Up to now, setting MaxAlloc requires typing or copy-pasting large numbers that are hard to read and visually parse. This PR allows using standard suffixes like "Gb" and "MiB". This responds to a request from Honeycomb field engineering.

- Import the docker library for parsing memory sizes
- Create a MemorySize type with marshal and unmarshal methods, and use it as appropriate
- Update the metadata and templates
- Regenerate all the things

### DIFF
--- a/config/file_config.go
+++ b/config/file_config.go
@@ -9,6 +9,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/docker/go-units"
 	"gopkg.in/yaml.v3"
 )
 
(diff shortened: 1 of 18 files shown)


config/file_config.go
```diff
@@ -9,6 +9,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/docker/go-units"
```
TylerHelmuth, line 12: I'm curious, are this units the same you'd use for k8s?

kentquirk: I believe so -- it supports KGMT and a few more, with or without i and B, as well as plain numbers. So "GiB", "Tb", etc.

tools/convert/templates/configV2.tmpl
```diff
@@ -749,11 +749,12 @@ Collection:
     ## MaxAlloc is the maximum number of bytes that should be allocated by
     ## the collector.
     ##
-    ## If set, it must be an integer >= 0. 64-bit values are supported. See
-    ## MaxMemory for more details.
+    ## If set, this must be a memory size. 64-bit values are supported. Sizes
+    ## with standard unit suffixes like "MB" and "GiB" are also supported.
+    ## See MaxMemory for more details.
```
TylerHelmuth, line 754: I think it would be good to either list out to link to the complete list of valid suffixes

kentquirk: Done and regenerated.

TylerHelmuth (approved):

PR #721, opened 2023-06-17
feat: MaxAlloc improvements

kentquirk:
Dependent on #719; can be reviewed before merging that.

In v1, Refinery allowed setting a `MaxAlloc` value that controlled the maximum amount of memory it would try to use. This is hard to keep configured properly, because the guidance is to set it to some fraction of the memory allocated to the pod. This PR implements two new values: 

- the full allocation amount, called `AvailableMemory`
- the percentage refinery should use, called `MaxMemory`

The existing `MaxAlloc` value is still supported. If the first two are nonzero, they are multiplied to get the maximum allocation; otherwise, `MaxAlloc` is used. As in v1, if all of them are 0, refinery doesn't attempt to limit its own use.

I spent a while trying to find a way to allow the app to read the max allocation from the system, but it's pretty platform-dependent. It seemed better to just make it so that the value can be put into the environment in the Helm chart.

- Add `AvailableMemory` and make it settable in the environment and on the command line
- Implement the logic for calculating the value refinery should use and use that in the allocation check
- Fix the names of the appropriate function calls and mocks
- Regenerate files


config/file_config.go
```diff
@@ -184,9 +184,17 @@ type RedisPeerManagementConfig struct {
 
 type CollectionConfig struct {
 	// CacheCapacity must be less than math.MaxInt32
-	CacheCapacity int        `yaml:"CacheCapacity" default:"10_000"`
-	MaxMemory     int        `yaml:"MaxMemory" default:"75"`
-	MaxAlloc      MemorySize `yaml:"MaxAlloc"`
+	CacheCapacity   int        `yaml:"CacheCapacity" default:"10_000"`
+	AvailableMemory MemorySize `yaml:"AvailableMemory" cmdenv:"AvailableMemory"`
+	MaxMemory       int        `yaml:"MaxMemory" default:"75"`
```
TylerHelmuth, line 189: My first thought reading this was that `MaxMemory` could also be a `MemorySize`, assuming that it represented the max amount of memory Refinery could use.  I didn't realize this is the max percentage of available memory Refinery can use.  Is there a way we can throw "percent"/"percentage" in the name somewhere?

kentquirk: Fixed

TylerHelmuth (approved):

PR #723, opened 2023-06-18
fix: Improve balance of the sharding algorithm

kentquirk:

- The deterministic sharding algorithm uses N randomly generated partitions for each shard. Since they're random, they can be unbalanced. For N > 1, we can move some partitions around to rebalance them. This does that.

- Add rebalancing code. 
- Clean up flaky tests.
- Add more comments.

### DIFF
--- a/sharder/deterministic.go
+++ b/sharder/deterministic.go
@@ -2,6 +2,7 @@ package sharder
 
 import (
 	"fmt"
+	"math"
 	"net"
 	"net/url"
 	"sort"
@@ -21,6 +22,7 @@ import (
 const (
 	shardingSalt        = "gf4LqTwcJ6PEj2vO"
 	peerSeed     uint64 = 6789531204236
+	shardSeed    uint64 = 2938476120934
 )
 
 // DetShard implements Shard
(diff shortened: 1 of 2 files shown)


sharder/deterministic.go
```diff
@@ -291,20 +294,101 @@ func (d *DeterministicSharder) loadPeerList() error {
 	} else {
 		d.peerLock.RUnlock()
 	}
+	// the random partitioning may have significant imbalances, so we try to
+	// balance them by moving partitions from the most heavily loaded nodes to
+	// the least heavily loaded nodes. But it's only possible if the number of
+	// partitions per peer is > 1.
+	if partitionsPerPeer > 1 {
```
TylerHelmuth, line 301: Does this also only need to happen if the number of refinery nodes is greater than 1?  If only a single instance of Refinery is running then balancing is unnecessary right?

kentquirk: Good catch, yes.

kentquirk:
After reviewing this with @cewkrupa and @TylerHelmuth I've come to the realization that this is not quite correct. In particular, the getMinMax function is incorrect -- the proper approach is to probe the hash space with random data to find the affinity of each partition. This is a more complex problem than we can solve this week, so I'm closing this for now and will revisit later.

PR #724, opened 2023-06-18
fix: make refinery run from minimal config

kentquirk:
Based on #723 so that should merge first, but this is independently reviewable.

- All the config changes had some minor issues that kept refinery from starting up with minimal configuration. This fixes that.

- logrus -> stdout
- fix sequencing of metrics startup
- fix default peer for file-based peer list
- add UnmarshalText method to log level


sharder/deterministic.go
```diff
@@ -291,20 +294,101 @@ func (d *DeterministicSharder) loadPeerList() error {
...
+	for i := range totals {
+		if totals[i] < totals[minix] {
+			minix = i
+		}
+		if totals[i] > totals[maxix] {
+			maxix = i
+		}
+	}
+	return minix, maxix, totals[maxix] / totals[minix]
+}
```
TylerHelmuth, line 328: We need to remove this content from the PR

config.yaml
```diff
@@ -1,1055 +1,5 @@
...
-    ##
-    ## Setss the minimum time that stress relief will stay enabled, once
-    ## activated. This helps to prevent oscillations.
-    ##
-    ## Accepts a duration string with units, like "10s".
-    ## default: 10s
-    ## Eligible for live reload.
-    # MinimumActivationDuration: "10s"
+  ConfigurationVersion: 2
+  MinRefineryVersion: v2.0
```
TylerHelmuth, line 3: Are these deletions expected?

TylerHelmuth (commented):
@kentquirk looks like there might be a bad rebase in here

sharder/deterministic.go
kentquirk: Done.

config.yaml
kentquirk: Yes; some changes crept in that shouldn't have, but this is the correct regeneration.

Issue #726, opened 2023-06-21
Add ability to turn off config polling


TylerHelmuth:
**Is your feature request related to a problem? Please describe.**
At the moment Refinery has the ability to configure how often the config file is polled for changes, but it does not let you turn off this feature.

**Describe the solution you'd like**
Add the ability to turn off config polling by passing in a value of 0 in the configuration section.

PR #730, opened 2023-06-21
feat: Allow disabling reload monitoring

TylerHelmuth:

- Closes https://github.com/honeycombio/refinery/issues/726

- Adds the ability to disable config polling by setting `ConfigReloadInterval` to `0s`.
- Adds new test that confirms the feature is disabled.
- Generates new configs with updated description

### DIFF
--- a/config/file_config.go
+++ b/config/file_config.go
@@ -411,9 +411,13 @@ func (f *fileConfig) monitor() {
 
 // Stop halts the monitor goroutine
 func (f *fileConfig) Stop() {
-	f.ticker.Stop()
-	close(f.done)
-	f.done = nil
+	if f.ticker != nil {
+		f.ticker.Stop()
+	}
+	if f.done != nil {
+		close(f.done)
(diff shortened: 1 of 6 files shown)


config/file_config.go
```diff
@@ -368,7 +368,9 @@ func NewConfig(opts *CmdEnv, errorCallback func(error)) (Config, error) {
 	cfg.callbacks = make([]func(), 0)
 	cfg.errorCallback = errorCallback
 
-	go cfg.monitor()
+	if cfg.mainConfig.General.ConfigReloadInterval > 0 {
```
kentquirk, line 371: Since we're not calling monitor any more, that means that f.ticker is nil, which means that if Stop() gets called it will panic. Please add a nil check to Stop().

TylerHelmuth: Nice catch

kentquirk (approved):
Thanks. Good test, too.

PR #731, opened 2023-06-21
docs: Refinery Rules copyediting

mjingle:
This PR hopes to make the generated `rules.yaml` file and associated documentation the best that it can be.

- [x] Add `codeblock` formatting for fields
- [x] Clarify existing wording
- [x] Check and standardize "weigh" vs. "weight" use
- [x] Check/Standardize use of Sampler Names in description text

### DIFF
--- a/config/metadata/rulesMeta.yaml
+++ b/config/metadata/rulesMeta.yaml
@@ -4,22 +4,23 @@ groups:
     sortorder: 0
     summary: is the list of samplers to use.
     description: >
-      Samplers is a mapping of targets to samplers. Each target is a Honeycomb
-      environment (or, for classic keys, a dataset). The value is the sampler to
-      use for that target. The target called `__default__` will be used for any
-      target that is not explicitly listed. A `__default__` target is required.
+      `Samplers` maps targets to sampler configurations. Each target is a 
+      Honeycomb environment, (or for Honeycomb Classic keys, a dataset).
+      The value is the sampler to use for that target. A `__default__` 
+      target is required. The target called `__default__` will be used 
+      for any target that is not explicitly listed.
 
       The targets are determined by examining the API key used to send the
-      trace. If the API key is a 'classic' key (which is a 32-character
-      hexadecimal value), the specified dataset name is used as the target. If
-      the API key is a new-style key (20-23 alphanumeric characters), the
-      key's environment name is used as the target.
+      trace. If the API key is a Honeycomb Classic key with a 32-character
+      hexadecimal value, then the specified dataset name is used as the target. 
+      If the API key is a key with 20-23 alphanumeric characters, then 
+      the key's environment name is used as the target.
 
   - name: DeterministicSampler
     title: Deterministic Sampler
     sortorder: 10
     description: >
-      The deterministic sampler uses a fixed sample rate to sample traces
+      The Deterministic Sampler uses a fixed sample rate to sample traces
       based on their trace ID. This is the simplest sampling algorithm
       - it is a static sample rate, choosing traces randomly to either keep or
       send (at the appropriate rate). It is not influenced by the contents of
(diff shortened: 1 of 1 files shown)


config/metadata/rulesMeta.yaml
```diff
@@ -495,24 +506,24 @@ groups:
...
         description: >
-          Indicates whether to drop the trace if it matches this rule. If true,
-          the trace will be dropped. If false, the trace will be kept.
+          Indicates whether to drop the trace if it matches this rule. If `true`,
+          the trace will be dropped. If `false`, the trace will be kept.
       - name: SampleRate
         type: int
         summary: is the sample rate to use.
         description: >
           If the rule is matched, there is no Sampler specified, and the Drop
```
mjingle, line 525: @kentquirk For this description, do the previous three conditions (rule is matched, no specified Sampler, and `Drop` flag is `false`) have to be met before `SampleRate` here is used?

mjingle, line 525: NTS: fix formatting for `Drop` in next commit.

config/metadata/rulesMeta.yaml
```diff
@@ -495,24 +506,24 @@ groups:
               - EMAThroughputSampler
               - WindowedThroughputSampler
               - TotalThroughputSampler
-        summary: is the dynamic sampler to use if the rule matches.
+        summary: is the Dynamic Sampler to use if the rule matches.
         description: >
           The sampler to use if the rule matches. If this is set, the sample
           rate will be determined by this downstream sampler. If this is not set,
-          the sample rate will be determined by the Drop flag or the SampleRate
+          the sample rate will be determined by the `Drop` flag or the `SampleRate`
```
mjingle, line 513: NTS: Add period to end of sentence in next commit.

config/metadata/rulesMeta.yaml
```diff
@@ -181,35 +192,35 @@ groups:
...
-          value should generally be <= Weight, unless you have very specific
-          reasons to set it higher.
+          value should generally be less than (<=) `Weight`, unless you 
+          have very specific reasons to set it higher.
       - name: BurstMultiple
         type: float
         summary: is the multiple of the sample rate to use when detecting bursts.
         description: >
           If set, this value is multiplied by the sum of the running average
           of counts to define the burst detection threshold. If total counts
```
mjingle, line 207: @kentquirk Is the "burst detection threshold" another named field within this documentation?

config/metadata/rulesMeta.yaml
kentquirk: That's the point of the note above -- one of 3 things happens, in the specified order.

mjingle: So all three conditions have to occur to use this Sample Rate. Okay, thank you for clarifying! I was having difficulty parsing the sentence and wanted to be sure.

config/metadata/rulesMeta.yaml
```diff
@@ -495,31 +506,31 @@ groups:
...
+          If the rule is matched, there is no Sampler specified, and the `Drop`
+          flag is `false`, then this is the sample rate to use.
       - name: Conditions
         type: objectarray
         summary: is the list of conditions to use to determine whether the rule matches.
         description: >
           Conditions is a list of conditions to use to determine whether the
           rule matches. All conditions must be met for the rule to match. If
-          there are no conditions, the rule will always match (this is typically
+          there are no conditions, then the rule will always match (this is typically
```
mjingle, line 533: @kentquirk Asking for clarity - What is "this" supposed to be in the sentence? The rule or the condition?

config/metadata/rulesMeta.yaml
kentquirk: No, this value defines the threshold used to start burst detection. Basically, if traffic suddenly increases by a factor of X, we want to recalculate the EMA immediately rather than waiting. This number is X.

config/metadata/rulesMeta.yaml
kentquirk: Maybe remove parens and say "A no-condition rule is typically used for the last rule to provide a default behavior."

mjingle: Sounds good! Adding now.

config/metadata/rulesMeta.yaml
mjingle: Understood. I asked because my docs senses were on alert given the phrasing around burst detection threshold.

kentquirk (approved):
This is awesome. Thank you. I'll regenerate the other files after this merges.

TylerHelmuth (approved):

PR #736, opened 2023-06-22
feat: add ability to convert a helm chart

kentquirk:

- This gives the convert tool the ability to read config and rules out of a helm chart, convert them, and write them back to a helm chart. All comments are stripped. If that's not cool for you, you'll have to do it manually.

- Add helm command
- Load the helm chart as a map with YAML
- Extract config and run it through the config converter, then read that back into a map
- Extract rules and convert them
- Put both back into the map
- Write out the YAML again

### DIFF
--- a/tools/convert/main.go
+++ b/tools/convert/main.go
@@ -272,6 +272,8 @@ func ConvertConfig(tmplData *configTemplateData, w io.Writer) {
 }
 
 func ConvertHelm(tmplData *configTemplateData, w io.Writer) {
+	const rulesConfigMapName = "RulesConfigMapName"
+	const liveReload = "LiveReload"
 	// convert config if we have it
 	helmConfigAny, ok := tmplData.Data["config"]
 	if ok {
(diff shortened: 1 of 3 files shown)


tools/convert/main.go
```diff
@@ -256,6 +255,79 @@ func loadRulesMetadata() *config.Metadata {
...
+		helmConfig, ok := helmConfigAny.(map[string]any)
+		if !ok {
+			panic("config in helm chart is the wrong format!")
+		}
+		convertedConfig := &bytes.Buffer{}
+		// make a copy of this tmplData and overwrite the Data part
+		config := *tmplData
+		config.Data = helmConfig
+		// convert the config into the buffer
+		ConvertConfig(&config, convertedConfig)
```
TylerHelmuth, line 287: The values.yaml `config` section may contain a field named `RulesConfigMapName` that is now incompatible with Refinery's stance against extraneous configuration fields.  We'll need to remove the entry and move it to the root of the values.yaml.

tools/convert/main.go
```diff
@@ -256,6 +255,79 @@ func loadRulesMetadata() *config.Metadata {
...
+	}
+
+	// now try the rules
+	helmRulesAny, ok := tmplData.Data["rules"]
+	if ok {
+		helmRules, ok := helmRulesAny.(map[string]any)
+		if !ok {
+			panic("config in helm chart is the wrong format!")
+		}
+		rules := convertRulesToNewConfig(helmRules)
```
TylerHelmuth, line 318: The values.yaml `rules` section may contain a field named `LiveReload ` that is now incompatible with Refinery's stance against extraneous rule fields.  We'll need to remove the entry and move it to the root of the values.yaml.

tools/convert/main.go
kentquirk: Whoops, I completely forgot about that one! Will fix.

TylerHelmuth (approved):

PR #737, opened 2023-06-22
docs: Fix up docs, especially envvar and cmdline

### LINKED ISSUE #727: docs: generate config and rules docs for use on Honeycomb site

Follow the specs given by honeycomb docs team

kentquirk:

- Cleans up the documentation and code around environment variables and command line.

- Fix and add info about envvars and cmdlines in metadata
- Add the info to templates so they show up in the docs
- Regenerate all the things

Closes #727

### DIFF
--- a/config/metadata/configMeta.yaml
+++ b/config/metadata/configMeta.yaml
@@ -80,8 +80,8 @@ groups:
         reload: false
         validations:
           - type: notempty
-        envvar: REFINERY_HTTP_LISTEN_ADDR
-        commandLine: http-listen-addr
+        envvar: REFINERY_HTTP_LISTEN_ADDRESS
+        commandLine: http-listen-address
         summary: is the address where Refinery listens for incoming requests.
         description: >
           This setting is the IP and port on which Refinery listens for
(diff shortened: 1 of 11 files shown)


config/metadata/configMeta.yaml
```diff
@@ -542,7 +539,8 @@ groups:
         reload: false
         validations:
           - type: requiredInGroup
-        envvar: REFINERY_METRICS_API_KEY
+        envvar: REFINERY_LEGACY_METRICS_API_KEY, HONEYCOMB_API_KEY
```
TylerHelmuth, line 542: Do we need `LEGACY` in the name?  Also, that is a breaking change we want to make sure we document because it can't be handled by the converter tool, users would need to update their pipelines

TylerHelmuth (commented):
If we want to reduce the breaking changes we might want to keep the original env var names and add support for the new ones.  Changing env var names means users have to go update their pipelines.

kentquirk:
@TylerHelmuth I've compared with documented env vars and reverted all the ones that were documented in the README to what they were in 1.21.

cartermp (approved):

PR #740, opened 2023-06-23
fix: metrics injection startup issues

kentquirk:

- Satisfying the injection system requires special magic which is neither special nor satisfying. But now, at least, refinery starts up reliably.

- Make metrics objects behave the way that injection insists they must (must implement start, must not cause gaps in the injection graph)
- Fix tests

### DIFF
--- a/metrics/multi_metrics.go
+++ b/metrics/multi_metrics.go
@@ -1,6 +1,8 @@
 package metrics
 
-import "sync"
+import (
+	"sync"
+)
 
 // MultiMetrics is a metrics provider that sends metrics to at least one
 // underlying metrics provider (StoreMetrics). It can be configured to send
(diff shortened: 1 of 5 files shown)


metrics/multi_metrics.go
```diff
@@ -10,30 +12,22 @@ import "sync"
 // which can then be retrieved with Get(). This is for use with StressRelief. It
 // does not track histograms or counters, which are reset after each scrape.
 type MultiMetrics struct {
-	children []Metrics
-	values   map[string]float64
-	lock     sync.RWMutex
+	MList  *MetricsList `inject:"metricslist"`
```
TylerHelmuth, line 15: @kentquirk thinking through this problem some more I think that the root issue was that we weren't implementing the `Start` interface correctly.  I think that we can actually keep the original `GetMetricsImplementations` and `MultiMetrics` implementations as long as we implement the interface correct.

TylerHelmuth: I just played with this idea and confirmed it works.  We can safely remove all the `MetricsList` stuff.

kentquirk:
Also too much.

PR #747, opened 2023-06-26
feat: enable pyroscope deltaprof

lizthegrey:

- On long-lived services like Refinery with large process uptimes, the number of distinct stacks and code addresses grows in the profile as time goes on; this causes high numbers of bytes to be scraped when collecting profiles for timeseries that aren't increasing in value very often but have been seen at least once in process lifetime.

- Opts into https://github.com/grafana/godeltaprof for delta profiles which do not have the weakness above.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -23,6 +23,7 @@ require (
 	github.com/pelletier/go-toml/v2 v2.0.8
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.15.1
+	github.com/pyroscope-io/godeltaprof v0.1.1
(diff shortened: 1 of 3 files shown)


TylerHelmuth (approved):
I don't know anything about profiling but all the names line up.  With 2.0 coming out so soon I'd like @kentquirk review and also opinion on whether this should be included in 2.0 or 2.1.

lizthegrey:
> I don't know anything about profiling but all the names line up. With 2.0 coming out so soon I'd like @kentquirk review and also opinion on whether this should be included in 2.0 or 2.1.

It's the same code as is in hound if there are any concerns :)

PR #754, opened 2023-06-29
perf: preallocate outbound libhoney attribute map

### LINKED ISSUE #749: Refinery spends unnecessary 5% cpu resizing maps
`EnqueueEvent` winds up looping over `AddField()` for a total of `len(ev.Data)` times.

Unfortunately due to behaviour in Go runtime (see https://github.com/golang/go/issues/56182) we can't resize the map once we have it (nor do we have direct access to the map from libhoney clients since it's a private and lock-protected struct field).

Thus, we need to be able to pass in `len(ev.Data)` to `builder.NewEvent()` to allow correctly pre-sizing the map where the number of fields is known, and avoiding reallocating and copying the map over and over again for 5% of time spent in all of refinery.

lizthegrey:
Fixes #749 by using https://github.com/honeycombio/libhoney-go/pull/228 new method

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -15,10 +15,10 @@ require (
 	github.com/hashicorp/golang-lru v0.5.4
 	github.com/honeycombio/dynsampler-go v0.5.1
 	github.com/honeycombio/husky v0.22.4
-	github.com/honeycombio/libhoney-go v1.18.0
+	github.com/honeycombio/libhoney-go v1.19.1-0.20230629150248-9227eb685909
 	github.com/jessevdk/go-flags v1.5.0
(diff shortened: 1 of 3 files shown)


kentquirk:
@lizthegrey thanks. This is great stuff, and we will approve it, but we're going to release 2.0 first and we can merge this right after.

TylerHelmuth:
Since we got a chance to run this internally for the last several days (and the 2.0 release was delayed) we now feel confident in its inclusion in 2.0

TylerHelmuth (approved):

PR #757, opened 2023-06-30
fix: Improve drop cache performance

kentquirk:

- The trace drop cache was causing cascading failures under high load when stress relief triggered. The problem was traced to the fact that the cuckoofilter library was internally locking the system random number generator repeatedly and the time spent locking and unlocking was significant. 

A test was done with a private fork of the library to remove this lock by using a different way to generate random values. This cut the time spent in the cuckoofilter by half, but this probably wouldn't have been enough to matter.

Instead, this change decouples the Add call for the drop cache; Add now simply puts an item into a large (10K item) channel that is monitored by a separate goroutine, which can in turn batch up to 5 calls to the cuckoo cache in one lock. This results in a serious performance improvement:

```
BenchmarkCuckooTraceChecker_Add-10                 	     100	      7057 ns/op
BenchmarkCuckooTraceChecker_AddParallel-10         	     100	      7385 ns/op
BenchmarkCuckooTraceChecker_Check-10               	     100	        15.83 ns/op
BenchmarkCuckooTraceChecker_CheckParallel-10       	     100	       172.9 ns/op
BenchmarkCuckooTraceChecker_CheckAddParallel-10    	     100	      7548 ns/op
```

-

### DIFF
--- a/collect/cache/cuckoo.go
+++ b/collect/cache/cuckoo.go
@@ -2,6 +2,7 @@ package cache
 
 import (
 	"sync"
+	"time"
 
 	"github.com/honeycombio/refinery/metrics"
 	cuckoo "github.com/panmari/cuckoofilter"
@@ -12,6 +13,7 @@ const (
 	CurrentLoadFactor = "cuckoo_current_load_factor"
 	FutureLoadFactor  = "cuckoo_future_load_factor"
 	CurrentCapacity   = "cuckoo_current_capacity"
+	AddQueueFull      = "cuckoo_add_queue_full"
 )
 
 // This wraps a cuckoo filter implementation in a way that lets us keep it running forever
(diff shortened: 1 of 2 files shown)


collect/cache/cuckoo.go
```diff
@@ -24,32 +26,78 @@ const (
...
 		met:      m,
+		addch:    make(chan string, AddQueueDepth),
 	}
+
+	// To avoid blocking on Add, we have a goroutine that pulls from a channel and adds to the filter.
+	traces := make([]string, AddQueueBatchSize)
+	go func() {
+		for {
+			ready := false
+			i := 0
```
TylerHelmuth, line 62: I think `i` needs instantiated outside the `for` loop and then reset inside `if ready && i > 0 {`

TylerHelmuth (approved):

TylerHelmuth:
We tested this change over the last 24 hours and feel confident that it fixes the problem.  Will merge in for 2.0.

Issue #766, opened 2023-07-06
Convert does not correctly handle converting `EMADynamicSamplerConfig.AdjustmentInterval`


TylerHelmuth:
**Versions**

- Refinery: 2.0.0

**Steps to reproduce**

input:

```
rules:
  Sampler: DeterministicSampler
  SampleRate: 1
  test: # E&S kibble rules (hence lowercase)
    Sampler: RulesBasedSampler
    rule:
      - name: test
        condition:
          - field: service.name
            operator: =
            value: test
        sampler:
          EMADynamicSampler:
            Sampler: EMADynamicSampler
            AdjustmentInterval: 60
      - name: default rule
        SampleRate: 1
```

expected output:

```
rules:
    RulesVersion: 2
    Samplers:
        __default__:
            DeterministicSampler:
                SampleRate: 1
        test:
            RulesBasedSampler:
                Rules:
                    - Name: test
                      Conditions:
                        - Field: service.name
                          Operator: =
                          Value: test
                      Sampler:
                        EMADynamicSampler:
                            AdjustmentInterval: 1m0s
                    - Name: default rule
                      SampleRate: 1
```

actual output:

```
panic: readV1RulesIntoV2Sampler unable to unmarshal config: json: cannot unmarshal number into Go struct field EMADynamicSamplerConfig.rule.sampler.emadynamicsampler.adjustmentinterval of type config.Duration

goroutine 1 [running]:
main.convertRulesToNewConfig(0x105213f20?)
        /Users/tylerhelmuth/projects/honeycomb/refinery/tools/convert/ruleconvert.go:165 +0x25c
main.ConvertHelm(0x1400000e348, {0x1052503c8?, 0x14000010060})
        /Users/tylerhelmuth/projects/honeycomb/refinery/tools/convert/main.go:346 +0x55c
main.main()
        /Users/tylerhelmuth/projects/honeycomb/refinery/tools/convert/main.go:216 +0x93c

Process finished with the exit code 2
```

**Additional context**

It appears the the rules convert logic doesn't handle the situation where the old value was an int and the new value is a duration.

TylerHelmuth:
I appears the issue is that the converter does not recursively transfrom the RulesBaseSampler's sampler.

PR #768, opened 2023-07-06
fix: hack a fix for adjustmentinterval not converting inside rules based samples

TylerHelmuth:

- closes https://github.com/honeycombio/refinery/issues/766

- hacks away at an issue where the convert tool couldnt handle the []Rules in the rules based sampler.


cartermp (commented):
I was going to ask about a representative test case, but it appears that testing for this isn't really representative. Given that this is a delightfully hacky fix, I think we should have a test that exemplifies the weird config people could do in the 1.x world that results in this bug.

TylerHelmuth:
@cartermp added another test case that handles the slice issue, but not the config.Duration change.  I'd like to take a look at this issue with fresh eyes anyways, this PR is more to show what the issue is.

tools/convert/ruleconvert.go
```diff
@@ -19,6 +19,14 @@ func transformSamplerMap(m map[string]any) map[string]any {
 		switch val := v.(type) {
 		case map[string]any:
 			v = transformSamplerMap(val)
+		case []any:
```
kentquirk, line 22: The example in the linked issue doesn't seem to show this case. Where is the slice of any?

TylerHelmuth: This case is handling the issue's list of rules:

```
rule:
      - name: test
        condition:
          - field: service.name
            operator: =
            value: test
        sampler:
          EMADynamicSampler:
            Sampler: EMADynamicSampler
            AdjustmentInterval: 60
      - name: default rule
        SampleRate: 1
```

In the current version of the tool we do not attempt to convert any RulesSamplerConfig rules.  Since the rules slice can contain samplers that need adjusted, we need to recurse through the slices as well.

Issue #770, opened 2023-07-07
`GRPCServerParameters.Enabled` is unused.


TylerHelmuth:
**Versions**

- Refinery: 2.0.0

**Steps to reproduce**

1. Set `GRPCServerParameters.Enabled=false` and set `GRPCServerParameters.ListenAddr` to an address
2. Observe that Refinery does start up a grpc endpoint.

**Additional context**

I appears that `GRPCServerParameters.Enabled` is not used anywhere.  Setting `GRPCServerParameters.ListenAddr` is effectively enabling to feature.

PR #771, opened 2023-07-07
fix: Update file_config to honor GRPCServerParameters.Enabled

TylerHelmuth:

- Closes https://github.com/honeycombio/refinery/issues/770

- Add a `GetGRPCEnabled` function to config interface.
- Update file_confg and mock to implement `GetGRPCEnabled`
- Update grpc server logic to check if enabled.
- Update tests

### DIFF
--- a/config.md
+++ b/config.md
@@ -1,7 +1,7 @@
 # Honeycomb Refinery Configuration Documentation
 
 This is the documentation for the configuration file for Honeycomb's Refinery.
-It was automatically generated on 2023-07-10 at 15:15:58 UTC.
+It was automatically generated on 2023-07-10 at 16:05:02 UTC.
 
 ## The Config file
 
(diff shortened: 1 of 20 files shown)


config.md
```diff
@@ -885,6 +885,7 @@ If `false`, then the gRPC server is not started and no gRPC traffic is accepted.
 
 - Not eligible for live reload.
 - Type: `bool`
+- Default: `true`
```
akvanhar, line 888: I thought we said we didn't want to default this to true if we're providing a default port as this will otherwise be a breaking change?

kentquirk: I wasn't in the conversation, but if the intent is to provide an enabled flag with defaults, the enabled flag should default to false.

TylerHelmuth: I have switched it back to default false.

kentquirk (approved):
This part looks good, but I think this also requires a change to configMeta.yaml and maybe the . Specifically, the Enabled flag needs:
```
        valuetype: conditional
        extra: "nonempty GRPCListenAddr"
```
and the "conditional" helper needs to get the "nonempty" case added.

It would be ok to do those as a separate PR, or as part of this one.

Issue #769, opened 2023-07-07
Refinery log level `debug` produces no logs


TylerHelmuth:
**Versions**

- Refinery: 2.0.0

**Steps to reproduce**

1. set `logger.level=debug`
2. Observe no logs

**Additional context**

PR #772, opened 2023-07-07
fix: Move Unknown log level to zero position

TylerHelmuth:

- closes https://github.com/honeycombio/refinery/pull/769

- Moves `Unknown` log level to the 0 position. This is necessary to ensure that https://github.com/honeycombio/refinery/blob/5c30fbf2e2d3a2cd673fd9414f09c7537fe2a505/config/configLoadHelpers.go#L175 doesn't view `Debug` log level as a zero value and overwrite it with `warn`.

### DIFF
--- a/config/level.go
+++ b/config/level.go
@@ -8,12 +8,12 @@ import (
 type Level int
 
 const (
-	DebugLevel Level = iota
+	UnknownLevel Level = iota
+	DebugLevel
 	InfoLevel
 	WarnLevel
 	ErrorLevel
 	PanicLevel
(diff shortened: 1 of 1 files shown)


akvanhar (approved):

kentquirk (changes requested):
I don't understand this. Please explain further? As the now-deleted comment noted, this was the designed behavior, and the conditional was there to implement it. You've changed half the behavior and I don't understand the explanation.

kentquirk (approved):
OK, this was linked to the wrong issue. I fixed the issue text and now agree with it.

PR #774, opened 2023-07-07
fix: Add missing validation for LegacyMetrics APIKey

TylerHelmuth:

- Adds missing validation to `LegacyMetrics.APIKey`.

- Add api key format validation
- Regenerate files


refinery_rules.md
```diff
@@ -77,7 +77,7 @@ The sample rate is calculated from the trace ID, so all spans with the same trac
 
 The duration after which the Dynamic Sampler should reset its internal counters.
 It should be specified as a duration string.
-For example, "30s" or "1m".
+For example,"30s" or "1m".
```
akvanhar, line 80: 🤨 We should probably have the space here. Was it auto-formatted out?

akvanhar (approved):

kentquirk:
The reason this wasn't specified as an apikey format is that we often have example files that specify "yourkeyhere" or something like it -- for classic keys, that won't conform to the format. "SetThisToAHoneycombKey" is a valid modern format. I'm OK with enforcing it but it might cause some problems with demo data.

Issue #773, opened 2023-07-07
The unit for AvailableMemory is not compatible with Kubernetes units


TylerHelmuth:
**Versions**

- Refinery: 2.0.0

**Steps to reproduce**

1. Try to set `AvailableMemory` to something like `2Gi`.
2. See Refinery fail to start.

**Additional context**

We need to allow these units:
- https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-memory
- https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-cpu

Currently the [package we use](https://github.com/docker/go-units) only supports `G` and `GiB`, not `Gi`

PR #778, opened 2023-07-09
fix: Add ability to handle k8s unit format

TylerHelmuth:

- https://github.com/honeycombio/refinery/issues/773

- Updates `MemorySize.UnmarshalText` to be able to handle k8s memory format.
- Add unit tests


kentquirk (approved):
This is a good solution.

PR #780, opened 2023-07-10
fix: Inject all metrics and config later

### LINKED ISSUE #776: Some Refinery metrics are not working correctly
Some Refinery metrics are reporting but not incrementing correctly.  At least the following metrics are always 0:

- peer_router_batch
- dynsampler_sample_rate_*

kentquirk:

- We had missing metrics because the metrics objects that were loaded from config were not being Start()ed.

- Rework MultiMetrics to inject all the internal implementations even if they're not being used.
- On Start(), copy the enabled ones to the children list so that the rest of metrics works right.
- Remove the previous attempt to avoid doing this.
- Also inject All The Things into a test that needs injection.

Fixes #776.


TylerHelmuth (approved):
Best of both worlds

PR #782, opened 2023-07-10
fix: Update validate logic to use MemorySize

TylerHelmuth:

- Related to https://github.com/honeycombio/refinery/issues/773

- Update config validation to use MemorySize custom logic


config/validate_test.go
```diff
@@ -76,6 +76,7 @@ func Test_validateType(t *testing.T) {
 		{"url blank", "k", "", "url", `field k may not be blank`},
 		{"url noscheme", "k", "example.com", "url", `field k (example.com) must be a valid URL with a host`},
 		{"url badscheme", "k", "ftp://example.com", "url", `field k (ftp://example.com) must use an http or https scheme`},
+		{"memorysize", "k", "test", "memorysize", `field k (test) must be a valid memory size like '1Gb' or '100_000_000'`},
```
kentquirk, line 79: I think this isn't enough. You've used RamInBytes, and I had used FromHumanSize, and I think we need *both* (or our own code) to be able to handle all of the cases like Gb, GiB, etc.

TylerHelmuth: I think using `FromHumanSize` was incorrect as the actual Unmarshalling has always used `RamInBytes`. The validator should validate based on what we place to use in the unmarshalling.  

I will add some positive test scenarios that confirms `G`, `Gi`, and `GiB` pass the validation.

TylerHelmuth:
@kentquirk updated the unmarshal logic to appropriately differential between decimal and binary units.  Added lots of tests.

config/file_config.go
```diff
@@ -30,6 +32,54 @@ func (d *Duration) UnmarshalText(text []byte) error {
...
+	Mi = 1024 * Ki
+	Gi = 1024 * Mi
+	Ti = 1024 * Gi
+	Pi = 1024 * Ti
+	Ei = 1024 * Pi
+
+	invalidSizeError = "invalid size: %s"
+)
+
+var binaryMap = map[string]uint64{
```
kentquirk, line 53: this isn't wrong, but why do you have two maps instead of one? The reason to have two is if you need to support one and not the other, but we don't have that need. Unifying these two maps would give you a single lookup and simplify the logic.

kentquirk (approved):
👍

kentquirk (approved):

Issue #790, opened 2023-07-11
Update docker tagging scheme


RainofTerra:
**Is your feature request related to a problem? Please describe.**
Our current docker tagging makes it impossible to to prefer a particular major or minor release, or internally to prefer a particular branch. Also, the "latest" tag is potentially confusing in the case where we need to release a minor update to a previous major version (for example)

**Describe the solution you'd like**
Images based off a release tag (`v2.1.5`) will be tagged with:
major (`2`), major.minor (`2.1`), and semver (`2.1.5`) as well as `latest` (for compatibility)

this will allow users to decide which versions they automatically upgrade to (`major.minor` probably being the recommended one)

Internal images based off a branch will be tagged with `branch-branchname` such as `branch-terra.myawesomebranch`

**Describe alternatives you've considered**
None but it is versioning so I'm sure there are any number of possible versions 😉

**Additional context**

cartermp:
Note for team: consider doing this alongside the work in https://github.com/honeycombio/refinery/issues/869

PR #791, opened 2023-07-11
maint: Docker tagging updates

RainofTerra:
#790 
-

- For internal dev builds, the version will be the build # from circle. They will also be tagged with this build number
- The tags for dev builds will include `branch-branchname` such as `branch-terra.myawesomebranch`
- For internal dev builds for `main` will be tagged with latest
- For releases, we will tag with `major`, `major.minor`, and `semver` in addition to `latest` (for compatibility) such as `2`, `2.1`, and `2.1.5`
- All dev builds will now be available in our ECR repo, I will separately add a lifecycle policy to it so it only keeps so many days of non-release builds (90 maybe?)

### DIFF
--- a/build-docker.sh
+++ b/build-docker.sh
@@ -1,14 +1,42 @@
+#!/usr/bin/env bash
 set -o nounset
 set -o pipefail
 set -o xtrace
 
-TAGS="latest"
+TAGS=""
+# Check if CIRCLE_BRANCH is set and not empty
+if [[ -n "${CIRCLE_BRANCH}" ]]; then
+    export TAGS="${CIRCLE_BUILD_NUM},branch-${CIRCLE_BRANCH}"
+fi
+
+# We only want to tag main with latest on ECR
+if [[ "${CIRCLE_BRANCH}" == "main" ]]; then
+    TAGS+=",latest"
+fi
+
+# Local builds just get "dev" for version
 VERSION="dev"
+
+# If we are doing a dev build on circle, we will version it as the circleci buildnumber
+if [[ -n "${CIRCLE_BUILD_NUM}" ]]; then
+    VERSION="${CIRCLE_BUILD_NUM}"
+fi
+
+# if we're running off a git tag, it is a release which we tag with the versions as well as latest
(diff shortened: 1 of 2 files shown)


build-docker.sh
```diff
@@ -1,14 +1,42 @@
...
-    # append version to image tags
-    TAGS+=",$VERSION"
+
+    # Extract major, major.minor, and major.minor.patch versions
+    MAJOR_VERSION=${VERSION%%.*}
+    MINOR_VERSION=${VERSION%.*}
+
+    # Append versions to image tags
+    # So 2.1.1 would be tagged with "2","2.1","2.1.1"
+    TAGS="$MAJOR_VERSION,$MINOR_VERSION,$VERSION,latest"
```
TylerHelmuth, line 39: Do all the places we publish images handle new images with previously used tags?

build-docker.sh
```diff
@@ -1,14 +1,42 @@
...
+# We only want to tag main with latest on ECR
+if [[ "${CIRCLE_BRANCH}" == "main" ]]; then
+    TAGS+=",latest"
+fi
+
+# Local builds just get "dev" for version
 VERSION="dev"
+
+# If we are doing a dev build on circle, we will version it as the circleci buildnumber
+if [[ -n "${CIRCLE_BUILD_NUM}" ]]; then
```
TylerHelmuth, line 21: Will CIRCLE_BUILD_NUM ever not be set?  Does this case render `VERSION="dev"` obsolete?

RainofTerra: I wasn't sure how people do local builds - do those go through circle?

build-docker.sh
RainofTerra: Yeah, the latest by time stamp is always chosen when there are multiples of a tag for both dockerhub and ECR (and everything else I'm aware of)

build-docker.sh
TylerHelmuth: You're right, you could technically run this script locally.

TylerHelmuth (approved):

PR #794, opened 2023-07-13
fix: Redis scan batch size increase

### LINKED ISSUE #793: Redis peer management doesn't scale that well to 300 instances.
A customer with Many (300) Refinery instances is seeing a thousand Redis timeouts per hour. The Redis peer list updates do SCANs that aren't that fast, and they seem to batch them 10 at a time, plus it repeats twice.

Consider:
* Using a larger value for COUNT in the scan() function
* Making that configurable
* Replacing the individual keys with a Redis SET and fetching them all with a single call

Some testing with setting COUNT to 1000 had this result:

```
BEFORE
  avg duration=33.230184ms count=36569 max duration=389.308ms    min duration=367.458µs
AFTER
  avg duration= 4.999529ms count=44768 max duration=166.902542ms min duration=190.833µs
```

The SET option doesn't have the automatic key expiration feature, so I think we'll just increase the value for COUNT and be happy about a 6.5x average performance increase for now.

nullren:

This week, we've bumped up our number of refinery pods from 250 to 300. After that change, we've noticed a lot more timeouts from redis.

```
time="2023-07-13T17:49:28Z" level=error msg="redis scan encountered an error" err="redis scan timeout" keys_returned=92 timeoutSec=5s
time="2023-07-13T18:39:41Z" level=error msg="get members failed during watch" error="context deadline exceeded" name="http://host:8081/" oldPeers="[list_of_the_300_hosts:8081]" timeout=5s
```

After digging into some of the code, we've noticed that peers are stored in redis and are retrieved in batches of 10. Because there are 300 nodes, each node is making 30 requests to redis to get all the peers, and doing it twice, so 60 total. This is done every 5 seconds. So our redis instance is handling 18000 requests every 5 seconds.

After a little bit of napkin math, with 300 nodes and an average payload size of   26 bytes (that's what we see reported by `oldPeers` in our logs, anyway), redis has within it about 7.8kb of data. That should be very easily retrieved in a single request and doesn't need to be broken down in batches of 10.

This PR proposes bumping that up to 1000. That would give us a total payload of about 26kb. For 1000 nodes, that would also only yield 2000 requests to redis every 5 seconds (total 72mb bandwidth for those 5 seconds, too). Leaving it as is would require 200,000 requests to redis over 5 seconds for 1000 nodes.

Getting a little extreme, for 1M nodes, that would give total payload of about 26MB and yield 2000 requests per node, so 2B requests to redis every 5 seconds for a total of 72GB data transferred (1/1000 of the total payload 26MB, 2000 times). This might kill the redis, but the data is still "pretty small".

Fixes #793

### DIFF
--- a/internal/redimem/redimem.go
+++ b/internal/redimem/redimem.go
@@ -37,7 +37,7 @@ const (
 	redisScanTimeout = 5 * time.Second
 
 	// redisScanBatchSize indicates how many keys to retrieve from Redis at a time.
-	redisScanBatchSize = "10"
+	redisScanBatchSize = "1000"
 )
 
 // RedisMembership implements the Membership interface using Redis as the backend
--- a/internal/redimem/redimem.go
+++ b/internal/redimem/redimem.go
@@ -35,6 +35,9 @@ const (
 
 	// redisScanTimeout indicates how long to attempt to scan for peers.
 	redisScanTimeout = 5 * time.Second
+
+	// redisScanBatchSize indicates how many keys to retrieve from Redis at a time.
+	redisScanBatchSize = "10"
 )
 
 // RedisMembership implements the Membership interface using Redis as the backend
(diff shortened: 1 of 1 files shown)


kentquirk (approved):
This is great, thank you. 

Apparently there's a build issue, but I'll fix ASAP and get this integrated.

TylerHelmuth:
Should we make this configurable?

kentquirk:
I don't think there's need for it to be configurable; this number gives us up to 1000 refineries in a single request, and if we end up with networks needing more than 1000 refineries we probably need to be looking at a different peer management strategy anyway. I think the original design was just thinking too small.

TylerHelmuth (approved):

PR #796, opened 2023-07-14
maint: replace slash with dash in branch name

TylerHelmuth:

- the build-docker.sh script has started including the branch name as a tag, but that breaks CI if the branch name includes a slash (`/`), which are not allowed in tags
- Unable to run CI against external PRs

- Updated the script to replace `/` with `-` in branch names.


kentquirk (approved):
Thanks.

PR #805, opened 2023-07-21
feat: add a field with the formula used to decide to activate stress relief

RainofTerra:

- Currently we don't emit the values we used to determine if stress relief should be activated unless you enable debug logging, but I think the numbers at the time of activation are valuable enough to be present at `WARN`

- adds a `stress_formula` field to the activation message, and adds `formula` to the `StressRelief` struct

### DIFF
--- a/collect/stressRelief.go
+++ b/collect/stressRelief.go
@@ -1,6 +1,7 @@
 package collect
 
 import (
+	"fmt"
 	"math"
 	"sync"
 	"time"
@@ -52,6 +53,7 @@ type StressRelief struct {
 	upperBound      uint64
 	stressLevel     uint
 	reason          string
+	formula         string
 	stressed        bool
 	stayOnUntil     time.Time
 	minDuration     time.Duration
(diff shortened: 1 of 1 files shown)


collect/stressRelief.go
```diff
@@ -268,11 +270,13 @@ func (s *StressRelief) Recalc() {
...
+	var formula string
 	for _, c := range s.calcs {
 		stress := 100 * s.algorithms[c.Algorithm](c.Numerator, c.Denominator)
 		if stress > level {
 			level = stress
 			reason = c.Reason
+			formula = fmt.Sprintf("%s(%v/%v)=%v", c.Algorithm, c.Numerator, c.Denominator, stress)
 		}
 	}
 	s.Logger.Debug().WithField("stress_level", level).WithField("reason", reason).Logf("calculated stress level")
```
kentquirk, line 282: I think we should add formula to this log message too, so you can see it with debug on, even if you're not actually in Monitor mode.

kentquirk (approved):
I think this is a good idea in general; one nit. Thanks.

collect/stressRelief.go
RainofTerra: Good call! will push that in a moment

PR #806, opened 2023-07-24
feat: Validate Collection memory config through adding `conflictsWith` validation

### LINKED ISSUE #803: [config] Validation should fail if MaxAlloc, AvailableMemory, and MaxMemoryPercentage are all set

Refinery currently allows setting MaxAlloc, AvailableMemory, and MaxMemoryPercentage all in the same config.  In this instance, AvailableMemory and MaxMemoryPercentage will be used.  This can lead to confusion because a more explicit field, MaxAlloc, is also set. 

Refinery falis validation if all 3 are set

**Additional context**

We ran into this with the Refinery helm chart.  The chart sets AvailableMemory and MaxMemoryPercentage by default, but nothing stops a user from setting MaxAlloc in their values.yaml.  This makes it seem like MaxAlloc will be used, but unless they remove AvailableMemory and MaxMemoryPercentage the resulting config will have all 3 values and use AvailableMemory and MaxMemoryPercentage instead.

MasonLegere:

Fixes https://github.com/honeycombio/refinery/issues/803 by ensuring that only both `AvailableMemory` and `MaxAlloc` cannot be set at the same. `MaxMemoryPercentage` may still be set regardless as it has a default value defined. 

While making this changes I noticed that when only `AvailableMemory` was defined I ran into a validation `requireWith` error. The validation only fails on the first pass before defaults are loaded in from the config struct. If the failure is intentional then we probably should remove the default -- only breaking change I could see is if someone was running with validations turned off explicitly. 

```bash
 $ docker-compose up
[+] Running 1/0
 ✔ Container refinery-refinery-1  Recreated                                                                   0.0s
Attaching to refinery-refinery-1
refinery-refinery-1  | 2023/07/24 07:57:40 maxprocs: Leaving GOMAXPROCS=6: CPU quota undefined
refinery-refinery-1  | Validation failed for config file /etc/refinery/config.yaml:
refinery-refinery-1  |   the group Collection includes AvailableMemory, which also requires MaxMemoryPercentage
refinery-refinery-1  |
refinery-refinery-1  |
refinery-refinery-1 exited with code 1
$ cat config.yaml 
General:
  ConfigurationVersion: 2
  MinRefineryVersion: v2.0


Collection:
  AvailableMemory: 2Gb

```

- Adds `conflictsWith` validation type acting in an opposite way to `requireWith` 
- Modified `requireWith` to check whether or not the required field has a default value

Fixes #803.

### DIFF
--- a/config/metadata/configMeta.yaml
+++ b/config/metadata/configMeta.yaml
@@ -160,8 +160,7 @@ groups:
           accepted. Events arriving with API keys not in the `ReceiveKeys` list
           will be rejected with an HTTP `401` error.
 
-          If `false`, then all traffic is accepted and `ReceiveKeys` is
-          ignored. Must be specified if `ReceiveKeys` is specified.
+          If `false`, then all traffic is accepted and `ReceiveKeys` is ignored.
 
   - name: RefineryTelemetry
     title: "Refinery Telemetry"
@@ -935,7 +934,7 @@ groups:
           allocation and `MaxAlloc` will be used instead. If set, then this must
           be a memory size. 64-bit values are supported. Sizes with standard
           unit suffixes (`MB`, `GiB`, etc.) and Kubernetes units (`M`, `Gi`, 
-          etc.) are also supported. If set, `Collections.maxAlloc` must not be
+          etc.) are also supported. If set, `Collections.MaxAlloc` must not be
           defined.
 
       - name: MaxMemoryPercentage
--- a/config/metadata/configMeta.yaml
+++ b/config/metadata/configMeta.yaml
@@ -935,7 +935,8 @@ groups:
           allocation and `MaxAlloc` will be used instead. If set, then this must
           be a memory size. 64-bit values are supported. Sizes with standard
           unit suffixes (`MB`, `GiB`, etc.) and Kubernetes units (`M`, `Gi`, 
-          etc.) are also supported.
+          etc.) are also supported. If set, `Collections.maxAlloc` must not be
+          defined.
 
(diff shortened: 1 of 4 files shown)


config/metadata/configMeta.yaml
```diff
@@ -935,7 +935,8 @@ groups:
           allocation and `MaxAlloc` will be used instead. If set, then this must
           be a memory size. 64-bit values are supported. Sizes with standard
           unit suffixes (`MB`, `GiB`, etc.) and Kubernetes units (`M`, `Gi`, 
-          etc.) are also supported.
+          etc.) are also supported. If set, `Collections.maxAlloc` must not be
```
kentquirk, line 938:
```suggestion
          etc.) are also supported. If set, `Collections.MaxAlloc` must not be
```

config/validate.go
```diff
@@ -368,10 +368,17 @@ func (m *Metadata) Validate(data map[string]any) []string {
 					// if the named key is specified then this one must be also
 					otherName := validation.Arg.(string)
```
kentquirk, line 369: Suggest adding a line that says "this validation cannot be used if the field has a default value".

kentquirk (approved):
This is good, thanks. Minor tweaks, please.

config/metadata/configMeta.yaml
MasonLegere: Ops thanks, I also found that the `AcceptOnlyListedKeys` docs need to be changed as well as it currently has the same issue (using requiredWith along with a default).

PR #811, opened 2023-07-31
fix: add missing metrics fields

### LINKED ISSUE #801: OTel metrics is missing a few fields
Report through support:

With legacy metrics in Refinery 2.0, we get [memory_in_use](https://github.com/honeycombio/refinery/blob/f295a8cbe2c49e5ad3a1e5855db21c3b3209f570/metrics/legacy.go#L137C32-L137C44) and hostname out of the box. But a customer who just rolled over to OTel metrics isn't seeing these fields come in. 

Turns out these fields are dynamically injected into legacy metrics by a mechanism in libhoney. We need to also do this in otel metrics.

kentquirk:

- Fixes #801. OTelMetrics was missing some fields that were present in LegacyMetrics. This adds them.
- Adds new sample rate metrics

- Add mechanism that adds the dynamic fields to otel.
- Add some histograms for aggregate sample rate and kept sample rate.

### DIFF
--- a/metrics/otel_metrics.go
+++ b/metrics/otel_metrics.go
@@ -119,6 +119,7 @@ func (o *OTelMetrics) Start() error {
 		resource.WithAttributes(resource.Default().Attributes()...),
 		resource.WithAttributes(attribute.KeyValue{Key: "service.name", Value: attribute.StringValue("refinery")}),
 		resource.WithAttributes(attribute.KeyValue{Key: "service.version", Value: attribute.StringValue(o.Version)}),
+		resource.WithAttributes(attribute.KeyValue{Key: "host.name", Value: attribute.StringValue(hostname)}),
 		resource.WithAttributes(attribute.KeyValue{Key: "hostname", Value: attribute.StringValue(hostname)}),
 	)
 
@@ -151,10 +152,9 @@ func (o *OTelMetrics) Start() error {
 	name = "memory_inuse"
 	// This is just reporting the gauge we already track under a different name.
 	var fmem metric.Float64Callback = func(_ context.Context, result metric.Float64Observer) error {
-		o.lock.RLock()
+		// this is an 'ok' value, not an error, so it's safe to ignore it.
 		v, _ := o.Get("memory_heap_allocation")
 		result.Observe(v)
-		o.lock.RUnlock()
 		return nil
 	}
 	g, err = o.meter.Float64ObservableGauge(name, metric.WithFloat64Callback(fmem))
--- a/metrics/otel_metrics.go
+++ b/metrics/otel_metrics.go
@@ -3,6 +3,8 @@ package metrics
 import (
 	"context"
 	"net/url"
+	"os"
+	"runtime"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


metrics/otel_metrics.go
```diff
@@ -105,10 +107,19 @@ func (o *OTelMetrics) Start() error {
...
+		hostname = "unknown: " + err.Error()
+	} else {
+		hostname = hn
+	}
+
 	res, err := resource.New(ctx,
 		resource.WithAttributes(resource.Default().Attributes()...),
 		resource.WithAttributes(attribute.KeyValue{Key: "service.name", Value: attribute.StringValue("refinery")}),
 		resource.WithAttributes(attribute.KeyValue{Key: "service.version", Value: attribute.StringValue(o.Version)}),
+		resource.WithAttributes(attribute.KeyValue{Key: "hostname", Value: attribute.StringValue(hostname)}),
```
TylerHelmuth, line 122: Can we use the semantic convention name `host.name`: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/resource/host.md#host

metrics/otel_metrics.go
```diff
@@ -124,6 +135,46 @@ func (o *OTelMetrics) Start() error {
...
+	if err != nil {
+		return err
+	}
+	o.gauges[name] = g
+
+	name = "memory_inuse"
+	// This is just reporting the gauge we already track under a different name.
+	var fmem metric.Float64Callback = func(_ context.Context, result metric.Float64Observer) error {
+		o.lock.RLock()
+		v, _ := o.Get("memory_heap_allocation")
```
TylerHelmuth, line 155: What is the value of `v` if this returns an error (I'm assuming `_` is an error)?  Do we need to capture that error?

metrics/otel_metrics.go
kentquirk: we could add that as a duplicate, but there are TONS of boards and queries using "hostname". I'd be very reluctant to change it. I suppose we could do it differently for OTel metrics vs legacy metrics, but we'd definitely have to tell people we did it.

metrics/otel_metrics.go
kentquirk: It's not an error, it's an "ok" value telling you that the value you queried was in the dataset. v will be 0 in that case, which should never occur because this is duplicating the other metric. Get is basically just a map query, but the ok value is not optional in that case. I mean, I could return a "can't happen" error here, but 0 is probably better.

metrics/otel_metrics.go
```diff
@@ -124,6 +135,46 @@ func (o *OTelMetrics) Start() error {
...
+		result.Observe(v)
+		o.lock.RUnlock()
```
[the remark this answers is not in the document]
kentquirk: The resource is usually unlocked; it's competing locks that cause the problem. 

However, I realized in looking at it that Get is already doing an RLock, so this one is redundant. Removing it.

metrics/otel_metrics.go
TylerHelmuth: I forgot the other metrics were already using hostname. Duplicating is best.

TylerHelmuth (approved):

Issue #592, opened 2023-01-31
Define constants for rules operators 


fchikwekwe:
We have several operators defined in `rules.go` that are reused in `sampler_config.go` for datatype matching. It would be a good idea to define constants for these operators to keep both places on consistent. The list as it stands seems to be. 
```
"exists"
"not-exists"
"!="
"="
 ">"
">="
"<"
"<="
"starts-with"
"contains"
"does-not-contain"
```

Figure out id there any that are missing and define them in `rules.go`.

kentquirk:
Closed by #813

PR #813, opened 2023-07-31
maint: convert hardcoded operators to constants

kentquirk:

- This was intended to be purely a refactoring to pull out constant values
- But it uncovered a bug -- we had a switch where we weren't checking the ">=" case!

- Constantize all the things
- Add the missing switch case
- Write some tests (not exhaustive but try to prove all the major spots work)
- Fix the boolean bug that was also uncovered by the tests

### DIFF
--- a/config/sampler_config.go
+++ b/config/sampler_config.go
@@ -6,7 +6,7 @@ import (
 	"strings"
 )
 
-// Define some constants for comparison operators
+// Define some constants for rule comparison operators
 const (
 	NEQ = "!="
 	EQ  = "="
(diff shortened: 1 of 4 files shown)


TylerHelmuth (commented):
With the a new case I expected a new test.  Did we have a test already that was passing when it should not have?

fchikwekwe (approved):
Great work, this actually closes #592 as well so great job.

fchikwekwe:
Oh, duh. I just now saw the reference to that issue in your branch name now.

PR #814, opened 2023-08-01
feat: Add `has-root-span` operator to rules

### LINKED ISSUE #610: Feature: allow testing whether traces have a root span

Traces without a root span can happen for a wide variety of reasons -- but some customers would prefer to treat such traces differently in their sampling -- possibly even drop them.

Refinery has enough information to do this, but right now, there is no way to express this desire in Refinery's query language. 

The rules-based sampler should allow expressing this condition so that it can be combined with other rules.

One solution is that we could add an operator similar to `exists` -- call it `has-root-span`. It would ignore any `field` specifier, and the `value` would be evaluated as a boolean. It would be evaluated like other rules, would be invalid in `span` scope, and look something like this in practice:

```yaml
                - Name: drop incomplete traces from the buggy service
                  Drop: true
                  Conditions:
                    - Operator: has-root-span
                      Value: false
                    - Field: service.name
                      Operator: =
                      Value: buggy-service
```

There doesn't seem to be one if you want to do this.

kentquirk:
Dependent on #813; that should be merged before this one.

- Fixes #610 

- Adds a new rule operator, `has-root-span`, that evaluates to a boolean based on whether the root span for the trace has arrived when the rules are evaluated. 
- Add tests for it
- Update metadata (but does not regenerate the docs yet; that will come before the release)

### DIFF
--- a/sample/rules.go
+++ b/sample/rules.go
@@ -149,6 +149,20 @@ func ruleMatchesTrace(t *types.Trace, rule *config.RulesBasedSamplerRule, checkN
 	var matched int
 
 	for _, condition := range rule.Conditions {
+		// This condition is evaluated for the trace as a whole.
+		// If RootSpan is nil, it means the trace timer has fired or the trace has been
+		// ejected from the cache before the root span has arrived.
+		if condition.Operator == config.HasRootSpan {
+			if (t.RootSpan != nil) == config.TryConvertToBool(condition.Value) {
+				matched++
+				continue
+			} else {
+				// if HasRootSpan is one of the conditions and it didn't match,
+				// there's no need to check the rest of the conditions.
+				return false
+			}
+		}
+
 	span:
 		for _, span := range t.GetSpans() {
 			value, exists := extractValueFromSpan(span, condition, checkNestedFields)
--- a/sample/rules_test.go
+++ b/sample/rules_test.go
@@ -591,6 +591,99 @@ func TestRules(t *testing.T) {
 			ExpectedKeep: false,
 			ExpectedRate: 0,
 		},
+		{
+			Rules: &config.RulesBasedSamplerConfig{
+				Rules: []*config.RulesBasedSamplerRule{
+					{
+						Name:       "Check that root span is missing",
+						Drop:       true,
+						SampleRate: 0,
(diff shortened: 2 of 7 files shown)


sample/rules.go
```diff
@@ -149,6 +149,20 @@ func ruleMatchesTrace(t *types.Trace, rule *config.RulesBasedSamplerRule, checkN
 	var matched int
 
 	for _, condition := range rule.Conditions {
+		// This condition is evaluated for the trace as a whole.
+		// If RootSpan is nil, it means the trace timer has fired or the trace has been
+		// ejected from the cache before the root span has arrived.
+		if condition.Operator == config.HasRootSpan {
+			if (t.RootSpan != nil) == config.TryConvertToBool(condition.Value) {
```
TylerHelmuth, line 156: Is there any thread safety we need to include here?  Is there a chance the root span arrives while we're in this function?

TylerHelmuth (approved):

sample/rules_test.go
```diff
@@ -591,6 +591,99 @@ func TestRules(t *testing.T) {
...
+							"trace.parent_id": "54321",
+							"test":            int64(3),
+						},
+					},
+				},
+			},
+			ExpectedName: "Check that root span is present",
+			ExpectedKeep: true,
+			ExpectedRate: 99,
+		},
```
fchikwekwe, line 686: The only other cases I'd consider are 1) the case where there is a root span present, but no other spans in the trace  and 2) a case where there is only a non-root span in the trace. This is just to see if your logic below for setting root span when there is only one span works.

fchikwekwe:
Oh, I made a request prematurely, it looks like those cases are already covered by pre-existing tests. All looks good then.

fchikwekwe (approved):

sample/rules.go
kentquirk: No; at this point the trace has been removed from the cache and we're iterating the list of traces that were removed. There is a slight gap in the time between when the traces are pulled from the cache until we make and record the trace decision, but that's a longstanding issue in refinery. It's probably the cause of the occasional missing root span where the root arrives at the exact instant it's been pulled out of cache.

Issue #725, opened 2023-06-21
Allow peers to gossip news of a new config


kentquirk:
This describes an algorithm for allowing peers in a cluster to share news of a new configuration, rather than hammering a configuration store with repeated requests.

A 2.x Refinery can load configurations remotely (via a url) but because of that, it can no longer listen for changes to the config -- it has to actively poll. 

In a fleet of refineries, we want to allow them to take turns randomly checking for a changed config, and then when one is seen, report that information.

The refineries are configured to check config every T seconds. They actually create two timers:
* A repeating ticker that fires every T seconds
* A one-shot random timer in the range [T*N/2, T*N*2] where N is the number of nodes in the cluster.

Whenever a config is loaded (on startup or because of a hash disagreement)
* It reads the config from the source location and calculates a hash, which it stores in its own state as the calculated hash
* It loads the current hash from redis (or central service)  (if there is none, it uses 0)
* If the calculated hash is not equal to the current hash, it writes the value to redis as the current hash

Whenever the one-shot timeout occurs:
* It loads the config as above
* It sets a new one-shot timeout within the specified range

Whenever the repeating timeout occurs:
* It reads the current hash from redis
* It compares to the calculated hash
* If it's different, it loads the config (see above)

The result is that on average, the config is reloaded only once every T seconds from the source no matter how many nodes there are, but all the nodes are notified of changes

MasonLegere:
If someone wanted to manually change config for testing in a single pod would we then have a flag for only referencing the local store?

kentquirk:
There are circumstances where every node needs a different config, and unless we also decide to support some form of layered config (load a common base and then apply local diffs) then we'd also have to have a flag that says "this is not shared".

One note -- this was the original intention in v2 and was dropped for lack of time - but it's documented in the metadata that it works this way. We need to either support it or fix the documentation.

kentquirk:
Kicking this out of 2.1; it's too complex and risky in a world where no one is yet using URLs for config. Will revisit as we look at other config options.

VinozzZ:
Implemented in https://github.com/honeycombio/refinery/pull/1241

PR #823, opened 2023-08-02
fix: Change default ConfigReloadInterval, add jitter, fix docs

kentquirk:

- The default config reload interval was too long, and the documentation for it was wrong.

- Change default for config reload to 15s from 5m. The original intent (see #725) was to have a fleet of refineries gossiping news of new configs, but that's tricky to get right and probably not needed, especially since for now at least, almost all installations will be using a config file, not a config url. 
- Add a random jitter to avoid thundering herds. 
- Update metadata to explain what it's really doing
- Regenerate documentation.

### DIFF
--- a/config.md
+++ b/config.md
@@ -1,7 +1,7 @@
 # Honeycomb Refinery Configuration Documentation
 
 This is the documentation for the configuration file for Honeycomb's Refinery.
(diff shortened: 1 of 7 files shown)


TylerHelmuth (approved):
@kentquirk once this is merged can you open an issue for the helm chart so we remember to update the default values.yaml to set this to 0s?

PR #827, opened 2023-08-02
feat: Allow setting throughput for the cluster

### LINKED ISSUE #533: Throughput sampler should operate at the cluster level rather than the individual instance

The ThroughputSampler only monitors throughput of an individual instance, not the cluster as a whole. This leads to surprising behavior when operators scale their cluster -- the volume of traces reaching Honeycomb changes based on cluster size. It also leads to different (inconsistent) sample rates from different instances. 

* ThroughputSampler should be able to share its throughput data via the Redis instance being used for sharing peer information
* The `GoalThroughputPerSec` and `FieldList` should be common across the cluster
* If #525 is implemented, the `throttle` value should also be common

**Additional context**

### LINKED ISSUE #608: Allow configuration values to be specified for the cluster rather than the instance
There are configuration values in Refinery, such as the GoalThroughputPerSec value used by the TotalThroughputSampler, that are calculated based on the value for a single instance, but the customer experience is to see refinery as an entire cluster. Can we find a way to allow these values to be adjusted by refinery based on the number of peers in the cluster?

This makes it easier to think about as well as meaning that the value wouldn't have to be adjusted when cluster size changes.

kentquirk:

- Throughput samplers now have an extra flag, `UseClusterSize`, which causes the `GoalThroughputPerSec` to be divided by the number of Refinery nodes in the cluster. This way, when cluster size changes, the configuration doesn't need to be changed.

- Add the flag to the config
- Update metadata for the new flag
- Add code to 3 throughput samplers to support it
- Write tests to verify it works
- Regenerate docs

Fixes #533.
Fixes #608.

### DIFF
--- a/config/metadata/rulesMeta.yaml
+++ b/config/metadata/rulesMeta.yaml
@@ -265,6 +265,8 @@ groups:
         type: int
         validations:
           - type: requiredInGroup
+          - type: minimum
+            arg: 1
         summary: is the desired throughput per second of events sent to Honeycomb.
         description: >
           The desired throughput **per second**. This is the number of events
@@ -274,6 +276,14 @@ groups:
           cluster has multiple instances, then you will need to divide your
           total desired sample rate by the number of instances to get this
           value.
+      - name: UseClusterSize
+        type: bool
+        summary: indicates whether to use the cluster size to calculate the goal throughput.
+        description: >
+          Indicates whether to use the cluster size to calculate the goal
+          throughput. If `true`, then the goal throughput will be divided by the
+          number of instances in the cluster. If `false`, then the goal
+          throughput will be the value specified in `GoalThroughputPerSec`.
       - name: InitialSampleRate
         type: int
         summary: is the sample rate to use.
@@ -363,6 +373,8 @@ groups:
         type: int
         validations:
           - type: requiredInGroup
+          - type: minimum
+            arg: 1
         summary: is the desired throughput per second of events sent to Honeycomb.
         description: >
           The desired throughput **per second**. This is the number of events
@@ -372,6 +384,14 @@ groups:
           cluster has multiple instances, then you will need to divide your
           total desired sample rate by the number of instances to get this
           value.
+      - name: UseClusterSize
+        type: bool
+        summary: indicates whether to use the cluster size to calculate the goal throughput.
+        description: >
+          Indicates whether to use the cluster size to calculate the goal
+          throughput. If `true`, then the goal throughput will be divided by the
+          number of instances in the cluster. If `false`, then the goal
+          throughput will be the value specified in `GoalThroughputPerSec`.
       - name: UpdateFrequency
         type: duration
         summary: is how often the sampling rate is recomputed.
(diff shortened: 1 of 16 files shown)


sample/ema_throughput.go
```diff
@@ -78,6 +81,8 @@ func (d *EMAThroughputSampler) Start() error {
 func (d *EMAThroughputSampler) GetSampleRate(trace *types.Trace) (rate uint, keep bool, reason string, key string) {
 	key = d.key.build(trace)
 	count := int(trace.DescendantCount())
+	// update the dynsampler with the latest goal throughput in case cluster size changed
+	d.dynsampler.GoalThroughputPerSec = d.goalThroughputPerSec / d.GetClusterSize(d.useClusterSize)
```
TylerHelmuth, line 85: Looking through the emathroughput sampler in dynsampler I don't see where `GoalThroughputPerSec` is used within `GetSampleRate`.  Whats happening that requires us to update `GoalThroughputPerSec` here?

sample/sample.go
```diff
@@ -18,9 +20,38 @@ type Sampler interface {
...
+func (s *SamplerFactory) GetClusterSize(useClusterSize bool) int {
+	if useClusterSize {
+		s.peerCountLock.RLock()
+		defer s.peerCountLock.RUnlock()
+		return s.peerCount
+	}
+	return 1
+}
+
+func (s *SamplerFactory) Start() error {
```
TylerHelmuth, line 40: Where does this end up getting invoked?

config/metadata/rulesMeta.yaml
```diff
@@ -265,6 +265,8 @@ groups:
         type: int
         validations:
           - type: requiredInGroup
+          - type: minimum
+            arg: 1
```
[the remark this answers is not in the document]
kentquirk: Hmm. I guess it is, technically -- the existing samplers treat it as an error and force the value to 100. 

I'm having a hard time imagining someone depending on that behavior, but if you think it's worrisome, I'll remove this part. I don't think it's worth the pain of bumping the config format to get this extra validation.

sample/ema_throughput.go
kentquirk: It's not used directly within GetSampleRate, it's used when the rates are recalculated after every AdjustmentInterval. But I hadn't thought of a way to not do it on every call to GetSampleRate. 

However, now that you point this out, I think I could instead register a callback that would adjust it and we wouldn't have to set it on every call. I'm going to try that and see how it feels.

sample/sample.go
kentquirk: Automatically as part of the injection of SamplerFactory from `app.go`.

config/metadata/rulesMeta.yaml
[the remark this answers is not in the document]
kentquirk: I just realized it's a required field, so someone would have to say `GoalThroughputPerSec: 0` and intend it to have the value 100.

sample/ema_throughput.go
kentquirk: I've done it this way and I think it's a little cleaner. Please look again.

TylerHelmuth: Ya I like this a lot

sample/ema_throughput.go
```diff
@@ -78,11 +81,16 @@ func (d *EMAThroughputSampler) Start() error {
 	return nil
 }
 
+func (d *EMAThroughputSampler) SetClusterSize(size int) {
+	if d.useClusterSize {
+		d.clusterSize = size
```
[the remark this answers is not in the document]
kentquirk: No need, because its caller does that check already.

TylerHelmuth (approved):

kamalmarhubi:
Amazing!

PR #832, opened 2023-08-04
docs: Touch up readme.

### LINKED ISSUE #560: Improve refinery documentation regarding behavior in an environments-centric world
The documentation regarding refinery's behavior in classic mode vs modern mode is lacking. The sample rules files use datasets rather than environments, and we should also update README as well as official Honeycomb docs (which are basically correct but still could use some updates).

#558 wanted to fix this, but it's not complete. This should be the responsibility of a Honeycomb employee.

kentquirk:
Some edits to clean up the readme which hadn't been touched in a while. 

There's still more that could be done, but I think this is quite a bit better.

Closes #560.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,34 +12,34 @@ For more readable information about recent changes, please see [RELEASE_NOTES.md
 
 ## Purpose
 
-Refinery is a trace-aware tail-based sampling proxy. It examines whole traces and intelligently applies sampling decisions (whether to keep or discard) to each trace.
+Refinery is a tail-based sampling proxy that operates at the the level of an entire trace. It examines whole traces and intelligently applies sampling decisions (whether to keep or discard) to each trace.
 
-Tail-based sampling allows you to inspect a whole trace and make a decision to sample based on its contents. For example, the root span has the HTTP status code to serve for a request, whereas another span might have information on whether the data was served from a cache. Using Refinery, you can choose to keep only traces that had a `500` status code and were also served from a cache.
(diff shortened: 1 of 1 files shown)


cartermp (approved):

README.md
```diff
@@ -58,54 +58,28 @@ This will use the default values file. You can also supply your own:
...
+When operating in a cluster, Refinery expects to gather all of the spans in a trace onto a single instance so that it can make a trace decision. Since each span arrives independently, each Refinery instance needs to be able to communicate with all of its peers in order to distribute traces to the correct instance.
 
-- Peer list: this is a list of all the other servers participating in this Refinery cluster. Traces are evenly distributed across all available servers, and any one trace must be concentrated on one server, regardless of which server handled the incoming spans. The peer list lets the cluster move spans around to the server that is handling the trace. (Not used in the Redis-based config.)
+This communication can be managed in two ways: via an explicit list of peers the configuration file, or by using self-registration via a shared Redis cache. Installations should generally prefer to use Redis. Even in large installations, the load on the Redis server is quite light, with each instance only making a few requests per minute. A single Redis instance with fractional CPU is usually sufficient.
```
TylerHelmuth, line 65:
```suggestion
This communication can be managed in two ways: via an explicit list of peers in the configuration file, or by using self-registration via a shared Redis cache. Installations should generally prefer to use Redis. Even in large installations, the load on the Redis server is quite light, with each instance only making a few requests per minute. A single Redis instance with fractional CPU is usually sufficient.
```

TylerHelmuth (approved):

README.md
kentquirk: Whoops, thanks. I'll fix in the prep pr.

PR #837, opened 2023-08-09
docs: Refinery Readme improvements

mjingle:

This PR updates the Refinery README file, based on feedback and on the content currently lives in https://docs.honeycomb.io/manage-data-volume/refinery/

Changes include:
- Update from `.toml` to `.yaml` file
- Sampling Types descriptions and examples
- Improve Troubleshooting Configuration section
- Improve Dry Run Mode section


README.md
```diff
@@ -33,43 +36,56 @@ Refinery is designed to sit within your infrastructure where all traces can reac
...
-Every Refinery instance should have a minimum of 2GB RAM and at least 2 cores, preferably more. Refinery clusters usually run best with fewer, larger servers; for large installations, 16 cores and 16GB of RAM per server is not unusual (with appropriate tuning).
+Every Refinery instance should have a minimum of:
 
+- a `linux/amd64` or `linux/arm64` operating system
+- 2GB RAM for each server used
+- Access to 2 cores for each server used
+
+In many cases, Refinery only needs one node.
+Refinery clusters usually run best with fewer, larger servers; for large installations, 16 cores and 16GB of RAM per server is not unusual (with appropriate tuning).
+If experiencing a large volume of traffic, you may need to scale out to multiple nodes, and likely need a small Redis cluster to handle scaling.
```
kentquirk, line 51: I am uncomfortable with this phrasing. If you have more than 1 node, you need redis (unless you're really hardcore but I don't want to encourage it). You probably don't need more than one redis, though, unless you're just paranoid.

README.md
```diff
@@ -33,43 +36,56 @@ Refinery is designed to sit within your infrastructure where all traces can reac
...
 
 When operating in a cluster, Refinery expects to gather all of the spans in a trace onto a single instance so that it can make a trace decision. Since each span arrives independently, each Refinery instance needs to be able to communicate with all of its peers in order to distribute traces to the correct instance.
 
 This communication can be managed in two ways: via an explicit list of peers in the configuration file, or by using self-registration via a shared Redis cache. Installations should generally prefer to use Redis. Even in large installations, the load on the Redis server is quite light, with each instance only making a few requests per minute. A single Redis instance with fractional CPU is usually sufficient.
 
-
 ## Configuration
 
-Configuration is controlled by Refinery's two configuration files. For information on all of the configuration parameters that control Refinery's operation, please see [the configuration documentation](https://docs.honeycomb.io/manage-data-volume/refinery/configuration/).
+Configuration is controlled by Refinery's two configuration files - `config.yaml` for general configuration and `rules.yaml` for sampling configuration.
```
kentquirk, line 84: I deliberately didn't give them names. I won't stop you, but in the past I think it has caused more confusion than it has helped.

README.md
```diff
@@ -79,41 +95,44 @@ Refinery is a typical linux-style command line application, and supports several
...
 
 ## Understanding Regular Operation
 
 Refinery emits a number of metrics to give some indication about the health of the process. These metrics should be sent to Honeycomb, typically with Open Telemetry, and can also be exposed to Prometheus. The interesting ones to watch are:
 
 - Sample rates: how many traces are kept / dropped, and what does the sample rate distribution look like?
-- [incoming|peer]_router_\*: how many events (no trace info) vs. spans (have trace info) have been accepted, and how many sent on to peers?
-- collect_cache_buffer_overrun: this should remain zero; a positive value indicates the need to grow the size of the collector's circular buffer (via configuration `CacheCapacity`).
-- process_uptime_seconds: records the uptime of each process; look for unexpected restarts as a key towards memory constraints.
+- `[incoming|peer]_router_\*`: how many events (no trace info) vs. spans (have trace info) have been accepted, and how many sent on to peers?
```
kentquirk, line 133:
```suggestion
- `[incoming|peer]_router_*`: how many events (no trace info) vs. spans (have trace info) have been accepted, and how many sent on to peers?
```

kentquirk: This was just a glitch.

README.md
```diff
@@ -79,41 +95,44 @@ Refinery is a typical linux-style command line application, and supports several
...
+- `collect_cache_buffer_overrun`: this should remain zero; a positive value indicates the need to grow the size of the Collector's circular buffer (via configuration `CacheCapacity`).
```
kentquirk, line 134: This isn't the otel Collector, and in fact let's drop the term entirely as it's now confusing.
```suggestion
- `collect_cache_buffer_overrun`: this should remain zero; a positive value indicates the need to grow the size of Refinery's circular trace buffer (via configuration `CacheCapacity`).
```

README.md
```diff
@@ -79,41 +95,44 @@ Refinery is a typical linux-style command line application, and supports several
...
+- `process_uptime_seconds`: records the uptime of each process; look for unexpected restarts as a key towards memory constraints.
```
kentquirk, line 135: Wondering if we should delete this reference, or at least add a note about StressRelief. Our guidance used to be "if it's crashing, give it more memory" but with StressRelief on it doesn't tend to crash.

README.md
```diff
@@ -127,26 +146,38 @@ Refinery validates its configuration on startup or when a configuration is reloa
 
 ### Configuration Query
 
-Because the configuration file formats (YAML) can sometimes be confusing to read and write, it may be valuable to check the loaded configuration by using one of the `/query` endpoints from the command line on a server that can access a refinery host.
+When using the standard configuration file formats, such as TOML and YAML, check the loaded configuration by using one of the `/query` endpoints from the command line on a server that can access a Refinery host.
```
kentquirk, line 149: I don't want to refer to TOML. I took it out deliberately. If someone wants TOML they can get it from the query endpoint, but let's not point them to it.

Honestly, this feature is now largely obsolete since we have validation.

README.md
```diff
@@ -156,20 +187,20 @@ Refinery does not yet buffer traces or sampling decisions to disk. When you rest
...
+`collect/collect.go` the Collector is responsible for bundling spans together into traces and deciding when to send them to Honeycomb or if they should be dropped. The first time a trace ID is seen, the Collector starts a timer. If the root span, which is a span with a trace ID and no parent ID, arrives before the timer expires, then the trace is considered complete. The trace is sent and the timer is canceled. If the timer expires before the root span arrives, the trace will be sent whether or not it is complete. Just before sending, the Collector asks the `sampler` for a sample rate and whether or not to keep the trace. The Collector obeys this sampling decision and records it (the record is applied to any spans that may come in as part of the trace after the decision has been made). After making the sampling decision, if the trace is to be kept, it is passed along to the `transmission` for actual sending.
```
kentquirk, line 196: I guess if you insist, but I don't want to overemphasize this term that dates from 5 years ago and is now confusing.

mjingle: I am suggesting formatting changes here. If you have issues with the long-existing content here, then this is an opportunity for a re-write  (or even a deletion) outside of this PR.

README.md
mjingle: Feel free to update with StressRelief!

README.md
mjingle:
```suggestion
Check the loaded configuration by using one of the `/query` endpoints from the command line on a server that can access a Refinery host.
```

README.md
mjingle: @kentquirk How about this instead?
```suggestion
Configuration is controlled by Refinery's two configuration files, which is generally referred to as `config.yaml` for general configuration and `rules.yaml` for sampling configuration.
```

README.md
mjingle: @kentquirk How about this then? It's the most recent version of the same content in the Honeycomb docs site [as seen here](https://docs.honeycomb.io/manage-data-volume/refinery/set-up/#system-requirements): 
> In many cases, Refinery only needs one node. If experiencing a large volume of traffic, you may need to scale out to multiple nodes, and likely need a small Redis instance to handle scaling.

mjingle: Refer to suggested change below.

README.md
```diff
@@ -12,17 +12,20 @@ For more readable information about recent changes, please see [RELEASE_NOTES.md
 
 ## Purpose
 
-Refinery is a tail-based sampling proxy that operates at the the level of an entire trace. It examines whole traces and intelligently applies sampling decisions (whether to keep or discard) to each trace.
+Refinery is a tail-based sampling proxy and operates at the level of an entire [trace](https://docs.honeycomb.io/concepts/tracing/). Refinery examines whole traces and intelligently applies sampling decisions to each trace. These decisions determine whether to include or discard the trace data in the sampled data forwarded to Honeycomb.
```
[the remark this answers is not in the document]
mjingle:
```suggestion
Refinery is a tail-based sampling proxy and operates at the level of an entire [trace](https://docs.honeycomb.io/concepts/tracing/). Refinery examines whole traces and intelligently applies sampling decisions to each trace. These decisions determine whether to keep or drop the trace data in the sampled data forwarded to Honeycomb.
```

README.md
kentquirk: Gonna save that for [2.3](https://github.com/honeycombio/refinery/issues/929)

kentquirk (approved):
Thank you for your help!

PR #859, opened 2023-09-15
feat: Add Redis Auth Authentication

### LINKED ISSUE #858: Support explicit specification of a redis AUTH string

Our Redis code was implemented several years ago, and in Redis 6 they implemented a command called AUTH, which GCP (at least) can use. Our client library supports it (see the [Dial example under Pool](https://pkg.go.dev/github.com/gomodule/redigo/redis#Pool)) but Refinery doesn't. 

We could add an Auth string to config -- it would also have to be added to cmdenv so that it could live in the environment or on the command line.

Not supporting it; our current use of redis only stores a list of internal and presumably non-routable IP addresses, so security really shouldn't be a concern.

But we will probably amp up our use of redis as we work on improving scalability.

**Additional context**

[GCP article about AUTH](https://cloud.google.com/memorystore/docs/redis/about-redis-auth#auth_behavior)
[Redis docs on AUTH](https://redis.io/commands/auth/)

tdarwin:

- #858 

This PR is to add the ability to use explicit AUTH strings when connecting to a Redis database.  This can be done using the `REFINERY_REDIS_AUTH_CODE` environment variable, or the following YAML:

```YAML
RedisPeerManagement:
  AuthCode: <AUTH PASSWORD STRING>
```

Fixes #858


kentquirk:
Will rebase this and approve after some dependency bumps merge.

kentquirk (approved):
I will likely tweak this a bit, but it's a real good start. Thank you!

PR #880, opened 2023-10-26
fix: ClearFrequencySec crash

### LINKED ISSUE #879: Specifying ClearFrequencySec in an input file crashes the `convert` tool
Converting rules (or helm) with a value specified for ClearFrequencySec crashes with `panic: interface conversion: interface {} is int, not int64`. The bug is in `tools/convert/ruleconvert.go` on line 40 -- we assume the value is read in as an int64, but it's an int. Use a conversion instead of an assertion.

kentquirk:

- Fixes #879 
- Adds tests


tools/convert/ruleconvert.go
```diff
@@ -37,17 +37,20 @@ func transformSamplerMap(m map[string]any) map[string]any {
 		switch k {
 		case "clearfrequencysec":
 			k = "clearfrequency"
-			v = config.Duration(v.(int64)) * config.Duration(time.Second)
+			switch vi := v.(type) {
+			case int:
```
VinozzZ, line 41: nit - I think we can do something like this. I don't know if there's a chance for this value to be types outside of int/int64. Should we add a default case to handle that?
```
case int, int32, int64:
   v = config.Duration(vi) * config.Duration(time.Second)
```

VinozzZ (approved):
LGTM. This definitely fixes the issue for crashing. I just left a nit for reducing code duplication and a question about an edge case

tools/convert/ruleconvert.go
fchikwekwe: I really like this suggestion, much easier to read.

tools/convert/ruleconvert_test.go
```diff
@@ -25,3 +28,44 @@ func Test_keysToLowercase(t *testing.T) {
...
 	}
 }
+
+func Test_ConvertTypes(t *testing.T) {
+	tests := []struct {
+		name string
+		m    map[string]any
+		want map[string]any
+	}{
+		{"clearfreq_int", map[string]any{
```
fchikwekwe, line 38: Don't we need test cases for `int32` as well?

fchikwekwe (approved):
One small nit, but this looks fine to me.

tools/convert/ruleconvert_test.go
kentquirk: As I understand it, it's not possible for YAML to represent an int32 in Go, so it's not necessary to test for it.

tools/convert/ruleconvert.go
VinozzZ: That's good to know! Thanks for explaining it

Issue #516, opened 2022-09-16
How should we handle linked traces?


cartermp:
Today, you can end up in an unfortunate scenario where:

* Trace A links to Trace B
* Trace A is sampled, but Trace B is not
* You click the link in the Honeycomb UI from Trace A to go to Trace B
* Trace B doesn't exist. WOMP WOMP

This came up in a discussion with a prospect who was just doing deterministic head sampling, and it makes sense that they'd see this with head sampling, but I realized we don't have any special handling of links in refinery either.

One thing I considered is:

* For each span with links on it, save the trace ID and always sample the trace
* For each trace with a span whose trace ID matches a saved trace ID from a link, always sample that trace

However, that could only work if the linked trace arrived after the trace that it's linked from. And OTel doesn't give you a way to tell if a given trace is being linked _from_ - you can only tell which trace is being linked _to_. So this can still result in a broken experience like described above.

Is there another way to approach this, or perhaps with some clever configuration?

kentquirk:
The problem with the solution you sketched out is a scenario like this:

Someone has a web request that shoves a message into Kafka for handling by a consumer. They turn the web request into a trace and end the trace when the request responds with the "message received".

Their consumer is given the trace ID of the initial request, but doesn't use it as a parent ID, because they don't need the entire thing as one big trace. When processing begins, they start a new trace and add a link to the original trace.

The result would be that every trace in the system gets sampled. 

One idea might be that we keep a longer-lived (maybe redis-based) record of traces sampled, and another record of traces referenced by links in sampled traces. 

The sampling decision for a trace then becomes:
* Is my trace ID in the list of referenced traces? If yes, then include me as well
* If I have links, have any of the linked IDs been included? If yes, then include me as well
* Otherwise, make a normal sampling decision

Does this sound plausible?

cartermp:
I think so, but couldn't this still result in the same scenario described, where a trace gets sampled but a trace that it links to isn't sampled? Or is the thinking that it would make that scenario less common?

kentquirk:
If both the linking and linked traces occur near each other in time, this would basically ensure that the sampling decisions are largely consistent between them. 

If you use a definitive cache, the only time they'd disagree is if one of them fell out of the cache; a central cache stored in Redis with a capacity of millions of trace IDs would mean that hours or days would have to pass before you could make an incorrect decision. For large installations, though, you might have a pretty big cache to get adequate storage.

[...]

PR #882, opened 2023-10-30
Draft: Add link-awareness

mterhar:

- https://github.com/honeycombio/refinery/issues/516

- Adds the ability for one refinery node to ask another one if it should keep a trace
- Adds the the decision-by-link from a root span back to a preceding trace
- Adds a configuration called `LinkStrategy` that can be enabled to allow link-awareness
- Caches it's answer in the cuckoo to limit additional api traffic

Things to wrap up:

- marshal the response into JSON to be more deliberate about how it's handled
- see how it handles links to links

Things it doesn't do: 

- Doesn't look at multiple links on the root span
- Doesn't look at additional spans farther down the trace. Just the root span linking back to any span on the other trace.
- Doesn't push a decision up if a the predecessor trace is still pending and the linked trace matches an error rule. 
- Doesn't lock on the API call so if a bunch of linked traces arrive at once, it'll make lots of API calls at the same time
- Doesn't allow the link to be considered among other things. Right now it's all-or-nothing. 
- Doesn't allow link handling to be done different ways depending on the environment
- Doesn't share the decisions into a shared storage (like the redis thingee)


bixu:
We've had folks at $job ask about this when we suggest they break up "megatraces" into smaller ones.

Issue #842, opened 2023-08-19
Structured: true does not generate json logs


KaoruDev:
Settings `Structured: true` does not generate JSON logs. I see the following:

```
time="2023-08-19T21:55:27Z" level=info msg="Sending trace" environment=dev reason=deterministic/always trace_id=9f15b4d3a2148cc63455e1d7fa638bb3
time="2023-08-19T21:55:51Z" level=info msg="Sending trace" environment=dev reason=deterministic/always trace_id=0362fb07db9239527ec430e8d6962098
time="2023-08-19T21:55:53Z" level=info msg="Sending trace" environment=dev reason=deterministic/always trace_id=31f573d8fcbbd81ce3847a4e14fbf812
time="2023-08-19T21:55:53Z" level=info msg="Sending trace" environment=dev reason=deterministic/always trace_id=2e737f6f82c154ba8f81f522eabd43da
time="2023-08-19T21:55:53Z" level=info msg="Sending trace" environment=dev reason=deterministic/always trace_id=2bd74614c0ba61e980c59eb517668360
```

**Versions**

Docker image: `honeycombio/refinery:2.1.0`

**Steps to reproduce**

1. Set the following config:
```yaml
Debugging:
  AdditionalErrorFields:
    - trace.span_id
    - name
  DryRun: false
Logger:
  Type: "stdout"
  Level: info
StdoutLogger:
  Structured: false
```

**Additional context**

kentquirk:
Wow. You're correct -- and it's sad, but it never has. This feature was added to config years ago but never used and you're apparently the first to notice.

Will fix in an upcoming release.

KaoruDev:
Thanks for the quick response and fast-follow up!

PR #884, opened 2023-10-31
fix: [config]: fix structured logging

VinozzZ:
#842
Currently, the structured logging is not working as expected when using stdout logger. 

This PR makes sure the `StdoutLogger.Structured` is applied to the logger implementation. If it's set to true, the logger will use JSON formatter. Otherwise, it's default to using plain text.

### DIFF
--- a/config.md
+++ b/config.md
@@ -1,7 +1,7 @@
 # Honeycomb Refinery Configuration Documentation
 
 This is the documentation for the configuration file for Honeycomb's Refinery.
-It was automatically generated on 2023-08-02 at 23:32:04 UTC.
+It was automatically generated on 2023-10-31 at 20:19:42 UTC.
 
 ## The Config file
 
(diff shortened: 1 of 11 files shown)


config.md
```diff
@@ -671,6 +670,16 @@ Many Redis installations do not use this field.
 - Type: `string`
 - Environment variable: `REFINERY_REDIS_PASSWORD`
 
+### `AuthCode`
```
VinozzZ, line 673: Hey @kentquirk , I'm not sure if this is just some missing doc that didn't generate before?

kentquirk: Yes; from a contributed PR -- I knew we were going to be regenerating docs several times before the release, so I didn't make them do it.

kentquirk (approved):
Congratulations! You seem to have found all the traps in the first round, and in just one day. I honestly have no complaints.

Issue #705, opened 2023-06-07
The ability to sample logrus logs or more robust controls on log messages


McSick:
**Is your feature request related to a problem? Please describe.**
Imagine getting an error sending an event and you send millions+ events. A log is generated for each failed send. The Honeycomb  Logger has a sampler which helps with this, but the logrus does not. 

**Describe the solution you'd like**
A sampler for logrus or finer control of log message output.

**Describe alternatives you've considered**
Use Honeycomb Logger but not everyone wants to do this.

**Additional context**

PR #892, opened 2023-11-01
feat: enable sampling for stdout logger

VinozzZ:

#705 

This PR implements the same sampling logic from HoneycombLogger in the StdoutLogger.

### DIFF
--- a/config/file_config.go
+++ b/config/file_config.go
@@ -276,6 +276,8 @@ type HoneycombLoggerConfig struct {
 
 type StdoutLoggerConfig struct {
 	Structured bool `yaml:"Structured" default:"false"`
+	SamplerEnabled	bool   `yaml:"SamplerEnabled" `
+	SamplerThroughput int    `yaml:"SamplerThroughput" default:"5"`
 }
 
 type PrometheusMetricsConfig struct {
--- a/config/file_config.go
+++ b/config/file_config.go
@@ -271,13 +271,13 @@ type HoneycombLoggerConfig struct {
 	APIKey            string `yaml:"APIKey" cmdenv:"HoneycombLoggerAPIKey,HoneycombAPIKey"`
 	Dataset           string `yaml:"Dataset" default:"Refinery Logs"`
 	SamplerEnabled    bool   `yaml:"SamplerEnabled" `
-	SamplerThroughput int    `yaml:"SamplerThroughput" default:"5"`
+	SamplerThroughput int    `yaml:"SamplerThroughput" default:"10"`
 }
 
 type StdoutLoggerConfig struct {
 	Structured bool `yaml:"Structured" default:"false"`
 	SamplerEnabled	bool   `yaml:"SamplerEnabled" `
-	SamplerThroughput int    `yaml:"SamplerThroughput" default:"5"`
+	SamplerThroughput int    `yaml:"SamplerThroughput" default:"10"`
 }
 
 type PrometheusMetricsConfig struct {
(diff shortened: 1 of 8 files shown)


kentquirk (commented):
Sorry, I didn't save this yesterday.

logger/logrus.go
```diff
@@ -98,24 +122,35 @@ func (l *LogrusEntry) WithField(key string, value interface{}) Entry {
...
 	}
 }
 
 func (l *LogrusEntry) Logf(f string, args ...interface{}) {
+	if l.sampler != nil {
+		rate := l.sampler.GetSampleRate(fmt.Sprintf("%s:%s", l.level, fmt.Sprintf(f, args...)))
+		if shouldDrop(uint(rate)){
+			return 
+		}
+		l.entry.WithField("sample_rate", rate)
```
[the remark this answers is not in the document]
VinozzZ: Oh, I'm not familiar with Beelines. Do you mind to explain a bit more on what the incompatibility looks like here?

config/metadata/configMeta.yaml
```diff
@@ -484,6 +484,27 @@ groups:
...
+        type: float
+        default: 10
+        example: 10
+        reload: false
+        summary: is the sampling throughput for logs in events per second.
+        description: >
+          The sampling algorithm attempts to make sure that the average
+          throughput approximates this value, while also ensuring that all
+          unique logs arrive at stdout at least once per sampling
+          period.
```
fchikwekwe, line 507: Is there a reason for using a ">" and the newline before this description?

config/file_config.go
```diff
@@ -276,6 +276,8 @@ type HoneycombLoggerConfig struct {
 
 type StdoutLoggerConfig struct {
 	Structured bool `yaml:"Structured" default:"false"`
+	SamplerEnabled	bool   `yaml:"SamplerEnabled" `
+	SamplerThroughput int    `yaml:"SamplerThroughput" default:"5"`
```
fchikwekwe, line 280: This `SamplerThroughput` default value conflicts with the default set in `configMeta.yaml`.

fchikwekwe: It also conflicts with the stated default value in the generated documentation.

fchikwekwe (commented):
Overall, this looks fine. Make sure the default values are consistent.

config/file_config.go
VinozzZ: Good find! I honestly don't really understand the relationship here. I got the default from line 274, which is the config for `HoneycombLogger`. It looks like that one also has the same issue. 
@kentquirk , do you mind to explain a bit more on the relationship between each default values?

config/metadata/configMeta.yaml
fchikwekwe: Oh I see, this formatting is consistent for this document. Please disregard.

config/file_config.go
fchikwekwe: Awesome! Then everything looks good to me.

fchikwekwe (approved):

logger/logrus.go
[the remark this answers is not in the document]
kentquirk: Whenever Honeycomb receives data in a column called SampleRate, we use that in calculations as the sample rate. It's how we compensate for sampling. So we should name the field the way Honeycomb expects to receive it, in case that data is ever forwarded to us.

config/metadata/configMeta.yaml
kentquirk: That's a YAML feature that means this is the start of a string that is a paragraph. A line with a smaller indent ends it.

config/file_config.go
kentquirk: Please see #894 -- turns out there were a few of these.

PR #893, opened 2023-11-01
fix: change sample key to not include args in honeycomb logger

VinozzZ:

Previously, Honeycomb logger uses the entire log message as the sampling key. If a log message contains high-cardinality fields, the throttle mechanism will no longer work.

This PR changes the key used for sampling from the entire log message to only use the format string

### DIFF
--- a/logger/honeycomb.go
+++ b/logger/honeycomb.go
@@ -237,7 +237,9 @@ func (h *HoneycombEntry) Logf(f string, args ...interface{}) {
 		level = "unknown"
 	}
 	if h.sampler != nil {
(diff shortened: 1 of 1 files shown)


kentquirk (approved):
LGTM!

(Also, I'm not sure "pervasively" is the word you want there. Did you mean "perversely"?)

fchikwekwe:
I actually think the word "pervasively" makes more sense in this context.

fchikwekwe (approved):

VinozzZ:
Hahaha, autocorrect definitely gave it more thought than me. My intention is `previously` lol

PR #894, opened 2023-11-02
fix: Correct defaults that got messed up in 2.x.

### LINKED ISSUE #259: Default OTLP port conflicts with Prometheus
The OTLP export endpoint and Prometheus metrics endpoint both try to bind to port 9090 by default.

OTel collector switched from default of 9090 to 4317, likely because of this conflict.

### LINKED ISSUE #843: Logger does not send logs to stdout even tho Logger.Type = stdout
Refinery still writes to `stderr` even tho `Logger.Type: stdout`. Below is a screenshot of the logs captured via GCP's Cloud Logging:

Logs_Explorer_–_Logging_–_US_Family_Dev_–_Google_Cloud_console

This is problematic because logs that are not errors show up as errors because they're being written to `stderr`

**Versions**

Docker image: `honeycombio/refinery:2.1.0`

**Steps to reproduce**

1. Set the following config:
```yaml
Debugging:
  AdditionalErrorFields:
    - trace.span_id
    - name
  DryRun: false
Logger:
  Type: "stdout"
  Level: info
StdoutLogger:
  Structured: false
```

**Additional context**

### LINKED ISSUE #874: PeerBufferSize default adjustment
PeerBufferSize is documented as both 10K (in [these docs](https://docs.honeycomb.io/manage-data-volume/refinery/scale-and-troubleshoot/#sizing-the-send-buffers)) and 100K (in the metadata / generated docs). The current value in the source is 10K but it was supposed to be bumped to 100K for 2.x. 

Fix it in the [default setting](https://github.com/honeycombio/refinery/blob/52beb1097f7525d97d2d5b6bc84da1a0176ec3fc/config/file_config.go#L331) and also fix the documentation.

### LINKED ISSUE #877: AddHostMetadataToTrace is documented as defaulting to true, but in fact is defaulting to false
AddHostMetadataToTrace is documented as defaulting to true, but in fact is defaulting to false.

We should fix this so that the default value is true; most people want this and it has no real negative impact.

kentquirk:

- Several default values in the config file (and therefore the documentation) differed from their values in the structure definitions. This is related to some of the changes that were made in 2.x -- some defaults were changed but the documentation didn't get changed in the same way. 
- This does NOT regenerate all docs as we expect to have to do that several more times in the coming days anyway.

This list will get revised before merging:
closes #877 
closes #874 
closes #843 
closes #259


VinozzZ:
Hey @kentquirk , it looks like some tests need to be modified for the new default value

config/metadata/configMeta.yaml
```diff
@@ -1211,7 +1213,7 @@ groups:
         v1name: MaxConnectionIdle
         type: duration
         valuetype: nondefault
-        default: 0s
```
VinozzZ, line 1214: I did a global search in the codebase. It looks like we also need to update the one in `cofig_complete.yaml` for this value

config/metadata/configMeta.yaml
```diff
@@ -1249,7 +1251,7 @@ groups:
         v1name: MaxConnectionAgeGrace
         type: duration
         valuetype: nondefault
-        default: 60s
```
VinozzZ, line 1252: same here

config/metadata/configMeta.yaml
```diff
@@ -1436,7 +1438,7 @@ groups:
         v1name: DeactivationLevel
         type: percentage
         valuetype: nondefault
-        default: 70
```
VinozzZ, line 1439: same here

VinozzZ (approved):

Issue #847, opened 2023-08-24
Please include the refinery version in telemetry


maplebed:
**Is your feature request related to a problem? Please describe.**
We were recently in the process of upgrading refinery and I wanted to confirm which version was running. I hoped I could find that information in telemetry to really know what's running (rather than looking in infra code which says which version is _supposed_ to be running). 

**Describe the solution you'd like**
I would like the refinery version included as an attribute in both the `refinery_metrics` and `refinery_logs` datasets. 

**Describe alternatives you've considered**
We could use the `AdditionalAttributes` config setting to manually record the version we are expecting to run in the config. While this would work it also introduces an easy opportunity for misleading information when doing an upgrade if the person upgrading forgets to also change the version in sync with the new deploy. 

**Additional context**
Especially when doing performance or usage comparisons, having the release version ID available makes it very easy to see the differences. "Does refinery 2.1.0 have the same memory footprint as 2.0.2?" is a question that requires extra work to answer when the version is not included in the memory metrics for an easy "group-by".

tdarwin:
@maplebed It's in the OTEL Metrics as service.name (refinery) and service.version (2.1.0).  However it's not in the Legacy Metrics. That doesn't help much, given that we're not using the OTEL metrics...  But it is somewhere!

PR #899, opened 2023-11-13
fix: add refinery version to refinery metric and log

VinozzZ:

- #847

This PR adds `refinery_version` to both legacy metrics and honeycomb logger.

Example of the newly added field:

Screenshot 2023-11-13 at 2 15 37 PM
Screenshot 2023-11-13 at 2 15 13 PM


kentquirk (approved):
LGTM

Issue #873, opened 2023-10-12
Refinery service file needs updated to point at yaml config and rules files


elainemccloskey:
**Versions**

- Refinery: v2.1.0

**Steps to reproduce**

https://github.com/honeycombio/refinery/blob/v2.1.0/refinery.service

Start the refinery service using systemd, it attempts to load the old toml files for config and rules rather than the new default yaml files.

MichielBijland:
We also ran into this bug as well

PR #900, opened 2023-11-13
fix: load default config and rules file in service file

VinozzZ:

- #873

This PR removes the old config and rules argument from the service file for systemd. This should make sure that the default config and rules files are correctly loaded


kentquirk (approved):
👍

PR #908, opened 2023-11-29
maint: update dependency for x/exp

kentquirk:

- golang.org/x/exp updated with breaking changes; this adapts to those

- Update dep
- Fix sorting in main()


fchikwekwe (approved):
LGTM

PR #910, opened 2023-11-29
feat(config): add AddCountsToRoot to report counts data for traces

### LINKED ISSUE #555: "AddSpanCountToRoot" field reports wrong span count due to including span annotations
# Versions

- **Go:** 1.19.3
- **Refinery:** [1.18.0](https://github.com/honeycombio/refinery/releases/tag/v1.18.0)

# Steps to reproduce

1. In the Refinery configuration, set the new "AddSpanCountToRoot" field (as of #532) to true.
1. Publish several traces that flow through Refinery to Honeycomb, including at least one span event on some of the trace spans.
1. Inspect the root span of one of these traces in a Honeycomb query.
1. Observe that the "meta.span_count" field is present on the root span and populated with a positive integer.
1. Observe that the "meta.span_count" field's count is greater than the number of spans in the trace, and is instead the sum of the trace count and the span event count.

# Additional context

[I reported this problem first in the "discuss-refinery" channel](https://honeycombpollinators.slack.com/archives/C01BZ3J7RMJ/p1667833743521379) of the "Honeycomb Pollinators" Slack workspace.

In [file _types/events.go_'s `(*types.Trace).SpanCount` method](https://github.com/honeycombio/refinery/blob/a72e763b59fb5eccbab9235f16d487c39dcd586f/types/event.go#L71-L74), the current implementation reports the length of the "spans" field's slice:
```go
// SpanCount gets the number of spans currently in this trace as int64
func (t *Trace) SpanCount() int64 {
	return int64(len(t.spans))
}
```
A slower but more accurate way to report this count is as follows:
```go
// SpanCount reports the count of spans currently in this trace.
func (t *Trace) SpanCount() int64 {
	var count int64
	for _, s := range t.spans {
		if annoType, ok := s.data["meta.annotation_type"]; !ok || annoType != "span_event" {
			count++
		}
	}
	return count
}
```
Alternately, if you don't mind using the zero value returned for a key that's not present in the map:
```go
// SpanCount reports the count of spans currently in this trace.
func (t *Trace) SpanCount() int64 {
	var count int64
	for _, s := range t.spans {
		if s.data["meta.annotation_type"] != "span_event" {
			count++
		}
	}
	return count
}
```
Are there other annotation types that we need to omit? Perhaps "link?"

VinozzZ:

- #555 

This PR introduce a new config value `AddCountsToRoot`. If `true`, then Refinery will ignore the `AddSpanCountToRoot` setting and add the following fields to the root span based on the values at the time the sampling decision was made: 
- `meta.span_count`: the number of child spans on the trace 
- `meta.span_event_count`: the number of span events on the trace 
- `meta.span_link_count`: the number of span links on the trace 
- `meta.event_count`: the number of honeycomb events on the trace

fix #555


kentquirk (commented):
I'm nervous about 2 things:
* We're increasing the size of the kept traces cache entry from 128 bytes (2 uint) to 320 bytes (5 uint). I feel like I would like to convert all of these values to uint32, which would mean that we'd only have 160 bytes per entry.
* I don't like adding all those additional int-typed parameters to the dealWithSentTrace call. Maybe the best bet is to rename `cuckooKeptRecord` to `KeptTraceCacheEntry` and then pass a pointer to it as a single value.

What do you think?

VinozzZ:
> * KeptTraceCacheEntry



> I'm nervous about 2 things:
> 
> * We're increasing the size of the kept traces cache entry from 128 bytes (2 uint) to 320 bytes (5 uint). I feel like I would like to convert all of these values to uint32, which would mean that we'd only have 160 bytes per entry.
> * I don't like adding all those additional int-typed parameters to the dealWithSentTrace call. Maybe the best bet is to rename `cuckooKeptRecord` to `KeptTraceCacheEntry` and then pass a pointer to it as a single value.
> 
> What do you think?

I like the suggestion! I made the changes you suggested. There's only one thing that I was hesitant to do. I see that the `SampleRate` is being transmitted through the wire as a `uint`. I don't know if truncate it to `uint32` would cause any weird behavior. It's probably fine since no one probably goes beyond the limit of uint32. I just want to call it out before I make the change

kentquirk:
The largest value I've ever seen for Sample rate is 1_000_000_000, which was used in a test scenario. Realistically, not even our biggest customers would use values even as large as 1_000_000. I think we're pretty safe in limiting it to 4 billion.

collect/cache/traceSentCache.go
```diff
@@ -11,7 +11,13 @@ type TraceSentRecord interface {
 	// Rate() returns the sample rate for the trace
 	Rate() uint
 	// DescendantCount returns the count of items associated with the trace, including all types of children like span links and span events.
-	DescendantCount() uint
+	DescendantCount() uint32
```
kentquirk, line 14: Can we plese make the int32 an implementation detail of the record, and have all these functions return a normal int or an int64?

kentquirk (approved):
Looks great!

PR #914, opened 2023-11-29
fix: Allow non-ints in memorysize

### LINKED ISSUE #904: Collection.AvailableMemory breaking when setting decimal values (recognize as bytes)
When we set the value of `Collection.AvailableMemory` in a decimal format (with a `.`, it does not calculate the Memory correctly. For example, when I set `Collection.AvailableMemory` as `2Gi` it works as expected, but when I set it as `2.0Gi`, it breaks totally as it converts to 2bytes.

I tweaked the test [TestAvailableMemoryCmdLine](https://github.com/honeycombio/refinery/blob/24a2add48e3d989e894001a8c8ca87abd706a65f/config/config_test.go#L455-L455) to test it. this is the version that works
```go
func TestAvailableMemoryCmdLine(t *testing.T) {
	cm := makeYAML("General.ConfigurationVersion", 2, "Collection.CacheCapacity", 1000, "Collection.AvailableMemory", "2Gi")
	rm := makeYAML("ConfigVersion", 2)
	config, rules := createTempConfigs(t, cm, rm)
	defer os.Remove(rules)
	defer os.Remove(config)
	c, err := getConfig([]string{"--no-validate", "--config", config, "--rules_config", rules})
	assert.NoError(t, err)

	expected := MemorySize(2 * 1024 * 1024 * 1024)
	inMemConfig, err := c.GetCollectionConfig()
	assert.NoError(t, err)
	assert.Equal(t, expected, inMemConfig.AvailableMemory)
}
```
It runs very nicely, as seen by the test output.
image

but, just changing the value of `Collection.AvailableMemory` to  `2.0Gi`, breaks the test. 

```go
func TestAvailableMemoryCmdLine(t *testing.T) {
	cm := makeYAML("General.ConfigurationVersion", 2, "Collection.CacheCapacity", 1000, "Collection.AvailableMemory", "2.0Gi")
	rm := makeYAML("ConfigVersion", 2)
	config, rules := createTempConfigs(t, cm, rm)
	defer os.Remove(rules)
	defer os.Remove(config)
	c, err := getConfig([]string{"--no-validate", "--config", config, "--rules_config", rules})
	assert.NoError(t, err)

	expected := MemorySize(2 * 1024 * 1024 * 1024)
	inMemConfig, err := c.GetCollectionConfig()
	assert.NoError(t, err)
	assert.Equal(t, expected, inMemConfig.AvailableMemory)
}
```
see the test result.

image

I am not an expert in k8s, but for me both `2Gi` and `2.0Gi` should give me the same output. I discovered this the hard way as I deployed it and start showing memory errors on the output.

If I change `2.0Gi` to `2048Mi` with works, but If I introduce a dot. it does not work `"2048.0Mi"`

**Versions**

- Go: 1.20.1
- Refinery:2.1.0

**Steps to reproduce**
Just copy the test above.

**Additional context**

kentquirk:

- #904 -- Specifying memory size as 2.5Gib didn't work. 

- Modify UnmarshalText for memorysize to accept floating point values
- Write lots of tests to make sure it works

Fixes #904


config/memorysize.go
```diff
@@ -0,0 +1,161 @@
...
+
+	// No unit means bytes
+	if unit == "" {
+		*m = MemorySize(number)
+	} else {
+		scalar, ok := unitMap[unit]
+		if !ok {
+			return fmt.Errorf(invalidSizeError, text)
+		}
+		*m = MemorySize(number * float64(scalar))
```
TylerHelmuth, line 151: I'm not sure converting back to a `unit64` in this way leads to expected behavior from a user's perspective.  If I'm allowed to configure `1.5Gi` I would expect to end up with roughly `1500Mi` not `1Gi`. If we end up dropping the decimal precision I'd prefer to enforce int in the input and not allow floats.

TylerHelmuth: Although looking at the unit tests I see that we are preserving the precision and I see now its because we're multiplying by the scalar

TylerHelmuth (approved):

PR #915, opened 2023-11-29
chore: Add some detail about release tasks

kentquirk:
This just adds some text to the RELEASING file so we don't forget these tasks.


VinozzZ (approved):
Thank you for updating the doc!

Issue #826, opened 2023-08-02
Allow incoming and peers to have separate sizes in config


kentquirk:
Right now both incoming and fromPeer (collect.go, line 120) use the CacheCapacity*3 for their sizes.

These values should be separately configurable. In large clusters the fromPeer should usually be quite a bit larger than incoming, since only 1/N of the traffic will come from incoming, while (N-1)/N will come from peers.

PR #916, opened 2023-11-30
feat(config): allow separate config for peer and incoming span queue

VinozzZ:

- #826 

This PR adds two new configuration option, `Collection.PeerQueueSize` and `Collection.IncomingQueueSize`. This enables users to have more control over each span queue size based on their refinery setup.

Before this change, both queue sizes are defined as 3*`Collection.CacheCapacity`. Since the queue size should always be larger than the trace cache, I decided to set the minimum value for both queue size to be 3*`Collection.CacheCapacity`. I'm not sure if there's a use case to have a smaller `PeerQueueSize` if refinery is not running in cluster. I would love to get some feedback on that.


config/file_config.go
```diff
@@ -172,18 +172,42 @@ type RedisPeerManagementConfig struct {
...
 		return c.MaxAlloc
 	}
 	return c.AvailableMemory * MemorySize(c.MaxMemoryPercentage) / 100
 }
 
+// GetPeerBufferCapacity returns the capacity of the in-memory channel for peer traces.
+// If PeerBufferCapacity is not set, it uses 3x the cache capacity.
+// The minimum value is 3x the cache capacity.
+func (c CollectionConfig) GetPeerQueueSize() int {
+	if c.PeerQueueSize == 0 || c.PeerQueueSize < c.CacheCapacity*3 {
```
VinozzZ, line 195: Question: Before this change, both queue sizes are defined as 3*Collection.CacheCapacity. Since the queue size should always be larger than the trace cache, I decided to set the minimum value for both queue size to be 3*Collection.CacheCapacity. I'm not sure if there's a use case to have a smaller PeerQueueSize if refinery is not running in cluster. I would love to get some feedback on that.

PR #917, opened 2023-11-30
feat: Add GRPC configuration option, rework grpc config a bit

### LINKED ISSUE #905: Can we have configuration to define GRPCMessageSizeMax?

We receive traces that are consistently over 5 MB in GRPC packets. Are there any limitations to allowing this configuration? We face significant trace loss due to a context where nearly 30% of the 5 billion daily traces have large label values resulting in direct gRPC packets that are nearly 20 MB in size.

[I would like to have gRPC max size configuration inside refinery configs.](https://github.com/honeycombio/refinery/blob/main/route/route.go#L52)

**Additional context**

Example of error and size of gRPC packet that we have. In the example below we have 10mb.
```txt
Permanent error: rpc error: code = ResourceExhausted desc = grpc: received message after decompression larger than max (10389172 vs. 5000000)
```

kentquirk:

 - Fixes #905 

- Adds two new GRPC config values that control the max size of send and receive blocks
- GRPC had a lot of helper functions; this eliminates most of them in favor of returning a config block
- Documents the new parameters
- Fixes a validation bug for memorysize related issues


config/metadata/configMeta.yaml
```diff
@@ -1330,6 +1330,40 @@ groups:
           activity, then it pings the client to see if the transport is still
           alive. "0s" sets duration to 20 seconds.
 
+      - name: MaxSendMsgSize
+        type: memorysize
+        valuetype: memorysize
+        default: 6MiB
+        reload: false
+        firstversion: v2.2
```
VinozzZ, line 1338: Oh, this is the first time I noticed we have this field for a config. Does this mean all new configs since last release should have this set?

kentquirk: Yes, technically. It's not currently used for anything, but it could be helpful both for validation and documentation. It might be worth an effort to go back to add those notations to new things that came along in 2.0 and later.

PR #919, opened 2023-11-30
feat(config): expose IdleTimeout for http.Server

### LINKED ISSUE #825: Expose http.Server.IdleTimeout via configuration

Some hosting options for Refinery offer limited configurability of the HTTP keepalive timeouts used by load balancers in front of Refinery. The example I'm most familiar with are GCP load balancers, which have fixed and non-configurable keepalive timeouts of ~10 minutes.

In this situation, it's crucial that upstream/backend services have a _longer_ idle timeout than their clients (the load balancers), otherwise there will always be situations where a load balancer attempts to use a connection that the backend has already discarded.

The impact to clients communicating with Refinery through such a load balancer is typically a low but consistent level of connection reset errors.

Would it be possible to expose configuration of `http.Server.IdleTimeout` for the [main Refinery HTTP listener](https://github.com/honeycombio/refinery/blob/00fe2b31667c1a3f7b68802ecc637d8ad5eb1ed5/route/route.go#L206) via configuration?

Technically I'm sure there are other solutions here, including asking operators of Refinery to deploy a different service (nginx/etc.) between the downstream load balancer and Refinery, but there don't seem to me to be compelling reasons _not_ to expose this knob in Refinery's configuration.

**Additional context**

- https://cloud.google.com/load-balancing/docs/https/#timeout-keepalive-backends
- https://cloud.google.com/load-balancing/docs/https/#timeout-keepalive-clients

VinozzZ:

- #825 

This PR adds a new config option `HTTPIdleTimeout` for users to control refinery's http server's idle timeout setting.

I also found and fixed an issue in the `minOrZero` validation logic that prevents users to actually set a field as 0

fix #825

### DIFF
--- a/config/validate.go
+++ b/config/validate.go
@@ -301,7 +301,7 @@ func (m *Metadata) Validate(data map[string]any) []string {
 					errors = append(errors, msg)
 				} else {
 					fm := mustFloat(validation.Arg)
-					if fv == 0 || fv < fm {
+					if fv != 0 && fv < fm {
 						errors = append(errors, fmt.Sprintf("field %s must be at least %v, or zero", k, validation.Arg))
 					}
 				}
--- a/config.md
+++ b/config.md
@@ -1,7 +1,7 @@
 # Honeycomb Refinery Configuration Documentation
 
 This is the documentation for the configuration file for Honeycomb's Refinery.
-It was automatically generated on 2023-11-30 at 19:50:23 UTC.
+It was automatically generated on 2023-11-30 at 22:36:51 UTC.
 
 ## The Config file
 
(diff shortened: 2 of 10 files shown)


config/validate.go
```diff
@@ -301,7 +301,7 @@ func (m *Metadata) Validate(data map[string]any) []string {
 					errors = append(errors, msg)
 				} else {
 					fm := mustFloat(validation.Arg)
-					if fv == 0 || fv < fm {
+					if fv != 0 && fv < fm {
```
VinozzZ, line 304: Here's the fix for allowing `minOrZero` validation to pass when a zero value is passed in

kentquirk (approved):
Good catch.

I had said before that I was trying NOT to include regenerating all the docs in every PR. I'd generally like to follow that, please, because besides making these PRs bigger and harder to read, it also leads to conflicts on every rebase. 

Don't undo it for this, but for future ones let's try to follow that and just regenerate them when we're about to release.

VinozzZ:
> Good catch.
> 
> I had said before that I was trying NOT to include regenerating all the docs in every PR. I'd generally like to follow that, please, because besides making these PRs bigger and harder to read, it also leads to conflicts on every rebase.
> 
> Don't undo it for this, but for future ones let's try to follow that and just regenerate them when we're about to release.

Yes! Thanks for the reminder. I keep forgetting lol

Issue #883, opened 2023-10-31
Create a local development guide


VinozzZ:
When a new contributor is getting started on the project, it would be helpful to have a document that goes through development workflow for Refinery.
Some of the topics that this document can cover:
- How to run Refinery locally
- How to run tests
- How to send test traffic to a local instance of Refinery

mterhar:
The setup below is what I used to build and test link-awareness. It requires multiple refinery nodes to be running and some data flowing through them.

`docker-compose.yml` file will build and run 2 refineries and a redis

```yaml
services:
  refinery1:
    build: ../
    ports:
      - 4317:4317
      - 8080:8080
    entrypoint:
      - "refinery"
    command:
      - "-c"
      - "/etc/refinery/config.yaml"
      - "-r"
      - "/etc/refinery/rules.yaml"
    volumes:
      - ./config.yaml:/etc/refinery/config.yaml
      - ./rules.yaml:/etc/refinery/rules.yaml
  refinery2:
    build: ../
    ports:
      - 4817:4317
      - 8880:8080
    entrypoint:
      - "refinery"
    command:
      - "-c"
      - "/etc/refinery/config.yaml"
      - "-r"
      - "/etc/refinery/rules.yaml"
    volumes:
      - ./config.yaml:/etc/refinery/config.yaml
      - ./rules.yaml:/etc/refinery/rules.yaml
  redis:
    image: redis:6.2.5
    expose:
      - "6379"
```

Using the [otel-go-batch](https://github.com/mterhar/otel-go-batch/tree/per-job-async) example, we'll create a bunch of
traffic that has span links.

### Config.yaml

```yaml
General:
  ConfigurationVersion: 2
  MinRefineryVersion: v2.0
Network:
  ListenAddr: "0.0.0.0:8080"
  PeerListenAddr: "0.0.0.0:8081"
RefineryTelemetry:
  AddRuleReasonToTrace: true
  AddSpanCountToRoot: true
  AddHostMetadataToTrace: true
Traces:
  LinkStrategy: RootLinkOverride
IDFields:
  TraceNames: [ trace.trace_id ]
  ParentNames: [ trace.parent_id ]
  LinkNames: [ trace.link.trace_id ]
Debugging:
  QueryAuthToken: sneaky-longer-token
  AdditionalErrorFields:
    - trace.span_id
  DryRun: true
Logger:
  Type: honeycomb
  Level: debug
HoneycombLogger:
  APIKey: $REFINERY_APIKEY
[...]

PR #932, opened 2023-12-05
maint: update doc for release process and config/rules doc generation process

VinozzZ:

- #883 

This PR adds more details on the release process and config/rules convert tool workflow.
It also fixes a mistake in the `CONTRIBUTING.md` doc regarding the config/rules doc update process

### DIFF
--- a/tools/convert/README.md
+++ b/tools/convert/README.md
@@ -22,9 +22,9 @@ Usage:
 	For config files, the new v2 config file is commented in detail to help explain what each
 	value does in the new configuration.
 
-	For example, if the v1 file specifield has the same value as the default value, it will be commented out in the output file. Only non-default
-	value is not commented out.d "MetricsAPIKey" in the "HoneycombMetrics" section, the v2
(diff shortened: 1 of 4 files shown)


tools/convert/README.md
```diff
@@ -22,7 +22,8 @@ Usage:
 	For config files, the new v2 config file is commented in detail to help explain what each
 	value does in the new configuration.
 
-	For example, if the v1 file specified "MetricsAPIKey" in the "HoneycombMetrics" section, the v2
+	For example, if the v1 file specifield has the same value as the default value, it will be commented out in the output file. Only non-default
+	value is not commented out.d "MetricsAPIKey" in the "HoneycombMetrics" section, the v2
```
kentquirk, line 26:
```suggestion
	Values that are the same as default values are commented out in the generated output file (since the default values are correct). Only values that do not match the default are enabled in the output file.
```

kentquirk (approved):
A few little grammar fixes, but generally looks good. Thank you!

PR #933, opened 2023-12-05
fix: handle otlp request with /v1/traces/ path

### LINKED ISSUE #867: Ignore trailing slash on /v1/traces endpoint

I just spent a couple of hours trying to figure out why refinery is not applying the sampling to the traces I sent and instead is forwarding all traces directly to honeycomb. As it turns out, I've set the endpoint in my application to `<refineryhost>:8080/v1/traces/`. Once I removed the trailing slash, everything started to work as expected.

Refinery should work like honeycomb itself. When sending traces to `/v1/traces/` it should ignore the trailing slash and process the trace as expected.

I guess an alternative could be to reject the request and respond with the same error that appears when accidentally sending to the root endpoint: 

```
invalid OTLP endpoint - should you be sending to /v1/traces?, see API docs https://docs.honeycomb.io/
```

_Side-Note: The question mark directly behind a URL is a little confusing. Maybe consider adding a space between `/v1/traces` and `?`_

**Additional context**

Of course in hindsight it's all very obvious, but I can easily see myself making that error again in the future.

VinozzZ:

- #867 

Since otlp spec allows Non-default URL paths for requests to be configured on the client and server sides, this PR adds support in Refinery to handle `/v1/traces/` in addition to the default `/v1/traces`

fix #867

### DIFF
--- a/route/route.go
+++ b/route/route.go
@@ -172,11 +172,7 @@ func (r *Router) LnS(incomingOrPeer string) {
 	authedMuxxer.HandleFunc("/batch/{datasetName}", r.batch).Name("batch")
 
 	// require an auth header for OTLP requests
-	otlpMuxxer := muxxer.PathPrefix("/v1/").Methods("POST").Subrouter()
-	otlpMuxxer.Use(r.apiKeyChecker)
-
-	// handle OTLP trace requests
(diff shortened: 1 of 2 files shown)


kentquirk (approved):
👍

PR #934, opened 2023-12-06
fix: try to deflake several flaky tests

kentquirk:
## Which problem is this PR solving?

We've had some problematic flaky tests. 
* Some were because of probabalistic assertions relating to samplers -- we now run the worst of those a second time if they fail.
* Some were waiting for events on another goroutine that might not have been scheduled, and these have been adjusted by using `assert.Eventually` which is actually pretty useful.

Fixes #896
Fixes #897
Fixes #901
Fixes #902

Plus a couple of other tests that hadn't gotten their own issue.

PR #935, opened 2023-12-06
fix: Integration tests fail in parallel

kentquirk:

- After merging #934, integration tests started failing because they can't safely run in parallel. 

- Remove the parallelization markers.


kentquirk:
You've just made me realize that the problem is the port conflict! I've made better changes.

fchikwekwe (approved):
Oh, clever fix! Nice to have parallel back.

PR #936, opened 2023-12-06
feat: add sent reason for late arriving spans

### LINKED ISSUE #835: Please provide a more descriptive reason for late-arriving spans
When customers first encounter the reason `late`, they're a bit confused and have to ask about it.  I was thinking a more descriptive reason would make it more intuitive.

If it's possible to know/store the rule name without a significant performance penalty, then perhaps something like `rules/trace/rulename - late arriving span` and could also aid debugging.   If not, perhaps changing the text to `late arriving span` would do enough to quell the confusion.

Documentation, although it might be difficult to make the keyword `late` searchable enough.

**Additional context**
Marking late-arriving spans is a super cool feature and has been a huge quality-of-life improvement!  Once people know what it means they totally get it, it's more of a new user issue.

VinozzZ:

- #835 

This PR adds a new cache for send_reasons that Refinery produces for a given configuration.
The new cache acts as a mapping between the string representation of send reason and a uint.
The cache is used to reduce the memory footprint of the trace cache.
On every config reload, the cache is cleared out since that usually mean a new rule or a new sampler is used.

I made the design based on the assumption that the send reason has a low cardinality for a given configuration. I'm not entirely sure if that assumption is correct. 

fix #835


kentquirk (approved):
I just want to note that we went over this on a call and the benchmarks showed that this cache design is extremely fast, since it's lockless on read. We also tested for race conditions in parallel operation.

Issue #822, opened 2023-08-01
Create a `match` and/or `in` (also `not-in`) operator


kentquirk:
**Is your feature request related to a problem? Please describe.**

Sometimes it would be nice to be able to test a field for multiple values or a more complex form of containment, since the rules language doesn't support an `or` semantic.

**Describe the solution you'd like**

One option is an operator `in`, where the Value field is a list of possible values. If the specified field matches any of the values then it would succeed. This requires allowing the value field to be a list and adding code to support it when reading configuration files. It could support lists of any datatype -- for example, a list of numeric error codes.

This is easy to document and validate, but limited in that it only supports a list of exact matches.

An alternative is an operator `match`, which would accept for its Value field a Go-style regular expression. During parse time the Matches function would be constructed to pre-parse the regexp so that rule evaluation is just calling the match function on it.  We would also have to make the validation code check that it could be parsed. All columns would be coerced to strings before matching (it would have an implicit Datatype of "string").

This is a bit harder to explain to non-programmers, but powerful in that it can match a list like 'in' would use, but could also match more complex patterns. 

**Context**
See original request in [Pollinators slack](https://honeycombpollinators.slack.com/archives/C01BZ3J7RMJ/p1690925558226369).

kamalmarhubi:
Original requestor here :-)

For string types, regexp is more general for sure. But I think, even for programmers, writing the `^(option|another_option|a_third_option)$` regexps gets pretty ugly / unwieldy once you're past about 3–5 options—my current rule has eight. Under the hood, the `in` operator could be desugared to regexp since that'd probably have better performance.

---
Aside: I had secretly been hoping that since you and @TylerHelmuth [are owners of `pkg/ottl`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/88c3006f7d3ce14ad9a887e98cd9257296a5c517/.github/CODEOWNERS#L128) that refinery might learn to use that for rules. But since v2 came without that, I figure that's not actually on the roadmap 🙈. (I realise it's difficult to make that fit with the default `Scope: trace` semantic for rules!)

TylerHelmuth:
OTTL in Refinery might happen some day, but it isn't a top priority at the moment.  We'd have to build our own contexts and the framework very much wants you to be using the collector's underlying data structure to be otlp but we have honeycomb events.

We are still very much involved with OTTL in the collector.

kentquirk:
[...]

PR #939, opened 2023-12-08
feat: Add `matches` operator to rules

kentquirk:

- Adds a regular expression `matches` operator to rules which should make some rules easier to write, especially when dealing with URLs or complex string fields.

- Add `matches` operator
- Remove some unused switch cases from `conditionMatchesOperator` (to avoid implementing another one)
- Add tests for it

This is part of #822.

This one will need further documentation, but that will be addressed in a separate PR.


MikeGoldsmith (approved):
Looks great @kentquirk  👍🏻

PR #941, opened 2023-12-08
maint: use command to check for other commands

robbkidd:

While `which` is commonly present on systems, `which` is usually a standalone executable. `command` is built-in to POSIX shells.

- `command` is even more likely to be present than the very-likely `which`
- `command` avoids spawning a subprocess to check for the presence of a command


kentquirk (approved):
Thanks!

MikeGoldsmith (approved):

PR #944, opened 2023-12-12
fix: Fix memory size parsing

tvdfly:

I pulled in the latest refinery code and it fails to parse this MaxAllow config value:

```yml
Collection:
  MaxAlloc: 1717600000 # 80% of 2GiB
```

The error is:

> Validation failed for config file refinery.yaml:
>   field Collection.MaxAlloc (1.7176e+06K) must be a valid memory size like '1Gb' or '100_000_000'

I tried updating to:
```yml
Collection:
  MaxAlloc: 1_717_600_000 # 80% of 2GiB
```

And that fails with:
> validateConfig unable to RELOAD config refinery.yaml: invalid size: 1_717_600_000

I think the regex was matching any character instead of periods, which is fixed here by escaping the period. It also was not matching underscores, which are now added to the match pattern.

The final issue was MemorySize.MarshalText() was using `%v` to serialize floats, which uses `%g` for "large" floats and `%f` for small floats. We always want `%f` so MemorySize.UnmarshalText() will work. I updated MemorySize.MarshalText() to use `%.0f`.


tvdfly:
The remaining issue is in MemorySize.MarshalText(). It will serialize `1717600000` as `1.7176e+06K`, which will break the validation when it tries to compare them as strings.

```
go test --run=TestMemorySize_Roundtrip ./config
--- FAIL: TestMemorySize_Roundtrip (0.00s)
    --- FAIL: TestMemorySize_Roundtrip/1717600000 (0.00s)
        memorysize_test.go:105: MemorySize.MarshalText() = 1.7176e+06K, want 1717600000
FAIL
FAIL    github.com/honeycombio/refinery/config  0.124s
FAIL
```

I think we probably want to compare MemorySize instances as uint64s; i'll add a commit that does that shortly.

config/memorysize.go
```diff
@@ -113,7 +113,7 @@ func (m MemorySize) MarshalText() ([]byte, error) {
 func (m *MemorySize) UnmarshalText(text []byte) error {
 	txt := string(text)
 
-	r := regexp.MustCompile(`^\s*(?P<number>[0-9.]+)(?P<unit>[a-zA-Z]*)\s*$`)
+	r := regexp.MustCompile(`^\s*(?P<number>[0-9\._]+)(?P<unit>[a-zA-Z]*)\s*$`)
```
kentquirk, line 116: Most metacharacters inside character classes (square brackets) don't apply, so you don't need to escape the `.`, so the backslash before is harmless but unnecessary. Please remove it as we don't use it elsewhere. The missing _ is a good catch, though.

config/memorysize_test.go
```diff
@@ -38,6 +38,12 @@ func TestMemorySize_UnmarshalText(t *testing.T) {
 		{"-1K", 0, true},
 		{"  4.5G", 4*G + 500*M, false},
 		{"4.5G  ", 4*G + 500*M, false},
+		{"1717600000 ", 1*G + 717*M + 600*K, false},
+		{" 1717600000 ", 1*G + 717*M + 600*K, false},
+		{" 1717600000", 1*G + 717*M + 600*K, false},
+		{"1717600K", 1*G + 717*M + 600*K, false},
+		{"100_000_000", 100 * M, false},
+		{"1G", 1 * G, false},
```
kentquirk, line 46: Thank you for the extra tests!

kentquirk (commented):
This is great, thank you. Please delete that one backslash and we can take this.

kentquirk (approved):
Thank you!

PR #945, opened 2023-12-12
maint: attempt to fix flaky integration tests

VinozzZ:

Since sending a trace is an asynchronous operation, there's a race condition between the writes from libhoney to the transmission buffer and the reads issued from our tests into the buffer. 

This PR changes tests to use `transmission.MockSender` instead. The mock sender uses a slice instead of an io.Writer as the buffer. This allows us to avoid race condition between reads and writes.

I also added a wait after sending the roundtrip request to make sure Refinery has a chance to flush its in-memory trace cache before we try to validate the sent trace.

### DIFF
--- a/app/app_test.go
+++ b/app/app_test.go
@@ -20,6 +20,7 @@ import (
 	"github.com/facebookgo/startstop"
 	"github.com/klauspost/compress/zstd"
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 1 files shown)


kentquirk (approved):
Oh! You found a MockSender, which I hope is less flaky. Nice.

PR #946, opened 2023-12-13
Add fly-dns peer management

tvdfly:

We run refinery in 30+ regions, and want to avoid running a centralized peer storage system. We could run redis-per-region; that's a lot of redis instance to track a couple peers in each region.

The Fly Platform provides a peer discovery mechanism with internal dns, which can do effectively the same thing as a redis-per-region without the extra infra.

Use the `<region>.<fly_app_name>.internal` dns record to discover peers in the same region. Configure it by setting the peer management type to `fly-dns`. The `fly-dns` peer manager will query dns record and update peers every 3 seconds (same as redis).


JamieDanielson (commented):
Thanks for the contribution @tvdfly ! This is something we'll need to think about more before we can give it a full review. As we're heading into the holiday season it will likely be the new year before we can circle back to this to ask some questions and dig into this more.

kentquirk:
We've moved to a pubsub model of finding peers, so unfortunately this is no longer a viable solution.

Issue #646, opened 2023-03-28
Rules based on Span Count don't work if the root span hasn't arrived


kentquirk:
**Is your feature request related to a problem? Please describe.**

See this [thread in Pollinators](https://honeycombpollinators.slack.com/archives/C4T3A32CX/p1680038083466499?thread_ts=1679947934.886969&cid=C4T3A32CX).

If we don't have a root span yet and the traceTimeout fires, we'd like it to be possible to drop an inprogress trace that has no root  because it's big -- but we can't, because the sampler is inspecting that `meta.refinery.span_count` field that is not added if the root span doesn't exist.

**Describe the solution you'd like**

A way to specify a rule that counts spans even if the count isn't attached to the root.

One option: a rule condition that applies at the trace level rather than a span field, that could inspect metadata about the trace. It needs further design work, but either specially-named fields could apply, or we could have specially-named conditions.

kentquirk:
Our existing meta value is badly named. We call it `meta.span_count` but it's actually the count of all the children of the root, including spans and span events. 

We should actually have several values:
* `meta.spanevent_count` -- the number of span events on the trace
* `meta.spanlink_count` -- the number of span links on the trace
* `meta.span_count` -- the number of spans on the trace
* `meta.event_count` -- the number of honeycomb events on the trace (the sum of the 3 preceding values)

In order to do this, we need to allow old configurations to continue to use `meta.span_count` and to opt into the new names.

There are 2 ways to do this that I've thought of:
* Bump the config file version number and add individual config values for each of these fields; if version is 2, then we use the old definition, but in version 3 we interpret the new way.
* Add a new, single config value called `AddSpanCountsToRoot`, with validation allowing only one of the two to be used, and if the plural form is used we use the new definitions.

Personally I prefer the second option, but I'd love to hear other thoughts.

kentquirk:
This should be addressed after #555 gets fixed, since they use the same underlying mechanisms.

PR #950, opened 2023-12-14
feat: use a computed field for current descendant count in rules

VinozzZ:

Metadata about a trace is added only to root spans. However, it's useful to make sampling decisions before root span arrives. In this PR, I addressed a specific use case described in [this issue](#646 )

This PR adds a new concept called `ComputedField` in rule based sampler. A `ComputedField` is a virtual field that's generated during rule evaluation for a trace. The virtual field isn't actually defined on a span. 

After some discussion with Kent, we decided to use `?.` as the prefix for computed field so that it shouldn't collide with any regular field names.

This PR doesn't include documentation. I will address it in [this PR for all rules related documentation](https://github.com/honeycombio/refinery/pull/951/files).

### DIFF
--- a/config/sampler_config.go
+++ b/config/sampler_config.go
@@ -30,12 +30,15 @@ const (
 	NotIn          = "not-in"
 )
 
+// ComputedField is a virtual field. It's value is calculated during rule evaluation.
+// We use the `?.` prefix to distinguish computed fields from regular fields.
+type ComputedField string
+
 const (
+	// ComputedFieldPrefix is the prefix for computed fields.
 	ComputedFieldPrefix               = "?."
 	NUM_DESCENDANTS     ComputedField = ComputedFieldPrefix + "NUM_DESCENDANTS"
 )
-
-type ComputedField string
 
 // The json tags in this file are used for conversion from the old format (see tools/convert for details).
 // They are deliberately all lowercase.
--- a/config/sampler_config.go
+++ b/config/sampler_config.go
@@ -30,6 +30,13 @@ const (
 	NotIn          = "not-in"
 )
 
+const (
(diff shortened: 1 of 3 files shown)


config/sampler_config.go
```diff
@@ -30,6 +30,13 @@ const (
 	NotIn          = "not-in"
 )
 
+const (
+	ComputedFieldPrefix               = "?."
+	NUM_DESCENDANTS     ComputedField = ComputedFieldPrefix + "NUM_DESCENDANTS"
+)
+
+type ComputedField string
```
kentquirk, line 38: Really minor nit, but I'd prefer to put the type before the const that uses it.

sample/rules.go
```diff
@@ -220,7 +220,15 @@ func ruleMatchesSpanInTrace(trace *types.Trace, rule *config.RulesBasedSamplerRu
 	return false
 }
 
-func extractValueFromSpan(span *types.Span, condition *config.RulesBasedSamplerCondition, checkNestedFields bool) (interface{}, bool) {
+func extractValueFromSpan(trace *types.Trace, span *types.Span, condition *config.RulesBasedSamplerCondition, checkNestedFields bool) (interface{}, bool) {
+	// If the condition is a descendant count, we extract the count from trace and return it.
+	if f, ok := condition.GetComputedField(); ok {
```
kentquirk, line 225: I like this design!

kentquirk (approved):
This is good work. I had a really minor nit that you can ignore if you wish.

config/sampler_config.go
VinozzZ: Yes, I agree that's better for readability. I will make that change

PR #953, opened 2023-12-15
docs: Add section on running tests to contributing guide

MikeGoldsmith:
Adds a section to the contributing guide to show to execute tests locally.

- Add test section to CONTRIBUTING.md
- Add asdf `.tools-version` file that sets the expected golang version to 1.21.1


CONTRIBUTING.md
```diff
@@ -16,6 +16,16 @@ at the root of the Refinery directory.
 
 `go run ./cmd/refinery/main.go -c config_complete.yaml -r rules_complete.yaml`
 
+# Running Refinery tests locally
+
+Refinery tests require a local instance of Redis to execute. A simple way to run this is using Docker:
+
+`docker run --name refinery-redis -p 6379:6379 -d redis`
```
kentquirk, line 23: This assumes that you have docker installed and the daemon is running, and that you've got an image for redis. And you get exactly the same result as if you had installed and run redis. Like, docker didn't actually do anything for you here and added a bunch of friction to the process.

Personally, I installed redis using brew, and it runs in the background all the time. Or I could run it on demand. 

I'm not anti-docker, but I do think it adds friction, and I'm not sure what it brings to the table here. In any case, if we're going to say this, we should probably also include how to get an appropriate image.

MikeGoldsmith: Sure, if you feel we don't need to hand-hold readers with getting redis setup I'm good to remove the docker suggestion.

Do you feel there's value in demonstrating how tests are expected to be run, with the caveat of requiring redis be available?

kentquirk: I do think there's value in demonstrating testing, and I think it's worth telling people they need redis. I was only questioning why you were running redis in docker instead of standalone. 

Maybe we just say "Tests require a local installation of redis. See [here](https://redis.io/docs/install/install-stack/) for how to get it running. "

MikeGoldsmith: Updated to refer to external docs for installing redis 👍🏻

kentquirk (approved):
Thank you.

PR #968, opened 2024-01-11
maint: bump deps for 2.4

fchikwekwe:

Closes #960 
Closes #961 
Closes #962 

Fixing all dependency updates for 2.4 release


fchikwekwe:
I'm looking into the build failures with the WindowedThroughput test. I think this might a test we've had trouble with in the past. The error message indicates a race condition so I was checking the `sample` package for any areas where we might need to lock data access.

kentquirk:
I just recreated the race condition on my machine. It's probably real. Looking into it now.

kentquirk:
Yup, it's real, and a problem in dynsampler-go. Fix is [here](https://github.com/honeycombio/dynsampler-go/pull/73). 

Once that's merged, we can pull it in as part of this PR.

fchikwekwe:
Oh shoot, I leave for dinner and you've already fixed it! Looking at that PR.

kentquirk:
You'll need to add an update of the dynsampler-go library, which should fix the test.

kentquirk (approved):
LGTM!

Issue #954, opened 2023-12-18
Refinery overwrites config settings by applying defaults after config was loaded


fchikwekwe:
[As seen in this section of code](https://github.com/honeycombio/refinery/blob/main/config/configLoadHelpers.go#L232), Refinery is loading in defaults after the config has already been loaded in. This can result in user settings being reset to default. We need to apply defaults first and then load in the config to overwrite sections with user settings.

robbkidd:
After our coding session today, I poked around at `creasty/defaults` and found [their test cases](https://github.com/creasty/defaults/blob/5220e08074d54cf55444eb46c6177f0beb84f43c/defaults_test.go#L579-L592) for "don't override an existing value with a default." I think [this change](https://github.com/creasty/defaults/compare/master...robbkidd:defaults:no-override-bools) demonstrates our failure scenario: a `true` default appearing as the final value despite `false` being set by the user.

fchikwekwe:
Opened an issue while still working on this https://github.com/creasty/defaults/issues/49.

fchikwekwe:
Closed by #969. Upstream work https://github.com/creasty/defaults/pull/50 and https://github.com/creasty/defaults/pull/51.

fchikwekwe:
Followup worked to be done for 2.5 release described in #970.

PR #969, opened 2024-01-18
fix: allow config bools to default to true

robbkidd:

- #954 

- add a failing test to reproduce the problem
- updated config struct fields that are bools defaulting to true to be bool pointers so that nil represents the unset state
- updated getter functions to handle the pointers
- [x] add a GetSamplerEnabled() function to HoneycombLoggerConfig to handle pointer defreferencing

## Maybe TODO

- [ ] use something (reflect?) to do struct tag lookups to find default values in the getter functions instead of declaring defaults in multiple places
- [ ] keep the default-true-bool-overridden-with-false test?
- [ ] contribute documentation about bool pointer workaround to upstream defaults library?

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -616,7 +616,7 @@ func TestHoneycombLoggerConfig(t *testing.T) {
 	assert.Equal(t, "http://honeycomb.io", loggerConfig.APIHost)
 	assert.Equal(t, "1234", loggerConfig.APIKey)
 	assert.Equal(t, "loggerDataset", loggerConfig.Dataset)
-	assert.Equal(t, true, *loggerConfig.SamplerEnabled)
+	assert.Equal(t, true, loggerConfig.GetSamplerEnabled())
 	assert.Equal(t, 5, loggerConfig.SamplerThroughput)
 }
 
@@ -639,7 +639,7 @@ func TestHoneycombLoggerConfigDefaults(t *testing.T) {
 
 	assert.NoError(t, err)
 
-	assert.Equal(t, true, *loggerConfig.SamplerEnabled)
+	assert.Equal(t, true, loggerConfig.GetSamplerEnabled())
 	assert.Equal(t, 10, loggerConfig.SamplerThroughput)
 }
 
@@ -910,7 +910,7 @@ func TestOverrideConfigDefaults(t *testing.T) {
 	assert.Equal(t, false, c.GetAddHostMetadataToTrace())
 	loggerConfig, err := c.GetHoneycombLoggerConfig()
 	assert.NoError(t, err)
-	assert.Equal(t, false, *loggerConfig.SamplerEnabled)
+	assert.Equal(t, false, loggerConfig.GetSamplerEnabled())
 	assert.Equal(t, false, c.GetCompressPeerCommunication())
 	assert.Equal(t, false, c.GetGRPCEnabled())
 }
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -890,6 +890,31 @@ func TestHoneycombIdFieldsConfigDefault(t *testing.T) {
 	assert.Equal(t, []string{"trace.parent_id", "parentId"}, c.GetParentIdFieldNames())
 }
 
+func TestOverrideConfigDefaults(t *testing.T) {
+	cm := makeYAML(
+		"General.ConfigurationVersion", 2,
+		"RefineryTelemetry.AddSpanCountToRoot", false,
+		"RefineryTelemetry.AddHostMetadataToTrace", false,
+		"HoneycombLogger.SamplerEnabled", false,
+		"Specialized.CompressPeerCommunication", false,
+		"GRPCServerParameters.Enabled", false,
+	)
+	rm := makeYAML("ConfigVersion", 2)
+	config, rules := createTempConfigs(t, cm, rm)
+	defer os.Remove(rules)
+	defer os.Remove(config)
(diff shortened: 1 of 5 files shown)


kentquirk:
Do you want to make this a draft PR until you have code for it?

robbkidd:
> Do you want to make this a draft PR until you have code for it?

Certainly! Drafted ... and there should be code for it shortly.

config/file_config.go
```diff
@@ -477,13 +477,28 @@ func (f *fileConfig) GetCompressPeerCommunication() bool {
 	f.mux.RLock()
 	defer f.mux.RUnlock()
 
-	return f.mainConfig.Specialized.CompressPeerCommunication
+	var compressPeerCommunication bool
+	if f.mainConfig.Specialized.CompressPeerCommunication == nil {
+		compressPeerCommunication = true
```
fchikwekwe, line 482: These true values are hardcoded, but we should try and lookup these values from the struct tag. That will require us to use the reflect package mimicking some behavior from the defaults package. Is this something that we want in the scope of this fix?

config/config_test.go
```diff
@@ -890,6 +890,31 @@ func TestHoneycombIdFieldsConfigDefault(t *testing.T) {
 	assert.Equal(t, []string{"trace.parent_id", "parentId"}, c.GetParentIdFieldNames())
 }
 
+func TestOverrideConfigDefaults(t *testing.T) {
```
robbkidd, line 893: Added in the first commit of this PR, demonstrate the 5 bools that default to true don't. Not sure if we keep this test long-term.

logger/honeycomb.go
```diff
@@ -59,7 +59,7 @@ func (h *HoneycombLogger) Start() error {
 		}
 	}
 
-	if loggerConfig.SamplerEnabled {
+	if *loggerConfig.SamplerEnabled {
```
fchikwekwe, line 62: We should consider adding a getter function for HoneycombLogger.SamplerEnabled so that we do not have to dereference in every time Sampler Enabled is used.

config/file_config.go
fchikwekwe: I'm not sure I'm happy with the hardcoded `true` values that are only present on struct values that are currently type `bool` and default `true`. It might be nice to write a helper function that we can add to add struct values to check is type is `bool` and default is `true`.

fchikwekwe: @robbkidd suggested that we either add some code to the validate function to check for struct tag type and value. Or perhaps there is a test that we can add later on to validate any changes to struct tags.

logger/honeycomb.go
robbkidd: During a chat this morning, @kentquirk, @fchikwekwe, and I agreed we should make the SamplerEnabled bool-pointer private and add a getter function to smooth the bool interface. That'll be work on this PR with other ideas to reduce tech debt in a follow-up issue.

config/file_config.go
```diff
@@ -119,10 +119,21 @@ type HoneycombLoggerConfig struct {
 	APIHost           string `yaml:"APIHost" default:"https://api.honeycomb.io"`
```
fchikwekwe, line 119: Not in scope for this PR, but accessor functions for all fields on `HoneycomLoggerConfig` would be nice.

config/file_config.go
```diff
@@ -119,10 +119,21 @@ type HoneycombLoggerConfig struct {
...
 	APIKey            string `yaml:"APIKey" cmdenv:"HoneycombLoggerAPIKey,HoneycombAPIKey"`
 	Dataset           string `yaml:"Dataset" default:"Refinery Logs"`
-	SamplerEnabled    bool   `yaml:"SamplerEnabled" default:"true"`
+	SamplerEnabled    *bool  `yaml:"SamplerEnabled" default:"true"` // Avoid pointer woe on access, use GetSamplerEnabled() instead.
```
robbkidd, line 122: I tossed this comment on the field so that it would be shown by IDEs that do the hover box thing. Like a ⚠️ sign saying "Hey, maybe don't use this directly."

robbkidd: Add this comment to all the `*bool`s in a separate commit (easily removed from this PR if that's too much commentary).

config/config_test.go
kentquirk: Let's just add a comment to this explaining why it exists.

config/file_config.go
kentquirk: I struggle with this; every accessor adds one form of tech debt to save another form of tech debt. In particular, when I broke up the config into subsections, I deliberately did not add accessors for everything. 

I don't feel strongly against it, I just wish there was a better way (JS/java properties would be nice but Go doesn't give us those).

config/file_config.go
kentquirk: Instead of reading the tags, we talked about something like:

```
type BoolDefaultTrue *bool
```

That would give us the hook to do some more useful linting or other work to make sure that we keep the contract for those types.

kentquirk (approved):
I think this is pretty good, and the only thing I'd like to see changed is to define that type for a bool that defaults to true, and use that appropriately; we don't have to get into modifying the defaults package or validating specially unless you see a really easy solution. I wouldn't spend a lot more time on it.

fchikwekwe:
Yes, I played around with adding the boolean `true` type a little bit this week, but I'll bump the rest of this work to be completed in #970.

PR #972, opened 2024-01-23
feat: Extract unique field names from sampler configs

VinozzZ:
To reduce the amount of across AZ zone data within a Refinery cluster, we came up with an idea to only store information needed for making a sampling decision in a central store. In order to extract those information from a span, we need to have a list of fields used in sampler configs.

This PR adds the code to generate the list mentioned above both during Refinery startup and config reload by adding the generating code in the `newFileConfig` function.

***Attention***
This PR is not merging into the current `main` branch. We are using a feature branch for the dynamic scaling work

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -331,7 +331,7 @@ func TestReadRulesConfig(t *testing.T) {
 
 	rules, err := c.GetAllSamplerRules()
 	assert.NoError(t, err)
-	assert.Len(t, rules.UniqueFields(), 10)
+	assert.Len(t, rules.UniqueSamplingFields(), 10)
 
 	d, name, err := c.GetSamplerConfigForDestName("doesnt-exist")
 	assert.NoError(t, err)
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -329,6 +329,10 @@ func TestReadRulesConfig(t *testing.T) {
 	c, err := getConfig([]string{"--no-validate", "--config", "../config.yaml", "--rules_config", "../rules_complete.yaml"})
 	assert.NoError(t, err)
 
+	rules, err := c.GetAllSamplerRules()
+	assert.NoError(t, err)
+	assert.Len(t, rules.UniqueFields(), 10)
+
 	d, name, err := c.GetSamplerConfigForDestName("doesnt-exist")
 	assert.NoError(t, err)
 	assert.IsType(t, &DeterministicSamplerConfig{}, d)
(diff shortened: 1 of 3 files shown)


config/config_test.go
```diff
@@ -329,6 +329,10 @@ func TestReadRulesConfig(t *testing.T) {
 	c, err := getConfig([]string{"--no-validate", "--config", "../config.yaml", "--rules_config", "../rules_complete.yaml"})
 	assert.NoError(t, err)
 
+	rules, err := c.GetAllSamplerRules()
+	assert.NoError(t, err)
+	assert.Len(t, rules.UniqueFields(), 10)
+
```
fchikwekwe, line 335: Won't this fail if any of the fields are empty?

VinozzZ: Do you mind to explain what do you mean by fields are empty? I'm not sure I understand what you are referring to

fchikwekwe: It appears to me that if a field is nil, then it wont be added to the fields map. That would change the length here right?

VinozzZ: The field is only added if it's present in the sampler configs. Currently, the test data this test is using has ten unique fields. If someone went to change the test data, then they would need to modify this assertion to the corresponding value. So you are right, it would change the length there

config/sampler_config.go
```diff
@@ -145,12 +145,46 @@ func (v *RulesBasedDownstreamSampler) NameMeaningfulRate() string {
...
+
+		if c, ok := s.(SamplerConfigger); ok {
+			for _, field := range c.GetFields() {
+				v.uniqueFields[field] = struct{}{}
+			}
+		}
+	}
+}
+
+type SamplerConfigger interface {
```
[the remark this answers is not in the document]
kentquirk: Doesn't look like you changed the interface?

kentquirk (approved):
Other than the name of that interface, this looks good.

config/sampler_config.go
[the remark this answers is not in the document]
VinozzZ: Weird, it must got lost in one of my commit. I will push it now

config/sampler_config.go
```diff
@@ -143,14 +143,52 @@ func (v *RulesBasedDownstreamSampler) NameMeaningfulRate() string {
...
+	GeFields() []string
+}
+
+var _ GetSamplingFielder = (*DeterministicSamplerConfig)(nil)
+
 type DeterministicSamplerConfig struct {
 	SampleRate int `json:"samplerate" yaml:"SampleRate,omitempty" default:"1" validate:"required,gte=1"`
 }
 
+func (d *DeterministicSamplerConfig) GeFields() []string {
```
kentquirk, line 186: Typo, and for consistency should be `GetSamplingFields`

PR #975, opened 2024-01-24
maint: Refactor cuckoo cache for reusability

VinozzZ:
In preparation of the upcoming dynamic scaling work, this PR does below two refactoring work for the cuckoo cache:

- make cuckooo cache reusable by accepting an interface instead of the type.Trace.
- Enclose sent reason cache in cuckoo cache so that states are not leaked.


collect/cache/cuckooSentCache.go
```diff
@@ -28,18 +28,29 @@ type keptTraceCacheEntry struct {
...
-func NewKeptTraceCacheEntry(trace *types.Trace) *keptTraceCacheEntry {
-	if trace == nil {
+type TraceInfo interface {
+	ID() string
+	SampleRate() uint
+	DescendantCount() uint32
+	SpanEventCount() uint32
+	SpanLinkCount() uint32
+	SpanCount() uint32
+	SetSentReason(uint)
```
VinozzZ, line 38: Since we are including the sentReasonCache inside the sent trace cache, we need a way to update the sent reason on the trace object inside the cache

collect/cache/cuckooSentCache.go
```diff
@@ -28,18 +28,29 @@ type keptTraceCacheEntry struct {
 	reason         uint32 // which rule was used to decide to keep the trace
 }
 
-func NewKeptTraceCacheEntry(trace *types.Trace) *keptTraceCacheEntry {
-	if trace == nil {
+type TraceInfo interface {
```
VinozzZ, line 31: This is the interface for both data sent to the central repository and to honeycomb

collect/collect.go
```diff
@@ -426,7 +424,14 @@ func (i *InMemCollector) processSpan(sp *types.Span) {
...
 	if trace.Sent {
-		i.dealWithSentTrace(cache.NewKeptTraceCacheEntry(trace), sp)
+		if sr, reason, found := i.sampleTraceCache.Check(sp); found {
+			// bump the count of records on this trace -- if the root span isn't
+			// the last late span, then it won't be perfect, but it will be better than
+			// having none at all
+			i.dealWithSentTrace(sr, reason, sp)
+			return
+		}
+		// TODO: would it be possible that it's not in the cache?
```
kentquirk, line 434: If trace.Sent is true, that means that we found it in the active trace cache and then to get here it would have to not be in the sampleTraceCache. I think in that weird edge case, we should just use "late" as the reason and do something like this:
`i.dealWithSentTrace(cache.NewKeptTraceCacheEntry(trace), "late", sp)`

kentquirk (approved):
This seems to be what we had talked about. Thanks!

PR #978, opened 2024-01-25
maint: finish conversion to generics.Set

kentquirk:
NOTE -- this is against the `feat_dynamic_scaling` branch.

- Finishes using `generic.Set` everywhere we can

- Convert to Set in sampler_config.go
- Rename SetUniqueSamplingFields to ExtractUniqueSamplingFields because Set implies that it takes a parameter.


robbkidd (approved):
It's a pretty good set.

<details><summary>Nice set.</summary>

Set
</details>

Issue #965, opened 2024-01-05
not-exists is a footgun


kentquirk:
We had an incident at Honeycomb where a `not-exists` rule was used with `Scope: trace`, which meant that all traces lacking a particular field in ANY ONE SPAN were set to a sample rate of 1. An otherwise innocent change happened to occur at a time when that field was not added to the span, and it caused every trace to be kept, resulting in DDOSing us. 

It's hard to think of a use case for `not-exists` and `Scope: trace` to be used together. 

[This internal slack thread](https://houndsh.slack.com/archives/C067QGXHWKC/p1704403927884669) discusses it. We should consider what to do about this. Options include:

- [x] #979 
- [x] #993
- ~nerf the feature entirely~

robbkidd:
Resolved with documentation in #979. Maybe we'll do more with warning when the negative operators are used for #993.

PR #979, opened 2024-01-29
docs: include a warning about surprising not-exists behavior

robbkidd:

- #965 

- Update rulesMeta description for the Condition Operator to include a warning about using `not-exists` in a rule without also scoping that rule to "span".

## TODO

- [x] agree on wording
- ~regenerate documentation with `tools/convert/make` and commit the results to this branch also~ Doc generation happens in one-pass for all the changes prior to release.

### DIFF
--- a/config/metadata/rulesMeta.yaml
+++ b/config/metadata/rulesMeta.yaml
@@ -651,12 +651,12 @@ groups:
         summary: is the comparison operator to use.
         description: >
           The comparison operator to use. String comparisons are case-sensitive.
-          For most cases, use the negative operators (`!=`, `does-not-contain`, and
-          `not-exists`) in a rule with a scope of "span". WARNING: A negative operator
-          in a rule with a scope of "trace" (the default) can cause the rule to match
-          a trace if *any* span in the trace matches the negative condition, for example
-          a single span in a trace lacks the specified field (`not-exists`).
-          This is rarely the matching behavior you want.
+          For most cases, use negative operators (`!=`, `does-not-contain`, and
+          `not-exists`) in a rule with a scope of "span".
+          WARNING: Rules can have `Scope: trace` or `Scope: span`; a negative
+          operator with `Scope: trace` will be true if **any** single span in the
+          entire trace matches the negative condition.
+          This is almost never desired behavior.
       - name: Value
         type: anyscalar
         summary: is the value to compare against.
--- a/config/metadata/rulesMeta.yaml
+++ b/config/metadata/rulesMeta.yaml
@@ -651,11 +651,11 @@ groups:
         summary: is the comparison operator to use.
         description: >
           The comparison operator to use. String comparisons are case-sensitive.
-          WARNING: using the `not-exists` operator in a rule with a scope of
+          For most cases, `not-exists` should only be used in a rule with a scope
+          of "span". WARNING: The `not-exists` operator in a rule with a scope of
           "trace" (the default) will cause the rule to match a trace if *any* span
-          in the trace does not have the field. This is almost certainly not the
-          matching behavior you want. Use `not-exists` in a rule with a scope of
-          "span".
+          in the trace lacks the specified field. This is rarely the matching
+          behavior you want.
       - name: Value
         type: anyscalar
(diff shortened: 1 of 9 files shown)


TylerHelmuth (approved):

kentquirk (approved):
I also suggest adding something like the same text, (or like below) to the `not-exists` section of the file file `rules_conditions.md` which is a relatively recent addition? 

```
Note that `not-exists` test, when applied at the trace level (`Scope: trace`) means that the test will succeed if any span in the trace is lacking the specified field. This is rarely the desired behavior.
```

Note -- I'm quite happy with your wording, this is an alternative you can use or drop at your preference.

config/metadata/rulesMeta.yaml
```diff
@@ -651,6 +651,11 @@ groups:
         summary: is the comparison operator to use.
         description: >
           The comparison operator to use. String comparisons are case-sensitive.
+          WARNING: using the `not-exists` operator in a rule with a scope of
+          "trace" (the default) will cause the rule to match a trace if *any* span
+          in the trace does not have the field. This is almost certainly not the
+          matching behavior you want. Use `not-exists` in a rule with a scope of
+          "span".
```
fchikwekwe, line 658:
```suggestion
          The comparison operator to use. String comparisons are case-sensitive.
          For most cases, use `not-exists` in a rule with a scope of "span".
          WARNING: using the `not-exists` operator in a rule with a scope of
          "trace" (the default) will cause the rule to match a trace if *any* span
          in the trace does not have the field. This is almost certainly not the
          matching behavior you want. 
```

I think this could be worded better, but the general suggestion is to put the desired behavior first and then outline the warning behavior.

robbkidd: I have reworded to "do this" first with a Frankenstein's monster of my original, Kent's suggestion, and your suggestion

config/metadata/rulesMeta.yaml
```diff
@@ -651,6 +651,12 @@ groups:
         summary: is the comparison operator to use.
         description: >
           The comparison operator to use. String comparisons are case-sensitive.
+          For most cases, use the negative operators (`!=`, `does-not-contain`, and
```
mjingle, line 654:
```suggestion
          For most cases, use negative operators (`!=`, `does-not-contain`, and
```

config/metadata/rulesMeta.yaml
```diff
@@ -651,6 +651,12 @@ groups:
...
+          `not-exists`) in a rule with a scope of "span". WARNING: A negative operator
+          in a rule with a scope of "trace" (the default) can cause the rule to match
+          a trace if *any* span in the trace matches the negative condition, for example
```
mjingle, line 657: We only use bold in docs, not italics.
```suggestion
          a trace if **any** span in the trace matches the negative condition. For example,
```

config/metadata/rulesMeta.yaml
```diff
@@ -651,6 +651,12 @@ groups:
...
+          a single span in a trace lacks the specified field (`not-exists`).
+          This is rarely the matching behavior you want.
```
mjingle, line 659: "This" somewhat obscures the intended outcome, and I understand the example but it seems to be an incomplete thought? (Or maybe more clarification is needed.)

kentquirk: I think an example is counterproductive unless you want to get much more detailed. Maybe breaking it up this way helps?

> WARNING: Rules can have `Scope: trace` or `Scope: span`; a negative operator with `Scope: trace` will be true if **any** single span in the entire trace matches the negative condition.

mjingle: Oooooh, I like this suggested phrasing a lot more. And I agree about omitting the example. Thank you @kentquirk! cc: @robbkidd

robbkidd: Revised to this wording in 893d2bc

mjingle (approved):
👍

fchikwekwe (approved):

fchikwekwe:
LGTM!

Issue #906, opened 2023-11-22
Allow rules and keys to specify some level of span context (particularly root span)


kentquirk:
**Is your feature request related to a problem? Please describe.**
Multiple customers ([one example](https://honeycombpollinators.slack.com/archives/C01BZ3J7RMJ/p1700673134994119)) have requested the ability to specify keys or rules that are restricted to a subset of possible spans. 

**Describe the solution you'd like**
Perhaps being able to specify something like `root.column_name` is enough to solve most use cases.

**Describe alternatives you've considered**
OTTL, or a more elaborate conditional structure.

**Additional context**

VinozzZ:
related to #646

fchikwekwe:
This issue will technically be solved when #981 merges, but the change is not optimal. Right now, we check root span context for every span in a trace, but since all spans share a root span we only need to do this once. @kentquirk suggested that we add a flag to `extractValueFromSpan` to see if the root span for that trace has already been checked.

kentquirk:
Closed by #1018

PR #981, opened 2024-01-31
feat: allow user to sample on root span context

fchikwekwe:

- Addresses #906, additional work for that issue outlined in comments 

- Allow specifying root span context to rules

### DIFF
--- a/config/metadata/rulesMeta.yaml
+++ b/config/metadata/rulesMeta.yaml
@@ -611,11 +611,24 @@ groups:
         description: >
           The field to check. This can name any field in the trace. If the field
           is not present, then the condition will not match. The comparison is
-          case-sensitive. The field can also include span context such as root 
-          span context (e.g. `root.http.status`). Specifying root span context 
-          limits matching conditions to those that match the root span. Using root
-          span context also means that if the trace matches, every span within the
-          trace will match since they share a root span.
+          case-sensitive. 
+          
+          The field can also include a prefix that changes the span used for evaluation of the field.
+          The only prefix currently supported is `root`, as in `root.http.status`. Specifying `root.` 
+          causes the condition to be evaluated against the root span. 
+          
+          For example:
+          
+          Field: root.url
+          Operator: contains
+          Value: login
+
+          This condition will evaluate to true when the field called `url` on the root span of the 
+          trace contains `login`.
+          
+          The rule setting `Scope: span` does not change the meaning of this prefix --
+          the condition is still evaluated on the root span and is treated as if it were part of the  
+          span being processed. 
       - name: Fields
         type: stringarray
         valuetype: stringarray
--- a/config/metadata/rulesMeta.yaml
+++ b/config/metadata/rulesMeta.yaml
@@ -618,8 +618,7 @@ groups:
           causes the condition to be evaluated against the root span. 
           
           For example, if the `Field` is `root.url`, then the condition will be processed using the url 
-          field from the root span. This condition will evaluate to true when the field called `url` 
-          on the root span of the trace contains `login`.
+          field from the root span. 
           
           The rule setting `Scope: span` does not change the meaning of this prefix --
           the condition is still evaluated on the root span and is treated as if it were part of the  
--- a/config/metadata/rulesMeta.yaml
+++ b/config/metadata/rulesMeta.yaml
@@ -619,9 +619,11 @@ groups:
           
           For example:
           
+          ```
           Field: root.url
           Operator: contains
           Value: login
+          ```
 
           This condition will evaluate to true when the field called `url` on the root span of the 
           trace contains `login`.
--- a/config/metadata/rulesMeta.yaml
+++ b/config/metadata/rulesMeta.yaml
@@ -611,7 +611,11 @@ groups:
         description: >
           The field to check. This can name any field in the trace. If the field
           is not present, then the condition will not match. The comparison is
-          case-sensitive.
+          case-sensitive. The field can also include span context such as root 
+          span context (e.g. `root.http.status`). Specifying root span context 
+          limits matching conditions to those that match the root span. Using root
+          span context also means that if the trace matches, every span within the
+          trace will match since they share a root span. 
       - name: Fields
         type: stringarray
         valuetype: stringarray
--- a/config/metadata/rulesMeta.yaml
+++ b/config/metadata/rulesMeta.yaml
@@ -620,7 +620,7 @@ groups:
           For example, if the `Field` is `root.url`, then the condition will be processed using the url 
           field from the root span. 
           
-          The rule setting `Scope: span` does not change the meaning of this prefix --
+          The setting `Scope: span` for a rule does not change the meaning of this prefix --
           the condition is still evaluated on the root span and is treated as if it were part of the  
           span being processed. 
       - name: Fields
--- a/config/metadata/rulesMeta.yaml
+++ b/config/metadata/rulesMeta.yaml
@@ -617,16 +617,9 @@ groups:
           The only prefix currently supported is `root`, as in `root.http.status`. Specifying `root.` 
           causes the condition to be evaluated against the root span. 
           
-          For example:
-          
-          ```
-          Field: root.url
-          Operator: contains
-          Value: login
-          ```
-
-          This condition will evaluate to true when the field called `url` on the root span of the 
-          trace contains `login`.
+          For example, if the `Field` is `root.url`, then the condition will be processed using the url 
+          field from the root span. This condition will evaluate to true when the field called `url` 
+          on the root span of the trace contains `login`.
           
           The rule setting `Scope: span` does not change the meaning of this prefix --
           the condition is still evaluated on the root span and is treated as if it were part of the  
--- a/config/metadata/rulesMeta.yaml
+++ b/config/metadata/rulesMeta.yaml
@@ -615,7 +615,7 @@ groups:
           span context (e.g. `root.http.status`). Specifying root span context 
           limits matching conditions to those that match the root span. Using root
           span context also means that if the trace matches, every span within the
-          trace will match since they share a root span. 
+          trace will match since they share a root span.
       - name: Fields
         type: stringarray
         valuetype: stringarray
--- a/rules_conditions.md
+++ b/rules_conditions.md
@@ -97,7 +97,7 @@ Conditions:
     Datatype: int
 ```
 
-## Adding Span Context to Conditions
+## Using a Prefix to Identify a Field in a Related Span
 
 You can limit conditions to a particular span context. Right now, the `root` prefix is the only context supported. 
 This limits matching spans to those where the root span matches the condition. Note that this behavior obsoletes matching 
--- a/rules_conditions.md
+++ b/rules_conditions.md
@@ -97,6 +97,23 @@ Conditions:
     Datatype: int
 ```
 
+## Adding Span Context to Conditions
+
+You can limit conditions to a particular span context. Right now, the `root` prefix is the only context supported. 
+This limits matching spans to those where the root span matches the condition. Note that this behavior obsoletes matching 
+specific spans within a trace since the entire trace shares a root span. 
+
+```yaml
+Rules:
+    - Name: limit by root span context
+      Conditions:
+        Field: "root.http.status"
+        Operator: =
+        Value: "500"
+        Datatype: string
+```
+
+
 ## `Operator`
 
 The `Operator` parameter controls how rules are evaluated.
--- a/rules_conditions.md
+++ b/rules_conditions.md
@@ -99,9 +99,8 @@ Conditions:
 
 ## Using a Prefix to Identify a Field in a Related Span
 
-You can limit conditions to a particular span context. Right now, the `root` prefix is the only context supported. 
-This limits matching spans to those where the root span matches the condition. Note that this behavior obsoletes matching 
-specific spans within a trace since the entire trace shares a root span. 
+Fields can contain a span selection prefix. Today, the only prefix supported is `root`.
+This prefix causes the root span to be searched for the specified field, rather than the span being evaluated.
 
 ```yaml
 Rules:
--- a/sample/rules.go
+++ b/sample/rules.go
@@ -120,7 +120,8 @@ func (s *RulesBasedSampler) GetSampleRate(trace *types.Trace) (rate uint, keep b
 				reason += rule.Name + ":" + samplerReason
 			} else {
 				rate = uint(rule.SampleRate)
-				keep = !rule.Drop && rule.SampleRate > 0 && rand.Intn(rule.SampleRate) == 0
+				randInt := rand.Intn(rule.SampleRate)
+				keep = !rule.Drop && rule.SampleRate > 0 && randInt == 0
 				reason += rule.Name
 				s.Metrics.Histogram(s.prefix+"sample_rate", float64(rate))
 			}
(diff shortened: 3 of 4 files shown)


fchikwekwe:
Something is still broken with this. I'll pick it up tomorrow.

fchikwekwe:
The broken thing was my tests. Should be working now.

config/metadata/rulesMeta.yaml
```diff
@@ -609,7 +609,11 @@ groups:
         description: >
           The field to check. This can name any field in the trace. If the field
           is not present, then the condition will not match. The comparison is
-          case-sensitive.
+          case-sensitive. The field can also include span context such as root 
+          span context (e.g. `root.http.status`). Specifying root span context 
+          limits matching conditions to those that match the root span. Using root
+          span context also means that if the trace matches, every span within the
+          trace will match since they share a root span.
```
kentquirk, line 616:
```suggestion
          case-sensitive. 
          
          The field can also include a prefix that changes the span used for evaluation of the field.
          The only prefix currently supported is `root`, as in `root.http.status`. Specifying `root.` 
          causes the condition to be evaluated against the root span. 
          
          For example:
          
          Field: root.url
          Operator: contains
          Value: login

          This condition will evaluate to true when the field called `url` on the root span of the 
          trace contains `login`.
          
          The rule setting `Scope: span` does not change the meaning of this prefix --
          the condition is still evaluated on the root span and is treated as if it were part of the  
          span being processed. 
```

kentquirk: I'm not finding the terms "limit" and "span context" to be the right phrasing for this. I tried something above -- feel free to wordsmith it if it's not working for you.

I also think we should tweak one of the examples in `rules_complete.md`.

rules_conditions.md
```diff
@@ -97,6 +97,23 @@ Conditions:
     Datatype: int
 ```
 
+## Adding Span Context to Conditions
+
+You can limit conditions to a particular span context. Right now, the `root` prefix is the only context supported. 
+This limits matching spans to those where the root span matches the condition. Note that this behavior obsoletes matching 
+specific spans within a trace since the entire trace shares a root span.
```
kentquirk, line 104:
```suggestion
Fields can contain a span selection prefix. Today, the only prefix supported is `root`.
This prefix causes the root span to be searched for the specified field, rather than the span being evaluated.
```

rules_conditions.md
```diff
@@ -97,6 +97,23 @@ Conditions:
     Datatype: int
 ```
 
+## Adding Span Context to Conditions
```
kentquirk, line 100:
```suggestion
## Using a Prefix to Identify a Field in a Related Span
```

sample/rules.go
```diff
@@ -233,6 +234,13 @@ func extractValueFromSpan(trace *types.Trace, span *types.Span, condition *confi
 	var value any
 	var exists bool
 	for _, field := range condition.Fields {
+		// check if field prefix "root."
+		if strings.HasPrefix(field, "root.") {
```
kentquirk, line 238: let's pull this prefix out into a constant, and when we strip it off on the next line, strip by len(constant). That way we're safe if we decide to change the name.

kentquirk (commented):
I'd like to find different wording for the docs, but the code looks good. Also think we should add or modify an example in the rules_complete.md file.

config/metadata/rulesMeta.yaml
fchikwekwe: I played with this a little bit, but I think your wording is good here. I'll go ahead and commit this change.

sample/rules.go
fchikwekwe: Updated this!

kentquirk:
We had talked about an optimization that would mean we wouldn't have to re-evaluate this on every span. Are you expecting to implement that as part of this PR or a different one?

fchikwekwe:
@kentquirk I'm trying to get my optimization working and passing tests today. If I'm able to do that, then I will bring it in with this PR. If not, then I will merge another PR with that change soon.

sample/rules.go
```diff
@@ -190,11 +202,21 @@ func ruleMatchesSpanInTrace(trace *types.Trace, rule *config.RulesBasedSamplerRu
...
 	for _, span := range trace.GetSpans() {
 		ruleMatched := true
-		for _, condition := range rule.Conditions {
-			// whether this condition is matched by this span.
-			value, exists := extractValueFromSpan(trace, span, condition, checkNestedFields)
+		for i, condition := range rule.Conditions {
+			var value interface{}
+			var exists bool
+			// if condition has a root prefix, call extractValueFromSpan and store index in seen
+			if !slices.Contains(seen, i) && strings.Contains(condition.Field, RootPrefix) { // also maybe check Fields
```
fchikwekwe, line 213: If a user is designating multiple fields using the `Fields` field (say that five times fast), then should we validate that they either all have root span prefix or none of them do? Should we allow users to mix and match span contexts when specifying multiple fields?

kentquirk: I don't think we should forbid it. I could imagine someone saying "if I don't have `user` on the current span, take `root.user`.

robbkidd: This has me realizing that `Fields` is sort of an `OR`. Or like a derived column. 🤔🤔🤔

config/metadata/rulesMeta.yaml
```diff
@@ -609,7 +609,24 @@ groups:
...
-          case-sensitive.
+          case-sensitive. 
+          
+          The field can also include a prefix that changes the span used for evaluation of the field.
+          The only prefix currently supported is `root`, as in `root.http.status`. Specifying `root.` 
+          causes the condition to be evaluated against the root span. 
+          
+          For example:
+          
+          Field: root.url
```
kentquirk, line 620: can you please add code markers around this section? That's something that `suggest` doesn't support.

fchikwekwe: Yes, good callout!

sample/rules.go
fchikwekwe: added code to support this behavior.

sample/rules.go
```diff
@@ -233,11 +272,20 @@ func extractValueFromSpan(trace *types.Trace, span *types.Span, condition *confi
 	var value any
 	var exists bool
 	for _, field := range condition.Fields {
+		// check if rule uses root span context
+		if strings.HasPrefix(field, RootPrefix) {
+			field = field[len(RootPrefix):]
+			span = trace.RootSpan
+			// if this condition has a root prefix and the root span has already been checked, set flag to true
+			checkedRootPrefix = true
```
kentquirk, line 280: I don't think this is safe in a multitasking environment. Instead of using a global here, why not return this value from this function as an extra return parameter?

fchikwekwe: Hey kent. Part of the reason that I tried to avoid returning this value from the `extractValueFromSpan` function is because the `trace` and `span` context have to handle these checks differently. I see your point about the global variable not being safe. Do you think that using a variable local to the `ruleMatches...` function would work here? I'll see if I can try pushing something to that effect soon-ish. a

fchikwekwe: I just wanted to elaborate here that my goal with this optimization is to call `extractValueFromSpan` as few times as necessary. Returning this value from that function would necessitate more function calls from the `trace` scope.

kentquirk (changes requested):
Hey there. I think I have a much simpler (and concurrency-safe) solution to this. Can we please pair so I can talk you through what I'm thinking?

fchikwekwe:
I think that this has been refactored to fit your specifications now @kentquirk would you take a look?

config/metadata/rulesMeta.yaml
```diff
@@ -611,7 +611,19 @@ groups:
...
           is not present, then the condition will not match. The comparison is
-          case-sensitive.
+          case-sensitive. 
+          
+          The field can also include a prefix that changes the span used for evaluation of the field.
+          The only prefix currently supported is `root`, as in `root.http.status`. Specifying `root.` 
+          causes the condition to be evaluated against the root span. 
+          
+          For example, if the `Field` is `root.url`, then the condition will be processed using the url 
+          field from the root span. This condition will evaluate to true when the field called `url`
```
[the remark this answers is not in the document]
fchikwekwe: Removed!

sample/rules.go
```diff
@@ -159,15 +162,21 @@ func ruleMatchesTrace(t *types.Trace, rule *config.RulesBasedSamplerRule, checkN
 				matched++
 				continue
 			} else {
-				// if HasRootSpan is one of the conditions and it didn't match,
```
kentquirk, line 162: why remove this comment? it explains the return false.

sample/rules.go
```diff
@@ -159,15 +162,21 @@ func ruleMatchesTrace(t *types.Trace, rule *config.RulesBasedSamplerRule, checkN
...
 				return false
 			}
+
 		}
+		var checkedRootPrefix bool
 
 	span:
 		for _, span := range t.GetSpans() {
-			value, exists := extractValueFromSpan(t, span, condition, checkNestedFields)
+			var value interface{}
```
kentquirk, line 173:
```suggestion
			var value any
```
We should prefer using `any` to `interface{}` in new code.

sample/rules.go
```diff
@@ -220,37 +253,44 @@ func ruleMatchesSpanInTrace(trace *types.Trace, rule *config.RulesBasedSamplerRu
 	return false
 }
 
-func extractValueFromSpan(trace *types.Trace, span *types.Span, condition *config.RulesBasedSamplerCondition, checkNestedFields bool) (interface{}, bool) {
+func extractValueFromSpan(trace *types.Trace, span *types.Span, condition *config.RulesBasedSamplerCondition, checkNestedFields bool) (value interface{}, exists bool, checkRootPrefix bool) {
+	checkedRootPrefix := false
```
kentquirk, line 257:
```suggestion
```
The specification of names in the return value declares those as local variables and initializes them to the zero value. With this initialization, you have created a new local variable that masks the return value. You should skip the initialization here.

sample/rules.go
```diff
@@ -190,11 +199,35 @@ func ruleMatchesSpanInTrace(trace *types.Trace, rule *config.RulesBasedSamplerRu
...
-		for _, condition := range rule.Conditions {
-			// whether this condition is matched by this span.
-			value, exists := extractValueFromSpan(trace, span, condition, checkNestedFields)
+		for i, condition := range rule.Conditions {
+			var value interface{}
+			var exists bool
+			if slices.Contains(seen, i) {
+				continue
+			} else {
+				if strings.Contains(condition.Field, RootPrefix) {
```
kentquirk, line 212: Validation checks that you don't define both Field and Fields for a given rule, and during the Init() function, if it's Field we just copy it into Fields[] as the only element, so that we never have to check anything but Fields. You can eliminate this condition entirely.

sample/rules.go
```diff
@@ -190,11 +199,35 @@ func ruleMatchesSpanInTrace(trace *types.Trace, rule *config.RulesBasedSamplerRu
...
+				}
+
+				for _, field := range condition.Fields {
+					if strings.Contains(field, RootPrefix) {
+						value, exists, _ = extractValueFromSpan(trace, span, condition, checkNestedFields)
+						if !slices.Contains(seen, i) { // only append to seen for first field with root prefix
+							seen = append(seen, i)
+						}
+					}
+					// call extractValueFromSpan for each field without root prefix
```
kentquirk, line 224: I'm sorry, but I don't understand this code. You are calling this function again unconditionally at two different levels, and in none of these cases have the parameters changed at all. This either needs some conditions or some extra assignments. 

This code also doesn't explain `seen` so that I understand its use. Can you please elaborate?

kentquirk (changes requested):
I'm sorry, but I think this still needs work -- I think it at least needs some explanatory text but I fear the logic still needs work.

sample/rules.go
fchikwekwe: Thanks, I think this just need a single else clause to be more concise. And I will add a comment for `seen`.

sample/rules.go
fchikwekwe: Great, I'll remove it.

sample/rules.go
fchikwekwe: Will do.

sample/rules.go
fchikwekwe: Yeah I think I overcorrected here. I'll add it back.

sample/rules_test.go
```diff
@@ -2152,3 +2157,426 @@ func TestRulesWithDeterministicSampler(t *testing.T) {
...
+				if span.TraceID != "" {
+					require.Nil(t, trace.RootSpan, "Only set the trace ID on one span in a test case to designate which is the root. This test case appears to have multiple spans with trace IDs set.")
+					trace.RootSpan = span
+				}
+			}
+
+			spans := trace.GetSpans()
+			assert.Len(t, spans, len(d.Spans), "should have the same number of spans as input")
+
+			rate, _, reason, key := sampler.GetSampleRate(trace)
```
fchikwekwe, line 2571: It does not make sense to test on `ExpectedKeep` here because it makes the test flakey. It we are sampling values at 1 in 10 then `ExpectedKeep` will fail on every tenth test. We do test `keep` in other tests.

sample/rules.go
```diff
@@ -190,11 +201,30 @@ func ruleMatchesSpanInTrace(trace *types.Trace, rule *config.RulesBasedSamplerRu
...
+
 	for _, span := range trace.GetSpans() {
 		ruleMatched := true
-		for _, condition := range rule.Conditions {
-			// whether this condition is matched by this span.
-			value, exists := extractValueFromSpan(trace, span, condition, checkNestedFields)
+		for i, condition := range rule.Conditions {
+			var value any
+			var exists bool
+			if slices.Contains(seen, i) {
```
kentquirk, line 212: Ok, now that I understand what "seen" is, we still have a problem. What it's trying to do is not yet correct.

If either all or none of the entries in Fields are in `root`, I think it works. But imagine that your condition is:
```
Fields: http.status, root.http.status
Operator: '='
Value: 500
```
You look at the first span, and neither root nor the span has a status code of 500, and you mark it as "seen". The next span's code IS 500, but because `seen` has been set for this rule, you won't detect it.

This is a problem in both the Trace and Span scopes. 

I know this is frustrating that we keep iterating on this, but I didn't understand until now what you were trying to do. 

I think what we should do is:
* Write one or more tests that would detect the above scenario.
* `extractValueFromSpan` needs to be fixed -- `checkedRootPrefix` needs to be renamed to `usedRootPrefix` and should only be true on return if `value` came from a field using `root.` rather than the span being tested.
* But now that we've done that, we're not going to use it yet. We should call it without optimizations in both scopes and prove that everything works.
* Then, and only then, should we consider whether there are paths to avoid evaluating the root on every span. I think it will be easier to think about once we have the basic logic correct.

I apologize for any frustration but we have to get this right.

fchikwekwe:
I reverted this PR to before the optimization work. Based on Kent's comments and some additional discussion, it might be best to make sure tests are passing for the unoptimized version of this code and work on the optimization in a follow-up PR.

sample/rules_test.go
```diff
@@ -2152,3 +2157,393 @@ func TestRulesWithDeterministicSampler(t *testing.T) {
 		assert.Len(t, spans, len(d.Spans), "should have the same number of spans as input")
 	}
 }
+
+func TestRulesRootSpanContext(t *testing.T) {
+	data := []TestRulesData{
+		//{
```
kentquirk, line 2163: Looks like you commented this out to have one test case but left it that way. Please fix.

robbkidd (approved):
Refinery, now with more root! 

Foamy root beer.

fchikwekwe:
Follow-up work here: https://github.com/honeycombio/refinery/pull/999

PR #982, opened 2024-02-01
feat: POC for redis remote store implementation

VinozzZ:


centralstore/redis_store.go
```diff
@@ -0,0 +1,605 @@
...
+
+	}
+
+	if !exists {
+		// new trace
+		err := collectingSet.AddTrace(span.TraceID)
+		if err != nil {
+			return err
+		}
+		err = r.traces.AddTrace(&CentralTraceStatus{
```
VinozzZ, line 89: This should be storing the CentralSpan object as well

PR #988, opened 2024-02-01
feat(kafka-receiver): add flags/configs for kafka receiving

lizthegrey:

-

-


config.md
```diff
@@ -136,6 +136,39 @@ This is the amount of time after which if the http server does not see any activ
 - Type: `duration`
 - Default: `0s`
 
+### `KafkaBootstrapAddr`
```
kentquirk, line 139: I'd like to see this moved out to its own group rather than just being in the network configuration, and I'd also give it its own config type. If you call the group "Kafka" you can remove that prefix from all of its names; I think you'll probably want an Enabled flag on it as well. 

Then, you could consider having a single accessor that returns the entire group, or leave the separate ones.

lizthegrey: Done.

PR #991, opened 2024-02-02
fix: flaky TestOriginalSampleRateIsNotedInMetaField

robbkidd:

- [a flaky test](https://app.circleci.com/pipelines/github/honeycombio/refinery/2623/workflows/95075937-b4b3-472e-8d34-a9e2a7e9ba31/jobs/7681/tests#failed-test-0)

- drop the assumption that only 2 spans are sent to transmission: don't inspect the second event on the transmission queue, inspect the last event

### DIFF
--- a/collect/collect_test.go
+++ b/collect/collect_test.go
@@ -196,7 +196,7 @@ func TestOriginalSampleRateIsNotedInMetaField(t *testing.T) {
 	// Find the Refinery-sampled-and-sent event that had no upstream sampling which
 	// should be the last event on the transmission queue.
 	var noUpstreamSampleRateEvent *types.Event
-	assert.Eventually(t, func() bool {
+	require.Eventually(t, func() bool {
 		transmission.Mux.RLock()
 		defer transmission.Mux.RUnlock()
 		noUpstreamSampleRateEvent = transmission.Events[len(transmission.Events)-1]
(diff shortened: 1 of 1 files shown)


collect/collect_test.go
```diff
@@ -156,41 +162,49 @@ func TestOriginalSampleRateIsNotedInMetaField(t *testing.T) {
...
 		},
-	}
+	})
+	require.NoError(t, err, "should be able to add the span")
 
-	coll.AddSpan(span)
+	// Find the Refinery-sampled-and-sent event that had no upstream sampling which
+	// should be the last event on the transmission queue.
+	var noUpstreamSampleRateEvent *types.Event
 	assert.Eventually(t, func() bool {
```
VinozzZ, line 199: I know you didn't change this part of the test in this PR. I'm just wondering if using `require` instead of `assert` here would be easier to identify failures here. If the assertion for `"no-upstream-stampling` has failed, the following check won't succeed either.

kentquirk (approved):
Yingrong's comment is appropriate - we should be using `require` more frequently, and this is a good place to start. But even without that, I'm a huge fan of deflakifying things. Thanks for the attention!

collect/collect_test.go
robbkidd: I'm a big fan of `require`—this PR adds it to this test file—and I'm not sure why I didn't use it here.

nodding kermit

VinozzZ (approved):
Thank you for fixing the flaky test and all the beautiful comments 💯

robbkidd:
Me liberally sprinkling tests and code comments everywhere.
From the movie Peter Pan, a young boy floating in the air shaking a fairy to sprinkle pixie dust on something so that it will float, too.

PR #995, opened 2024-02-14
docs: Add sampler default intervals to docs

### LINKED ISSUE #992: docs: document default intervals for samplers
**Steps to reproduce**

Customer says:

> I'm wondering what the default Refinery "window" is. I know that the [EMA Dynamic Sampler ](https://docs.honeycomb.io/manage-data-volume/refinery/sampling-methods/#ema-dynamic-sampler)has an AdjustmentInterval value that can be set. The docs have an example of "30s" or "1m" but it does not state what the default is.

**Additional context**

Defaults are in the dynsampler-go repository but should be documented in refinery docs.

MikeGoldsmith:
Updates refinery rules metadata to include the default intervals for each of the sampler types. `rulesmeta.md` was updated manually and the other files were re-generated using the conver tool.

- Closes #992 

- Update each sampler to include the default interval
  - [DynamicSampler](https://github.com/honeycombio/dynsampler-go/blob/main/avgsamplerate.go#L30) & [TotalThroughputSampler](https://github.com/honeycombio/dynsampler-go/blob/main/totalthroughput.go#L31) -> 30s
  - [EMADynamicSampler](https://github.com/honeycombio/dynsampler-go/blob/main/emasamplerate.go#L41) & [EMAThroughputSampler](https://github.com/honeycombio/dynsampler-go/blob/main/emathroughput.go#L38) -> 15s
  - [WindowedThroughputSampler](https://github.com/honeycombio/dynsampler-go/blob/main/windowedthroughput.go#L34) -> 1s

### DIFF
--- a/config/metadata/rulesMeta.yaml
+++ b/config/metadata/rulesMeta.yaml
@@ -67,6 +67,7 @@ groups:
         description: >
           The duration after which the Dynamic Sampler should reset its internal
           counters. It should be specified as a duration string. For example, "30s" or "1m".
+          Defaults to "30s".
       - name: FieldList
         type: stringarray
         validations:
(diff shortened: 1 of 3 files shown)


kentquirk (approved):
Awww, you rock. Thank you. After this, the generated file needs to get copied to docs as well in a docs pr.

Issue #970, opened 2024-01-19
Config bools that default to true require more work


fchikwekwe:
In #954, we discovered that booleans that default to true were being overridden when applying default values. When boolean values are processed by the defaults library, there is not way to distinguish between a `false` that was intentionally set and a zero boolean value. This results in the default boolean value being used (`true` in this case) making it not possible for a boolean that defaults to `true` to be set to `false`. 

We fixed this by setting a pointer to a boolean value so that there are three possible states for a boolean: `true`, `false` and `nil`. If we send a `nil` boolean to the defaults library it is properly interpreted as unset and can be updated to `false`. 

There is some technical debt leftover from this fix: 

- [x] add a getter function for `HoneycombLoggerConfig.SamplerEnabled` 
- [ ] add a `default-true` type in order to avoid hardcoding true values in getter functions for related fields. 
- [ ] use something (reflect?) to do struct tag lookups to find default values in the getter functions instead of declaring defaults in multiple places

fchikwekwe:
@robbkidd Already added the `GetSamplerEnabled` getter function in #969. I suggested in that PR that perhaps getter functions for the other fields on `HoneycombLoggerConfig` should be added as well.

fchikwekwe:
So, I did the work of trying to look up struct tags values using reflect in https://github.com/honeycombio/refinery/pull/996, but the consensus is that its wasteful to look up struct tags when we can instead put effort towards creating a `defaultTrue` type. I will try and work on adding that type.

PR #996, opened 2024-02-16
refactor: look up default values on struct tag

fchikwekwe:

- resolves part of #970 

- add a struct tag lookup to getter funcs for fields that default to boolean true


config/mock.go
```diff
@@ -423,11 +429,11 @@ func (m *MockConfig) GetIsDryRun() bool {
 	return m.DryRun
 }
 
-func (m *MockConfig) GetAddHostMetadataToTrace() bool {
+func (m *MockConfig) GetAddHostMetadataToTrace() (addHostMetadataToTrace bool, err error) {
 	m.Mux.RLock()
 	defer m.Mux.RUnlock()
 
-	return m.AddHostMetadataToTrace
+	return m.AddHostMetadataToTraceVal, m.AddHostMetadataToTraceErr
```
VinozzZ, line 436: If we are changing `return m.AddHostMetadataToTrace` to use `m.AddHostMetadataToTraceVal` as the return value, we will need to assign `AddHostMetadataToTraceVal` to true in `newStartedApp` in `app_test.go`

config/mock.go
```diff
@@ -75,13 +77,17 @@ type MockConfig struct {
 	DryRun                           bool
 	DryRunFieldName                  string
 	AddHostMetadataToTrace           bool
```
VinozzZ, line 79: If we choose to use `AddHostMetadataToTraceVal`, we probably should delete `AddHostMetadataToTrace`.

fchikwekwe: That's fair. I was trying to limit my changes a bit so I introduced some duplication here, but if I get the build passing I will try to not duplicate these values.

fchikwekwe: Resolved this with the build fix. Thanks for the help!

VinozzZ (commented):
Yahoo for fixing the default tag to true issue. I'm not certain that it's a good idea to parse the default tag each time when we try to read the value though. It feels wasted work for me. It would be nice if we could get the default value only once. I think that's maybe what you are working on in a following PR?

I'm thinking maybe it's worth taking the time to see whether using a pointer to represent the three states route works

fchikwekwe:
@VinozzZ Yeah I just spoke with @kentquirk and he agreed that its maybe not the best idea to lookup struct tags for every getter func call. There is an option to cache the struct tag values to maybe make this better. But I think the consensus here is that using a `defaultTrue` type would be the best solution. I'll open a separate PR for that work.

PR #998, opened 2024-02-17
refactor: add default true type

### LINKED ISSUE #970: Config bools that default to true require more work

fchikwekwe:

- Closes #970 

- adds a `DefaultTrue` type to use when specifying values with a default of boolean `true` this improves some getter functions by abstracting away the `*bool` behavior and allows boolean `true` be a true default

### DIFF
--- a/config/validate.go
+++ b/config/validate.go
@@ -184,6 +184,15 @@ func validateDatatype(k string, v any, typ string) string {
 		if u.Scheme != "http" && u.Scheme != "https" {
 			return fmt.Sprintf("field %s (%v) must use an http or https scheme", k, v)
 		}
+	case "defaulttrue":
+		if !isString(v) {
+			return fmt.Sprintf("field %s must be of type DefaultTrue", k)
+		}
+		var dt DefaultTrue
+		err := dt.UnmarshalText([]byte(v.(string)))
+		if err != nil {
+			return fmt.Sprintf("field %s (%v) must be a valid boolean which can be converted to DefaultTrue", k, v)
+		}
 	default:
 		panic("unknown data type " + typ)
 	}
--- a/config/validate.go
+++ b/config/validate.go
@@ -189,11 +189,13 @@ func validateDatatype(k string, v any, typ string) string {
 		case bool:
 			return ""
 		case string:
-			if val != "t" && val != "f" && val != "true" && val != "false" {
+			switch strings.ToLower(val) {
+			case "t", "true", "f", "false":
+			default:
 				return fmt.Sprintf("field %s (%v) must be 'true', 'false', 't', or 'f'", k, v)
 			}
 		default:
-			return fmt.Sprintf("field %s (%v) must be 'true', 'false', 't', or 'f'", k, v)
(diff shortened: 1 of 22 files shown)


config/validate.go
```diff
@@ -184,6 +184,15 @@ func validateDatatype(k string, v any, typ string) string {
 		if u.Scheme != "http" && u.Scheme != "https" {
 			return fmt.Sprintf("field %s (%v) must use an http or https scheme", k, v)
 		}
+	case "defaulttrue":
+		if !isString(v) {
+			return fmt.Sprintf("field %s must be of type DefaultTrue", k)
```
kentquirk, line 189: This error message wouldn't be generated *unless* the field was of type DefaultTrue. This validation is checking the value portion of the configuration, so you actually want to check `if _, ok := v.(bool)` to make sure the user is specifying either true or false for the value.

I'm not sure you even need the UnmarshalText part below. But to confirm, you should try a couple of different values in a configuration and run the validator to make sure it does the right thing.

fchikwekwe: This value comes in as a string, so I used `strconv.ParseBool` instead.

fchikwekwe: According the docs, it does not seem that 1/0 are supported as bools yaml. I'll fix the validation to check for "true"/"false", "t"/"f".

fchikwekwe: Its a little ugly, but I've updated it.

config/validate.go
```diff
@@ -184,6 +184,17 @@ func validateDatatype(k string, v any, typ string) string {
 		if u.Scheme != "http" && u.Scheme != "https" {
 			return fmt.Sprintf("field %s (%v) must use an http or https scheme", k, v)
 		}
+	case "defaulttrue":
+		switch val := v.(type) {
+		case bool:
+			return ""
+		case string:
+			if val != "t" && val != "f" && val != "true" && val != "false" {
```
kentquirk, line 192: Please use `strings.ToLower(val)`. Another way to do this if you prefer would be something like:
```
switch strings.ToLower(val) {
case "t", "true", "f", "false":
default:
    // error
```

fchikwekwe: Sure, the second switch statement looks better to me.

kentquirk (approved):
Looks good!

PR #999, opened 2024-02-20
Faith/root span optimization

fchikwekwe:

-

-


kentquirk:
Supplanted by #1018

PR #1000, opened 2024-02-20
docs: update refinery docs

fchikwekwe:

- Update docs in prep for release

### DIFF
--- a/refinery_rules.md
+++ b/refinery_rules.md
@@ -499,6 +499,11 @@ The field to check.
 This can name any field in the trace.
 If the field is not present, then the condition will not match.
 The comparison is case-sensitive.
+The field can also include a prefix that changes the span used for evaluation of the field.
+The only prefix currently supported is `root`, as in `root.http.status`.
+Specifying `root.`  causes the condition to be evaluated against the root span.
+For example, if the `Field` is `root.url`, then the condition will be processed using the url  field from the root span.
+The setting `Scope: span` for a rule does not change the meaning of this prefix -- the condition is still evaluated on the root span and is treated as if it were part of the   span being processed.
 
 - Type: `string`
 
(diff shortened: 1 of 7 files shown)


refinery_rules.md
```diff
@@ -499,6 +499,11 @@ The field to check.
 This can name any field in the trace.
 If the field is not present, then the condition will not match.
 The comparison is case-sensitive.
+The field can also include a prefix that changes the span used for evaluation of the field.
+The only prefix currently supported is `root`, as in `root.http.status`.
+Specifying `root.`  causes the condition to be evaluated against the root span.
```
mjingle, line 504: Remove extra space here.
```suggestion
Specifying `root.` causes the condition to be evaluated against the root span.
```

refinery_rules.md
```diff
@@ -499,6 +499,11 @@ The field to check.
...
+For example, if the `Field` is `root.url`, then the condition will be processed using the url  field from the root span.
```
mjingle, line 505: Remove extra space here.
```suggestion
For example, if the `Field` is `root.url`, then the condition will be processed using the url field from the root span.
```

refinery_rules.md
```diff
@@ -499,6 +499,11 @@ The field to check.
...
+The setting `Scope: span` for a rule does not change the meaning of this prefix -- the condition is still evaluated on the root span and is treated as if it were part of the   span being processed.
```
mjingle, line 506: Remove extra space here. 
```suggestion
The setting `Scope: span` for a rule does not change the meaning of this prefix -- the condition is still evaluated on the root span and is treated as if it were part of the  span being processed.
```

rules.md
```diff
@@ -538,6 +538,11 @@ The field to check.
 This can name any field in the trace.
 If the field is not present, then the condition will not match.
 The comparison is case-sensitive.
+The field can also include a prefix that changes the span used for evaluation of the field.
+The only prefix currently supported is `root`, as in `root.http.status`.
+Specifying `root.`  causes the condition to be evaluated against the root span.
+For example, if the `Field` is `root.url`, then the condition will be processed using the url  field from the root span.
+The setting `Scope: span` for a rule does not change the meaning of this prefix -- the condition is still evaluated on the root span and is treated as if it were part of the   span being processed.
```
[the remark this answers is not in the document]
mjingle: @fchikwekwe Do we have to make this spacing fix at the root file where this content is generated from?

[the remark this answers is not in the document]
fchikwekwe: @mjingle good call. I'll update the RulesMeta.yml file with this change.

fchikwekwe:
@mjingle I think it should be up to date now.

mjingle (approved):
Looks great to me! Thank you so much @fchikwekwe!

TylerHelmuth (approved):

PR #1001, opened 2024-02-20
chore: prep refinery 2.4 release

fchikwekwe:

- Prep release for Refinery 2.4

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -12,7 +12,7 @@ fixes, improvements, and dependency updates.
 Users can now specify rules that match only the root span of a trace (i.e. `root.http.status`). 
 
 ### Notable Fixes 
-* Previously, rules with a default of boolean `true` would be overridden when defaults were applied to the config. We have fixed this by using the `*bool` type for these values as well as adding helper functions to avoid strange behavior related to how booleans work in Go. 
+* Previously, rules with a default of boolean `true` that we set to `false` by configuration would be overridden back to `true` when defaults were applied to the config. We have fixed this by using the `*bool` type for these values as well as adding helper functions to avoid strange behavior related to how booleans work in Go. 
 * Use of a generics set to avoid complex map patterns. Behavior mostly borrowed from another Honeycomb repository.
 
 ## Version 2.3.0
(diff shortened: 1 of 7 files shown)


RELEASE_NOTES.md
```diff
@@ -1,6 +1,19 @@
...
+
+This release includes an update to allow users to specify root span context in their rules. It also includes some bug 
+fixes, improvements, and dependency updates.
+
+### Root Span Context
+
+Users can now specify rules that match only the root span of a trace (i.e. `root.http.status`). 
+
+### Notable Fixes 
+* Previously, rules with a default of boolean `true` would be overridden when defaults were applied to the config. We have fixed this by using the `*bool` type for these values as well as adding helper functions to avoid strange behavior related to how booleans work in Go.
```
TylerHelmuth, line 15:
```suggestion
* Previously, rules with a default of boolean `true` that we set to `false` by configuration would be overridden back to `true` when defaults were applied to the config. We have fixed this by using the `*bool` type for these values as well as adding helper functions to avoid strange behavior related to how booleans work in Go. 
```

RELEASE_NOTES.md
```diff
@@ -1,6 +1,19 @@
...
+This release includes an update to allow users to specify root span context in their rules. It also includes some bug 
+fixes, improvements, and dependency updates.
+
+### Root Span Context
+
+Users can now specify rules that match only the root span of a trace (i.e. `root.http.status`). 
+
+### Notable Fixes 
+* Previously, rules with a default of boolean `true` would be overridden when defaults were applied to the config. We have fixed this by using the `*bool` type for these values as well as adding helper functions to avoid strange behavior related to how booleans work in Go. 
+* Use of a generics set to avoid complex map patterns. Behavior mostly borrowed from another Honeycomb repository.
```
TylerHelmuth, line 16: If this change isn't user facing I feel we can exclude it from the RELEASE_NOTES and leave it in CHANGELOG

TylerHelmuth (approved):

RELEASE_NOTES.md
fchikwekwe: Done!

PR #1006, opened 2024-02-22
fix: handle root prefix when no root span on trace

fchikwekwe:

- Handle case where `root.` prefix is used but a root span is not present for the trace at the time a sampling decision is being made, for example: a late-arriving root span.

- Check for root span when trying to use `root.` prefix; document current behavior when root span is not present on trace

### DIFF
--- a/config.md
+++ b/config.md
@@ -1,7 +1,7 @@
 # Honeycomb Refinery Configuration Documentation
 
 This is the documentation for the configuration file for Honeycomb's Refinery.
-It was automatically generated on 2024-02-20 at 18:13:44 UTC.
+It was automatically generated on 2024-02-22 at 18:03:10 UTC.
 
 ## The Config file
 
@@ -200,7 +200,7 @@ This value is available to the rules-based sampler, making it possible to write 
 If `true` and `AddCountsToRoot` is set to false, then Refinery will add `meta.span_count` to the root span.
 
 - Eligible for live reload.
-- Type: `bool`
+- Type: `defaulttrue`
 - Default: `true`
 
 ### `AddCountsToRoot`
@@ -223,7 +223,7 @@ AddHostMetadataToTrace specifies whether to add host metadata to traces.
 If `true`, then Refinery will add the following tag to all traces: - `meta.refinery.local_hostname`: the hostname of the Refinery node
 
 - Eligible for live reload.
-- Type: `bool`
+- Type: `defaulttrue`
 - Default: `true`
 
 ## Traces
@@ -420,7 +420,7 @@ The sample rate is controlled by the `SamplerThroughput` setting.
 The sampler used throttles the rate of logs sent to Honeycomb from any given source within Refinery -- it should effectively limit the rate of redundant messages.
 
 - Not eligible for live reload.
-- Type: `bool`
+- Type: `defaulttrue`
 - Default: `true`
 
 ### `SamplerThroughput`
@@ -917,7 +917,7 @@ If it costs money to transmit data between Refinery instances (for example, when
 The option to disable it is provided as an escape hatch for deployments that value lower CPU utilization over data transfer costs.
 
 - Not eligible for live reload.
-- Type: `bool`
+- Type: `defaulttrue`
 - Default: `true`
 
 ### `AdditionalAttributes`
@@ -969,7 +969,8 @@ Enabled specifies whether the gRPC server is enabled.
 If `false`, then the gRPC server is not started and no gRPC traffic is accepted.
 
 - Not eligible for live reload.
-- Type: `bool`
+- Type: `defaulttrue`
+- Default: `true`
 
 ### `ListenAddr`
 
--- a/config.md
+++ b/config.md
@@ -1,7 +1,7 @@
 # Honeycomb Refinery Configuration Documentation
 
 This is the documentation for the configuration file for Honeycomb's Refinery.
-It was automatically generated on 2024-02-22 at 18:03:10 UTC.
+It was automatically generated on 2024-02-23 at 17:42:44 UTC.
 
 ## The Config file
 
--- a/config/metadata/rulesMeta.yaml
+++ b/config/metadata/rulesMeta.yaml
@@ -641,7 +641,7 @@ groups:
           named field that contains a value will be used for the condition. Only
           the first populated field will be used, even if the condition fails.
           
-          If a `root.` prefix is present on a field but the root span is not on
+          If a `root.` prefix is present on a field, but the root span is not on
           the trace, that field will be skipped.
 
           If none of the fields are present, then the condition will not match.
--- a/config/metadata/rulesMeta.yaml
+++ b/config/metadata/rulesMeta.yaml
@@ -625,7 +625,7 @@ groups:
           span being processed.
           
           When using the `not-exists` operator to check a condition with a `root.` prefix, if the root
-          span does not exist on the trace, then the condition will evaluate to false.
+          span does not exist on the trace, as in the case when a root span is taking a long time to complete and does not arrive before a sampling decision is being made, then the condition will evaluate to false.
 
       - name: Fields
         type: stringarray
--- a/config/metadata/rulesMeta.yaml
+++ b/config/metadata/rulesMeta.yaml
@@ -625,7 +625,7 @@ groups:
           span being processed.
           
           When using the `not-exists` operator to check a condition with a `root.` prefix, if the root
-          span does not exist on the trace then the condition will evaluate to false.
+          span does not exist on the trace, then the condition will evaluate to false.
 
       - name: Fields
         type: stringarray
--- a/config/metadata/rulesMeta.yaml
+++ b/config/metadata/rulesMeta.yaml
@@ -624,8 +624,9 @@ groups:
           the condition is still evaluated on the root span and is treated as if it were part of the 
           span being processed.
           
-          When using the `not-exists` operator to check a condition with a `root.` prefix, if the root
-          span does not exist on the trace, as in the case when a root span is taking a long time to complete and does not arrive before a sampling decision is being made, then the condition will evaluate to false.
+          When using the `root.` prefix on a field with a `not-exists` operator, include the `has-root-span: true` condition in the rule.
+          The `not-exists` condition on a `root.`-prefixed field will evaluate to false if the existence of the root span is not checked and the root span does not exist.
+          The primary reason a root span is not present on a trace when a sampling decision is being made is when the root span takes longer to complete than the configured TraceTimeout.
 
       - name: Fields
         type: stringarray
--- a/config/metadata/rulesMeta.yaml
+++ b/config/metadata/rulesMeta.yaml
@@ -615,14 +615,18 @@ groups:
           
           The field can also include a prefix that changes the span used for evaluation of the field.
           The only prefix currently supported is `root`, as in `root.http.status`. Specifying `root.`
-          causes the condition to be evaluated against the root span. 
+          causes the condition to be evaluated against the root span.
           
           For example, if the `Field` is `root.url`, then the condition will be processed using the url
-          field from the root span. 
+          field from the root span.
           
           The setting `Scope: span` for a rule does not change the meaning of this prefix --
           the condition is still evaluated on the root span and is treated as if it were part of the 
-          span being processed. 
+          span being processed.
+          
+          When using the `not-exists` operator to check a condition with a `root.` prefix, if the root
+          span does not exist on the trace then the condition will evaluate to false.
+
       - name: Fields
         type: stringarray
         valuetype: stringarray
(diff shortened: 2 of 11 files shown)


config.md
```diff
@@ -200,7 +200,7 @@ This value is available to the rules-based sampler, making it possible to write
 If `true` and `AddCountsToRoot` is set to false, then Refinery will add `meta.span_count` to the root span.
 
 - Eligible for live reload.
-- Type: `bool`
+- Type: `defaulttrue`
```
fchikwekwe, line 203: Weird that this change did not update the last time I generated docs.

config/metadata/rulesMeta.yaml
```diff
@@ -615,14 +615,18 @@ groups:
...
+          field from the root span.
           
           The setting `Scope: span` for a rule does not change the meaning of this prefix --
           the condition is still evaluated on the root span and is treated as if it were part of the 
-          span being processed. 
+          span being processed.
+          
+          When using the `not-exists` operator to check a condition with a `root.` prefix, if the root
+          span does not exist on the trace then the condition will evaluate to false.
+
```
fchikwekwe, line 629: This is part of the main documentation change.

config/metadata/rulesMeta.yaml
```diff
@@ -636,6 +640,9 @@ groups:
           trace. The fields are checked in the order defined here, and the first
           named field that contains a value will be used for the condition. Only
           the first populated field will be used, even if the condition fails.
+          
+          If a `root.` prefix is present on a field but the root span is not on
+          the trace, that field will be skipped.
```
fchikwekwe, line 645: This is part of the main documentation change.

config/metadata/rulesMeta.yaml
```diff
@@ -636,6 +640,9 @@ groups:
           trace. The fields are checked in the order defined here, and the first
           named field that contains a value will be used for the condition. Only
           the first populated field will be used, even if the condition fails.
+          
+          If a `root.` prefix is present on a field but the root span is not on
```
mjingle, line 644:
```suggestion
          If a `root.` prefix is present on a field, but the root span is not on
```

mjingle (approved):
Docs side LGTM! 👍

sample/rules.go
```diff
@@ -223,7 +223,8 @@ func ruleMatchesSpanInTrace(trace *types.Trace, rule *config.RulesBasedSamplerRu
 	return false
 }
 
-func extractValueFromSpan(trace *types.Trace, span *types.Span, condition *config.RulesBasedSamplerCondition, checkNestedFields bool) (interface{}, bool) {
+// exists does it refer to valid cond field present
```
robbkidd, line 226:
```suggestion
// extractValueFromSpan extracts from a span the value found at the first of the given condition's fields found on a span.
// It returns the extracted value and an exists boolean indicating whether any of the condition's fields are present on the span.
```

sample/rules_test.go
```diff
@@ -2500,7 +2500,154 @@ func TestRulesRootSpanContext(t *testing.T) {
...
+						Data: map[string]interface{}{
+							"trace_id": "abc123",
+							"foo":      true,
+						},
+					},
+				},
+			},
+			ExpectedKeep: true,
+			ExpectedRate: 1,
+			ExpectedName: "no rule matched",
```
robbkidd, line 2650: Now that I see it, I don't think I understand the spirit of this test case we talked about together. 😕

config/metadata/rulesMeta.yaml
```diff
@@ -615,14 +615,18 @@ groups:
...
-          field from the root span. 
+          field from the root span.
           
           The setting `Scope: span` for a rule does not change the meaning of this prefix --
           the condition is still evaluated on the root span and is treated as if it were part of the 
-          span being processed. 
+          span being processed.
+          
+          When using the `not-exists` operator to check a condition with a `root.` prefix, if the root
+          span does not exist on the trace, then the condition will evaluate to false.
```
robbkidd, line 628: What do you think of including a common reason why a root span would not exist on the trace in Refinery?

```suggestion
          span does not exist on the trace, as in the case when a root span is taking a long time to complete and does not arrive before a sampling decision is being made, then the condition will evaluate to false.
```

robbkidd (changes requested):
A couple documentation wording suggestions and I'm confused about our new test case.

rules.md
```diff
@@ -540,9 +540,10 @@ If the field is not present, then the condition will not match.
 The comparison is case-sensitive.
 The field can also include a prefix that changes the span used for evaluation of the field.
 The only prefix currently supported is `root`, as in `root.http.status`.
-Specifying `root.`  causes the condition to be evaluated against the root span.
-For example, if the `Field` is `root.url`, then the condition will be processed using the url  field from the root span.
-The setting `Scope: span` for a rule does not change the meaning of this prefix -- the condition is still evaluated on the root span and is treated as if it were part of the   span being processed.
+Specifying `root.` causes the condition to be evaluated against the root span.
+For example, if the `Field` is `root.url`, then the condition will be processed using the url field from the root span.
+The setting `Scope: span` for a rule does not change the meaning of this prefix -- the condition is still evaluated on the root span and is treated as if it were part of the  span being processed.
+When using the `not-exists` operator to check a condition with a `root.` prefix, if the root span does not exist on the trace then the condition will evaluate to false.
```
kentquirk, line 546: I don't think this is true, or should be reworded.

If the root span does not exist, then the `exists` variable will be false, so `not-exists` will be true.

```suggestion
When using the `exists` or `not-exists` operators to check a field containing a `root.` prefix, if the trace has no root span, it will be treated as if the field did not exist.
```

rules.md
```diff
@@ -552,6 +553,7 @@ An array of field names to check.
 These can name any field in the trace.
 The fields are checked in the order defined here, and the first named field that contains a value will be used for the condition.
 Only the first populated field will be used, even if the condition fails.
+If a `root.` prefix is present on a field but the root span is not on the trace, that field will be skipped.
```
kentquirk, line 556:
```suggestion
If a `root.` prefix is present on a field but the trace does not contain a root span, that field will be skipped.
```

sample/rules.go
```diff
@@ -233,13 +234,16 @@ func extractValueFromSpan(trace *types.Trace, span *types.Span, condition *confi
...
 		// check if rule uses root span context
 		if strings.HasPrefix(field, RootPrefix) {
-			field = field[len(RootPrefix):]
-			span = trace.RootSpan
+			// make sure root span exists
+			if trace.RootSpan != nil {
+				field = field[len(RootPrefix):]
+				span = trace.RootSpan
+			} else {
+				continue
```
kentquirk, line 245:
```suggestion
				// we wanted root span but this trace doesn't have one, so just skip it
				continue
```

sample/rules_test.go
kentquirk: What we wanted here was to be able to tell the difference between "my root span is missing the service name" and "my root span is missing entirely".

We could do this with a two-condition test:

Condition:
    Field: root.trace_id
    Operator: exists
Condition:
    Field: root.service_name
    Operator: not-exists

That test will be true if the root span exists but the service name is missing, and false if the root span is missing entirely.

kentquirk (approved):
I think the code is correct but I'd like to fix that one test.

sample/rules.go
robbkidd: Surely we should write the comment as "When a nil root comes along, you must skip it. Skip it good." 😀

kentquirk: OK, I would approve a PR with that comment in a heartbeat. Wearing a lampshade and a shiny jacket.

sample/rules_test.go
kentquirk: OMG -- I forgot we added `has-root-span`!

My scenario is now kind of silly, but it doesn't hurt to have it.

robbkidd: I don't think the scenario is silly! I think putting a test case with `has-root-span true` + `root.x not-exists` is the test confirmation for the advice we should give in the docs: don't use `root.x not-exists` in a rule without also including a `has-root-span true` condition or you're gonna have a bad time.

"Gonna have a bad time" ski instructor says "That's right!"

sample/rules_test.go
```diff
@@ -2500,7 +2500,232 @@ func TestRulesRootSpanContext(t *testing.T) {
...
+		},
+		{
+			Rules: &config.RulesBasedSamplerConfig{
+				Rules: []*config.RulesBasedSamplerRule{
+					{
+						Name:       "NotExists🦶🔫/no root span, no has-root-span guard (not recommended!)",
```
robbkidd, line 2698: Added emoji for fun.

Added the slash to sub-categorize these tests in the test suite run output:

In a fancy terminal:
Screenshot 2024-02-23 at 9 47 35 AM

In VS Code's test runner:
Screenshot 2024-02-23 at 9 48 23 AM

sample/rules_test.go
```diff
@@ -2500,7 +2500,232 @@ func TestRulesRootSpanContext(t *testing.T) {
...
+			},
+			ExpectedKeep: false,
+			ExpectedRate: 10,
+			ExpectedName: "no root span, checking other spans in trace",
+		},
+		{
+			Rules: &config.RulesBasedSamplerConfig{
+				Rules: []*config.RulesBasedSamplerRule{
+					{
+						Name:       "NotExists🦶🔫/root span exists, rule uses has-root-span guard, field is missing",
```
robbkidd, line 2618: Added emoji for fun.

Added the slash to sub-categorize these tests in the test suite run output:

In a fancy terminal:
Screenshot 2024-02-23 at 9 47 35 AM

In VS Code's test runner:
Screenshot 2024-02-23 at 9 48 23 AM

fchikwekwe: This is a good add. Appreciate the categorization as well.

config/metadata/rulesMeta.yaml
```diff
@@ -615,14 +615,18 @@ groups:
...
-          field from the root span. 
+          field from the root span.
           
           The setting `Scope: span` for a rule does not change the meaning of this prefix --
           the condition is still evaluated on the root span and is treated as if it were part of the 
-          span being processed. 
+          span being processed.
+          
+          When using the `not-exists` operator to check a condition with a `root.` prefix, if the root
+          span does not exist on the trace, as in the case when a root span is taking a long time to complete and does not arrive before a sampling decision is being made, then the condition will evaluate to false.
```
robbkidd, line 628: With our (re)discovery of the `has-root-span` condition, I think we can include that as a recommended pairing with any `root.something exists/not-exists` condition. These words are being added to the `Field` docs; I wonder if these not-exists notes should also/or-instead appear under the not-exists docs?

Here's my go at more word-smithing:

```suggestion
          When using the `root.` prefix on a field with a `not-exists` operator, include the `has-root-span: true` condition in the rule.
          The `not-exists` condition on a `root.`-prefixed field will evaluate to false if the existence of the root span is not checked and the root span does not exist.
          The primary reason a root span is not present on a trace when a sampling decision is being made is when the root span takes longer to complete than the configured TraceTimeout.
```

kentquirk: Love it!

robbkidd:
@mjingle We've done more doc word-smithing, so this could use a re-review. 😅

rules.md
```diff
@@ -540,9 +540,12 @@ If the field is not present, then the condition will not match.
 The comparison is case-sensitive.
 The field can also include a prefix that changes the span used for evaluation of the field.
 The only prefix currently supported is `root`, as in `root.http.status`.
-Specifying `root.`  causes the condition to be evaluated against the root span.
-For example, if the `Field` is `root.url`, then the condition will be processed using the url  field from the root span.
-The setting `Scope: span` for a rule does not change the meaning of this prefix -- the condition is still evaluated on the root span and is treated as if it were part of the   span being processed.
+Specifying `root.` causes the condition to be evaluated against the root span.
+For example, if the `Field` is `root.url`, then the condition will be processed using the url field from the root span.
+The setting `Scope: span` for a rule does not change the meaning of this prefix -- the condition is still evaluated on the root span and is treated as if it were part of the  span being processed.
```
rkhoriander, line 545:
```suggestion
The setting `Scope: span` for a rule does not change the meaning of this prefix -- the condition is still evaluated on the root span and is treated as if it were part of the span being processed.
```

sample/rules.go
```diff
@@ -223,7 +223,9 @@ func ruleMatchesSpanInTrace(trace *types.Trace, rule *config.RulesBasedSamplerRu
 	return false
 }
 
-func extractValueFromSpan(trace *types.Trace, span *types.Span, condition *config.RulesBasedSamplerCondition, checkNestedFields bool) (interface{}, bool) {
+// extractValueFromSpan extracts from a span the value found at the first of the given condition's fields found on a span.
```
rkhoriander, line 226: "from a span"..."found on a span". This feels a bit vague. Is it the same span? Is it just duplicate info?

fchikwekwe: This is the span that gets passed into the function. I'll fix the wording so its not ambiguous.

PR #1007, opened 2024-02-23
fix: Fix a typo from PR #1006

rkhoriander:

Small typo in the sample wording when reviewing #1006 (but it had already been merged).

Moves one word into the correct place.


robbkidd (approved):
Yup! 👍🏻

PR #1011, opened 2024-02-25
feat: improve test_stores, rename some constants

kentquirk:
I think this is now basically usable.


kentquirk:
I'm going to merge this because the test is in an area I don't think I changed and this is merging to a branch. We'll fix it there.

PR #1013, opened 2024-02-27
feat: add tracing

kentquirk:
This adds tracing to the test_stores system -- you can send results to honeycomb.


cmd/test_stores/main.go
```diff
@@ -408,15 +508,26 @@ func (fri *FakeRefineryInstance) runProcessor(opts CmdLineOptions, nodeIndex int
...
+			_, span := fri.tracer.Start(ctx, "get_status_for_traces")
+			addSpanField(span, "num_trace_ids", len(tids))
 			statuses, err := fri.store.GetStatusForTraces(tids)
 			if err != nil {
-				fmt.Println(err)
+				fri.addException(span, err)
+				span.End()
 				continue
 			}
+			span.End()
```
MikeGoldsmith, line 529: Using `defer span.End()` would save you from having to call end in both cases here

cmd/test_stores/main.go
```diff
@@ -261,26 +326,35 @@ func (t *traceGenerator) generateTrace(out chan *centralstore.CentralSpan, stop
...
 			// only send the span if it's ours
 			if t.mySpan(traceid, trace.Spans[counter-1].SpanID) {
-				fmt.Printf(" sending %d tid: %s spid: %s\n", counter-1, trace.TraceID, trace.Spans[counter-1].SpanID)
+				addSpanField(span, "sent", true)
 				out <- trace.Spans[counter-1]
 			} else {
-				fmt.Printf("skipping %d tid: %s spid: %s\n", counter-1, trace.TraceID, trace.Spans[counter-1].SpanID)
+				addSpanField(span, "sent", false)
 			}
+			span.End()
```
MikeGoldsmith, line 342: Normally better to `defer span.End()` 😄

centralstore/smartwrapper.go
```diff
@@ -34,13 +36,14 @@ type SmartWrapperOptions struct {
 }
 
 // NewSmartWrapper creates a new SmartWrapper.
-func NewSmartWrapper(options SmartWrapperOptions, basic BasicStorer) *SmartWrapper {
+func NewSmartWrapper(options SmartWrapperOptions, basic BasicStorer, tracer trace.Tracer) *SmartWrapper {
```
MikeGoldsmith, line 39: Would it make sense to add the tracer to the `SmartWrapperOptions` and default as `noopTracer`?

MikeGoldsmith (approved):
Looks good - left some suggestions around ending spans and maybe moving tracer into options struct

cmd/test_stores/main.go
kentquirk: same as above -- that wouldn't work here because the function is in an infinite loop.

centralstore/smartwrapper.go
kentquirk: Two things:
* SmartWrapper doesn't use the tracer, the test program does so it doesn't really belong there.
* I think of the individual SmartWrapperOptions as things you can specify in configuration text (not complex types). I expect them to move into config eventually.

cmd/test_stores/main.go
MikeGoldsmith: I thought the defer would trigger as part of the case statement evaluation?

kentquirk: [Defers are only called in reverse order after the function containing them returns.](https://go.dev/blog/defer-panic-and-recover)

PR #1014, opened 2024-02-27
feat: properly handle keep state and timestamp for state transition

VinozzZ:
- Keep state
    - this PR makes sure that the keep decision is stored as part of the state transition for a given trace
    - It also adds logic to remove spans information for a given trace when a decision is made
- Timestamp
    - we nee record the current time whenever a state change happens.
    - the precision used before is not sufficient to determine the order of state change events
    - this PR changes the timestamp precision from second to nanosecond

- Modified `KeepTrace`
- return traceIDs that has successfully changed their states from `toNextState` function
- added more tests

### DIFF
--- a/centralstore/redis_store.go
+++ b/centralstore/redis_store.go
@@ -36,12 +36,12 @@ type RedisBasicStoreOptions struct {
 	MaxTraceRetention time.Duration
 }
 
-var _ BasicStorer = (*RedisBasicStore)(nil)
-
-func NewRedisBasicStore(opt RedisBasicStoreOptions) *RedisBasicStore {
-	host := "localhost:6379"
-	if opt.Host != "" {
-		host = opt.Host
+func EnsureRedisBasicStoreOptions(opt *RedisBasicStoreOptions) {
+	if opt == nil {
+		opt = &RedisBasicStoreOptions{}
+	}
+	if opt.Host == "" {
+		opt.Host = "localhost:6379"
 	}
 
 	if opt.MaxTraceRetention == 0 {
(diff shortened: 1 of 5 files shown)


centralstore/redis_store.go
```diff
@@ -39,15 +39,27 @@ type RedisBasicStoreOptions struct {
 var _ BasicStorer = (*RedisBasicStore)(nil)
 
 func NewRedisBasicStore(opt RedisBasicStoreOptions) *RedisBasicStore {
-	var host string
-	if opt.Host == "" {
-		host = "localhost:6379"
+	host := "localhost:6379"
```
kentquirk, line 42: This pattern of setting defaults in the New() function is probably not the way we want to do it -- we want the defaults set by the client, not by the BasicStore. Ideally they're set in the config package, but you could, for now, put them in RedisBasicStoreOptions (which will move to the config package eventually).

centralstore/redis_store.go
```diff
@@ -113,10 +125,13 @@ func (r *RedisBasicStore) WriteSpan(span *CentralSpan) error {
 	case DecisionDrop:
 		return nil
 	case DecisionKeep, AwaitingDecision:
-		_ = r.traces.incrementSpanCounts(conn, span.TraceID, span.Type)
+		err := r.traces.incrementSpanCounts(conn, span.TraceID, span.Type)
+		if err != nil {
+			r.errs <- err
```
kentquirk, line 130: If r.errs is full because no one is reading from it, this will block forever. Perhaps you want to write to this chan as a select with a default that drops the error?

centralstore/redis_store.go
```diff
@@ -261,6 +279,7 @@ func (r *RedisBasicStore) ChangeTraceStatus(traceIDs []string, fromState, toStat
 	if toState == DecisionDrop {
 		traces, err := r.traces.getTraceStatuses(conn, traceIDs)
 		if err != nil {
+			r.errs <- err
```
kentquirk, line 282: Hmmmm...do we really want to BOTH have a channel of errors and an error return? I would have thought we'd do one or the other.

kentquirk (approved):
I ran out of time to do a complete review (I really didn't read the test code at all) but don't want to block things, so let's keep moving forward.

centralstore/redis_store.go
VinozzZ: Oops, you are right. That's a leftover. I removed it

PR #1018, opened 2024-02-29
feat: Shortcut evaluation of rules containing 'root.'

kentquirk:

The new `root.` rule can be shortcut in many cases -- if you only check the root span, then you don't have to check every span since it's going to have the same result.

This also uncovered a bug in the original logic -- in a situation where there were multiple entries in Fields with `root.` preceding an entry without it, we would have been testing the wrong span for the later entries.

- Add optimization
- Add a couple of new tests


MikeGoldsmith (approved):
Looks good to me 👍🏻

PR #1019, opened 2024-02-29
maint: Update Husky to handle invalid content type errors

MikeGoldsmith:
Updates Husky to v0.25.1 which includes fixes handling invalid content types for OTLP requests.

See the following PR for changes:
- https://github.com/honeycombio/husky/pull/243

**NOTE**: This PR is based on an unreleased version of husky. We'll wait for husky to be released before merging this fix.

- Updating Husky to 0.25.1
- Updating the error handler to use husky to write a text/plain response for invalid content type errors
- Move validation of API keys from config after request headers check

### DIFF
--- a/route/errors.go
+++ b/route/errors.go
@@ -77,22 +77,12 @@ func (r *Router) handlerReturnWithError(w http.ResponseWriter, he handlerError, 
 
 func (r *Router) handleOTLPFailureResponse(w http.ResponseWriter, req *http.Request, otlpErr husky.OTLPError) {
 	r.Logger.Error().Logf(otlpErr.Error())
-
-	if otlpErr != husky.ErrInvalidContentType {
-		err := husky.WriteOtlpHttpFailureResponse(w, req, otlpErr)
-		if err != nil {
-			// If we made it here we had a problem writing an OTLP HTTP response
-			resp := fmt.Sprintf("failed to write otlp http response, %v", err.Error())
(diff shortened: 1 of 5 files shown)


route/errors.go
```diff
@@ -77,22 +77,12 @@ func (r *Router) handlerReturnWithError(w http.ResponseWriter, he handlerError,
...
-			// If we made it here we had a problem writing an OTLP HTTP response
-			resp := fmt.Sprintf("failed to write otlp http response, %v", err.Error())
-			r.Logger.Error().Logf(resp)
-			w.Header().Set("Content-Type", "text/plain")
-			w.WriteHeader(http.StatusInternalServerError)
-			_, _ = io.WriteString(w, resp)
-		}
-	} else {
-		// Since we have an invalid content type for an OTLP HTTP response we choose to use to text/plain
+	if err := husky.WriteOtlpHttpFailureResponse(w, req, otlpErr); err != nil {
```
codeboten, line 80: is the error here always a content type error?

TylerHelmuth (approved):

route/errors.go
MikeGoldsmith: No, API key missing or invalid errors are also possible to come through here

codeboten: right, the description says it will use plain for content type errors, but this changes the behaviour for all errors?

PR #1020, opened 2024-02-29
fix: make sure decision keep and drop is persisted in redis

VinozzZ:

Before this PR, decisions made for traces were not stored in redis. It also wasn't part of the state machine when transition a trace from awaiting for decision to either decision keep or drop.

- Introduced DecisionDrop and DecisionKeep into the state machine
- check against in-memory cache before fetching from Redis for trace status
- run smartwrapper tests against redis in the ci
- removed some unnecessary code

### DIFF
--- a/centralstore/smartwrapper_test.go
+++ b/centralstore/smartwrapper_test.go
@@ -302,7 +302,7 @@ func TestSetTraceStatuses(t *testing.T) {
 
 	statuses := make([]*CentralTraceStatus, 0)
 	assert.EventuallyWithT(t, func(collect *assert.CollectT) {
-		statuses, err = store.GetStatusForTraces(traceids)
+		statuses, err = store.GetStatusForTraces(toDecide)
 		assert.NoError(collect, err)
 		assert.Equal(collect, numberOfTraces, len(statuses))
 		for _, state := range statuses {
--- a/centralstore/smartwrapper_test.go
+++ b/centralstore/smartwrapper_test.go
@@ -35,7 +35,7 @@ func noopTracer() trace.Tracer {
 	return pr.Tracer("test")
 }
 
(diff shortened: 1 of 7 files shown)


centralstore/smartwrapper_test.go
```diff
@@ -330,7 +331,7 @@ func TestSetTraceStatuses(t *testing.T) {
 		if status.TraceID == traceids[0] {
 			assert.Equal(t, DecisionKeep, status.State)
 			// TODO: save the reason in the reason cache correctly
```
kentquirk, line 333: remove this TODO, please

internal/redis/redis.go
```diff
@@ -574,6 +575,28 @@ func (c *DefaultConn) SetHash(key string, val interface{}) error {
...
+	}
+
+	args := redis.Args{key}.AddFlat(val)
+	for i := 1; i < len(args); i += 2 {
+		if err := c.conn.Send("HSETNX", key, args[i], args[i+1]); err != nil {
+			return nil, err
+		}
+	}
+
+	// TODO: values is always "OK", but we should be able to get the values
```
kentquirk, line 590: I don't understand what this TODO is saying?

centralstore/smartwrapper_test.go
```diff
@@ -433,14 +434,14 @@ func BenchmarkStoreGetTracesForState(b *testing.B) {
 	}
 }
 
-func cleanupRedisStore(store BasicStorer) error {
+func cleanupRedisStore(t testing.TB, store BasicStorer) error {
```
kentquirk, line 437: Given that you're logging and not returning an error, should probably remove the return parameter.

kentquirk (approved):
Some nits, but otherwise cool.

PR #1021, opened 2024-03-01
feat: parallelize decider and tracing support for basic store

VinozzZ:

- allow running decider in parallel
- tracing support for redis_store
- one round trip per trace state transition

- parallelize `GetTrace` call
- pass tracer into redis store
- running one lua script for a trace state transition

[Result honeycomb query
](https://ui.honeycomb.io/island-of-misfit-toys/environments/dynamic-scaling-refinery/datasets/refinery-store-test/result/AGisWg8rn49)


kentquirk (approved):
We've been talking in Slack, and I think this is good to go.

PR #1027, opened 2024-03-01
feat: Add metrics support to smartstore and cache and localremote

kentquirk:

- Beginning to put some metrics stuff in

- Implement GetMetrics for a few things

### DIFF
--- a/cmd/test_stores/main.go
+++ b/cmd/test_stores/main.go
@@ -185,5 +185,12 @@ func main() {
 			wg.Done()
 		}(i)
 	}
+	time.Sleep(10 * time.Second)
+	mets, err := store.GetMetrics()
+	if err != nil {
+		fmt.Println("Error getting metrics:", err)
+	} else {
+		fmt.Println("Metrics:", mets)
+	}
 	wg.Wait()
 }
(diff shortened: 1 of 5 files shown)


collect/cache/cuckooSentCache.go
```diff
@@ -285,3 +285,14 @@ func (c *cuckooSentCache) Resize(cfg config.SampleCacheConfig) error {
 	go c.monitor()
 	return nil
 }
+
+func (c *cuckooSentCache) GetMetrics() (map[string]interface{}, error) {
+	metrics := map[string]interface{}{
```
VinozzZ, line 290: Do we need a lock here before we get the counts?

kentquirk: No, because the counts are fetched from functions that take their own locks.

cmd/test_stores/main.go
```diff
@@ -185,5 +186,12 @@ func main() {
 			wg.Done()
 		}(i)
 	}
+	time.Sleep(10 * time.Second)
+	mets, err := store.GetMetrics(context.Background())
```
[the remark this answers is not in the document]
kentquirk: I mistakenly left that in -- it's not needed at all.

VinozzZ (approved):

PR #1033, opened 2024-03-04
feat: process state transitions in batches

VinozzZ:

Perviously, each trace state change needs to make a redis call. When client can't take advantage of parallelism since each call is serialized in redis store anyway. This PR makes it possible to process multiple traces in one redis call for state change.

**Before**
Screenshot 2024-03-04 at 6 47 53 PM

**After**
Screenshot 2024-03-04 at 6 48 22 PM

- Processing state transitions in a batch instead of one trace per redis call

### DIFF
--- a/centralstore/redis_store.go
+++ b/centralstore/redis_store.go
@@ -8,6 +8,7 @@ import (
 	"sort"
 	"time"
 
+	redigo "github.com/gomodule/redigo/redis"
 	"github.com/honeycombio/refinery/collect/cache"
 	"github.com/honeycombio/refinery/config"
 	"github.com/honeycombio/refinery/internal/otelutil"
@@ -125,7 +126,8 @@ func (r *RedisBasicStore) GetMetrics(ctx context.Context) (map[string]interface{
 }
 
 func (r *RedisBasicStore) WriteSpan(ctx context.Context, span *CentralSpan) error {
-	writeSpan := trace.SpanFromContext(ctx)
+	_, writeSpan := r.tracer.Start(ctx, "WriteSpan")
+	defer writeSpan.End()
 
 	conn := r.client.Get()
 	defer conn.Close()
@@ -773,7 +775,11 @@ func (t *traceStateProcessor) Stop() {
 func (t *traceStateProcessor) addNewTrace(ctx context.Context, conn redis.Conn, traceID string) error {
 	ctx, span := t.tracer.Start(ctx, "addNewTrace", trace.WithAttributes(attribute.KeyValue{Key: "trace_id", Value: attribute.StringValue(traceID)}))
 	defer span.End()
-	return t.applyStateChange(ctx, conn, newTraceStateChangeEvent(Unknown, Collecting), traceID)
+	_, err := t.applyStateChange(ctx, conn, newTraceStateChangeEvent(Unknown, Collecting), []string{traceID})
+	if err != nil {
+		return err
+	}
+	return nil
 }
 
 func (t *traceStateProcessor) stateByTraceIDsKey(state CentralTraceState) string {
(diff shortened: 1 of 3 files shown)


centralstore/redis_store.go
```diff
@@ -900,77 +894,108 @@ func (t *traceStateProcessor) removeExpiredTraces(client redis.Client) {
...
-func (t *traceStateProcessor) applyStateChange(ctx context.Context, conn redis.Conn, stateChange stateChangeEvent, traceID string) error {
+// applyStateChange runs a lua script that atomically moves traces between states and returns the trace IDs that has completed a state change.
+func (t *traceStateProcessor) applyStateChange(ctx context.Context, conn redis.Conn, stateChange stateChangeEvent, traceIDs []string) ([]string, error) {
 	ctx, span := t.tracer.Start(ctx, "applyStateChange")
 	defer span.End()
 
-	otelutil.AddSpanField(span, "trace_id", traceID)
+	otelutil.AddSpanField(span, "num_traces", len(traceIDs))
+
+	args := redigo.Args{validStateChangeEventsKey, stateChange.current.String(), stateChange.next.String(),
```
VinozzZ, line 904: TODO: I need to think about how to extract this redigo stuffs back into our internal redis package

centralstore/redis_store.go
```diff
@@ -773,7 +775,11 @@ func (t *traceStateProcessor) Stop() {
 func (t *traceStateProcessor) addNewTrace(ctx context.Context, conn redis.Conn, traceID string) error {
 	ctx, span := t.tracer.Start(ctx, "addNewTrace", trace.WithAttributes(attribute.KeyValue{Key: "trace_id", Value: attribute.StringValue(traceID)}))
 	defer span.End()
-	return t.applyStateChange(ctx, conn, newTraceStateChangeEvent(Unknown, Collecting), traceID)
+	_, err := t.applyStateChange(ctx, conn, newTraceStateChangeEvent(Unknown, Collecting), []string{traceID})
+	if err != nil {
```
kentquirk, line 779: you can just return err here, since it's either nil or not.

centralstore/redis_store.go
```diff
@@ -843,24 +849,12 @@ func (t *traceStateProcessor) toNextState(ctx context.Context, conn redis.Conn,
...
-		}
-		succeed = append(succeed, traceID)
-	}
-	otelutil.AddSpanField(span, "num_of_succeed", len(succeed))
-
-	if len(succeed) == 0 {
-		err := errors.New("failed to apply state change")
-		otelutil.AddSpanField(span, "error", err)
+	result, err := t.applyStateChange(ctx, conn, changeEvent, traceIDs)
+	if err != nil {
```
kentquirk, line 853: same here -- you don't need to return nil, err, you can just return result, err. If err is non-nil, the value of result is basically undefined so it doesn't matter what you return.

centralstore/redis_store.go
kentquirk: Maybe make a helper function args() that returns a redigo.Args?

centralstore/redis_store.go
```diff
@@ -900,77 +894,108 @@ func (t *traceStateProcessor) removeExpiredTraces(client redis.Client) {
...
-	do return -1 end
+  local timestamp = ARGV[4]
+  local result = {}
+
+  -- iterate through the traceIDs and move them to the next state
+  for i, traceID in ipairs(ARGV) do
+    -- unfortunately, Lua doesn't support "continue" statement in for loops.
+	-- even though, Lua 5.2+ supports "goto" statement, we can't use it here because
+	-- Redis only supports Lua 5.1. 
+	-- using "repeat" and "do break end" is a common pattern to simulate "continue" in Lua
```
kentquirk, line 958:
```suggestion
	-- The interior "repeat ... until true" is a way of creating a do-once loop, and "do break end" is a way to
	-- spell "break" that indicates it's not just a normal break. 
	-- This is a common pattern to simulate "continue" in Lua versions before 5.2.
```

centralstore/redis_store.go
```diff
@@ -1002,6 +1027,8 @@ func newTraceStateChangeEvent(current, next CentralTraceState) stateChangeEvent
 	}
 }
 
+// string returns a string representation of the state change event
+// this formatting logic should match the one in the stateChange lua script
```
kentquirk, line 1031: I know that Go will be faster with `return s.current + "-" + s.next`. I'm reasonably sure that in Lua, it will be faster to say `local stateChangeEvent = currentState .. "-" .. nextState`.

You should also echo this comment in the lua code that it should be the same as the go code.

kentquirk (approved):
A few nits, but nothing blocking.

PR #1043, opened 2024-03-07
feat: support Classic Ingest Keys

jharley:

We've now released Ingest Keys, but in order for them to work with Classic environments properly we need to update the key detection logic.

- upgrades husky to 0.26.0
- updates `types.IsLegacyAPIKey` to use `(husky.otlp).IsClassicApiKey`

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1206669928018696


kentquirk (approved):
Thank you! Is it OK if this isn't released for a few weeks?

brookesargent (approved):

Issue #1009, opened 2024-02-23
Docs generator should translate defaulttrue types to bool


robbkidd:
We added a `defaulttrue` type to the configuration in #998, but that's an implementation detail we ought to hide from user documentation. When we generate docs from the Meta Source of Truth, those config items ought to show type as `bool`.

kentquirk:
Closed by #1045

PR #1045, opened 2024-03-08
fix: `defaulttrue` now shows up in docs as `bool`

kentquirk:

- #1009 

- Add a condition to the generator for the config docs.


tools/convert/templates/cfg_docrepo.tmpl
mx-ruthie, line 1: I love it when the easy solution is *the* solution! Thanks for keeping me in the loop. Was great taking that ride with you.

mx-ruthie (approved):

JamieDanielson (approved):
Nice!

Issue #1044, opened 2024-03-08
Unable to pass peer address as host:ip


VinozzZ:
**Versions**

- Go: 1.21.5
- Refinery: 2.4.3

**Steps to reproduce**

1. Create a config.yaml file with
```yaml
    Peers:
      - "127.0.0.1:8901"
      - "127.0.0.1:8902"
```
2. run refinery with this config.yaml file
3. refinery errors out with 
```
failed to reload peer list: parse "0.0.0.0:9901": first path segment in URL cannot contain colon
couldn't parse peer as a URL
github.com/honeycombio/refinery/sharder.(*DeterministicSharder).loadPeerList
```

**Additional context**
If a valid url, "http://0.0.0.0:9901",  is passed in instead, refinery also errors out with
```
Validation failed for config file:
  field PeerManagement.Peers[0] (http://127.0.0.1:8901) must be a hostport: address http://127.0.0.1:8901: too many colons in address
  field PeerManagement.Peers[1] (http://127.0.0.1:8902) must be a hostport: address http://127.0.0.1:8902: too many colons in address
```

I'm able to workaround this by running refinery with `--no-validate` flag

PR #1046, opened 2024-03-08
fix: change validation type for PeerManagement.Peers to be url

VinozzZ:

we are using `url.Parse` for peer address list.  `host:port` is not a valid url.

- Change validation type for `PeerManagement.Peers` to `url` instead of `hostport`
- After this change, the convert tool will always convert `host:port` to `http://host:port`

fix: #1044

### DIFF
--- a/config/metadata/configMeta.yaml
+++ b/config/metadata/configMeta.yaml
@@ -824,7 +824,7 @@ groups:
         summary: is the list of peers to use when Type is "file", excluding self.
         description: >
           This list is ignored when Type is "redis". The format is a list of
-          strings of the form "schema://host:port".
+          strings of the form "scheme://host:port".
 
   - name: RedisPeerManagement
     title: "Redis Peer Management"
--- a/config/metadata/configMeta.yaml
+++ b/config/metadata/configMeta.yaml
@@ -816,15 +816,15 @@ groups:
         v1name: Peers
         type: stringarray
         valuetype: stringarray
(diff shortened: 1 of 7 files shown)


kentquirk:
This is the reverse of what I was thinking -- do we actually want to parse as a URL, or do we want hostport? (And either way, there might be docs in the metadata we need to fix)

VinozzZ:
@kentquirk I had the same thought. I landed on this because eventually we will need to have a schema. https://github.com/honeycombio/refinery/blob/02f6c6ae08886d37e4f9c5150e18316c4385ab53/sharder/deterministic.go#L236
We could also add the schema internally if we only support `http` and `https`

config/metadata/configMeta.yaml
```diff
@@ -816,15 +816,15 @@ groups:
...
         reload: false
         validations:
           - type: elementType
-            arg: hostport
+            arg: url
         summary: is the list of peers to use when Type is "file", excluding self.
         description: >
           This list is ignored when Type is "redis". The format is a list of
-          strings of the form "host:port".
+          strings of the form "schema://host:port".
```
kentquirk, line 827:
```suggestion
          strings of the form "scheme://host:port".
```
Please make this change and regenerate. For URLs, it's "scheme" not "schema". Yay English. :)

kentquirk (commented):
Ok, you're correct -- this never worked as documented, and the validation only proved that. Please make that one change.

kentquirk (approved):
LGTM!

PR #1047, opened 2024-03-09
feat: new metric for drops caused by rules

### LINKED ISSUE #974: Create a metric for traces/spans dropped by drop rules specifically
We get some decent metrics now on the different samplers and the applied sample rates and drops vs. keeps, but I can't find anything for spans/traces dropped by non-sampler drop rules.  I think folks would like to know how many traces their drop rules are catching and dropping overall.

It might not let them know *what* is being dropped, but at least will give them a good idea of how much traffic is matching those drop rules and being eliminated

kentquirk:

- When a user puts a Drop rule in a rules file, there's no separate metric that distinguishes that from a drop because of other sampling. See #974 for details.

- Adds a new metric, `num_dropped_by_drop_rule`, which tracks this value.

Fixes #974

### DIFF
--- a/sample/rules.go
+++ b/sample/rules.go
@@ -126,16 +126,16 @@ func (s *RulesBasedSampler) GetSampleRate(trace *types.Trace) (rate uint, keep b
 				keep = !rule.Drop && rule.SampleRate > 0 && rand.Intn(rule.SampleRate) == 0
 				reason += rule.Name
 				s.Metrics.Histogram(s.prefix+"sample_rate", float64(rate))
-				if rule.Drop {
-					// If we dropped because of an explicit drop rule, then increment that counter.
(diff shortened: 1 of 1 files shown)


sample/rules.go
```diff
@@ -125,6 +126,10 @@ func (s *RulesBasedSampler) GetSampleRate(trace *types.Trace) (rate uint, keep b
 				keep = !rule.Drop && rule.SampleRate > 0 && rand.Intn(rule.SampleRate) == 0
 				reason += rule.Name
 				s.Metrics.Histogram(s.prefix+"sample_rate", float64(rate))
+				if rule.Drop {
```
VinozzZ, line 129: Should we check that it's a drop before incrementing this metrics?
I was thinking this code would be in the else block from line 137

VinozzZ (approved):
LGTM!

PR #1057, opened 2024-03-15
feat: make redis store injectable

VinozzZ:

Redis implementation of the BasicStore can now be used with the injection library

- RedisBasicStore now implements the Starter and Stopper interface
- Added configuration to `RedisPeerManagementConfig` and `SmartWrapperOptions`


VinozzZ:
The test is failing because of the lack of metadata about the new configs. We will finalize the config options once we have the end-to-end running

centralstore/redis_store.go
```diff
@@ -742,28 +701,26 @@ func newTraceStateProcessor(cfg traceStateProcessorConfig, clock clockwork.Clock
...
 		clock:  clock,
+		done:   make(chan struct{}),
 		tracer: tracer,
 	}
 
 	return s
 }
 
-func (t *traceStateProcessor) Start(ctx context.Context, redis redis.Client) error {
+func (t *traceStateProcessor) Start(redis redis.Client) error {
```
kentquirk, line 711: Why do you have a separate Start function here? Why not make it part of the newTraceStateProcessor method? 

If it is valuable to keep it, maybe consider setting the client in the new... function and having this Start() being compatible with the startstopper model, in case we ever want to inject this.

refinerytrace/tracer.go
```diff
@@ -0,0 +1,30 @@
+package refinerytrace
```
kentquirk, line 1: I had deleted this ... i think it's no longer in use, so you should remove it from this PR too.

kentquirk (approved):
I think you need to delete refinerytrace, and I have a question about the Start method on the processor. But in general, this is great -- let's land it and keep moving!

centralstore/redis_store.go
VinozzZ: This probably is not a good name for this function and shouldn't have been exported. I changed it to `init`. 
I'm not putting it in `new` is because I don't want to `traceStateProcessor` to manage the `redis.Client`. I would like to keep the connection management in the `RedisBasicStore`

PR #1060, opened 2024-03-18
fix: deal with locking issues at startup

kentquirk:

- During startup, refinery sometimes panics. One of them was tracked down to concurrent read/write of a map in otel metrics, but the problem was a bit bigger -- some assumptions about how startup worked that weren't true, and required extra locking.

- Add locks to several metrics-related functions in different types of metrics
- Remove now-unneeded rlock inside one of the new locks.

### DIFF
--- a/metrics/legacy.go
+++ b/metrics/legacy.go
@@ -65,6 +65,9 @@ type updown struct {
 }
 
 func (h *LegacyMetrics) Start() error {
+	h.lock.Lock()
+	defer h.lock.Unlock()
+
 	h.Logger.Debug().Logf("Starting LegacyMetrics")
 	defer func() { h.Logger.Debug().Logf("Finished starting LegacyMetrics") }()
 	mc := h.Config.GetLegacyMetricsConfig()
--- a/metrics/multi_metrics.go
+++ b/metrics/multi_metrics.go
@@ -50,8 +50,9 @@ func (m *MultiMetrics) Start() error {
 	return nil
 }
 
-// This is not safe for concurrent use!
 func (m *MultiMetrics) AddChild(met Metrics) {
+	m.lock.Lock()
+	defer m.lock.Unlock()
 	m.children = append(m.children, met)
 }
 
--- a/metrics/otel_metrics.go
+++ b/metrics/otel_metrics.go
@@ -47,6 +47,9 @@ type OTelMetrics struct {
 func (o *OTelMetrics) Start() error {
 	cfg := o.Config.GetOTelMetricsConfig()
 
+	o.lock.Lock()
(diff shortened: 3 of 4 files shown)


VinozzZ:
If we made the `metrics` as a dependency to both the transmissions, at least we know there shouldn't be race condition in `Start() error` calls. 
I'm curious what assumptions about how startup worked that weren't true? Is it that `Start` is called in parallel? I have a hard time to see how `Register` would be called concurrently.

kentquirk:
I believe that what happened was this:
* Start was called on multimetrics (without locking)
* Start was also called on something that started a listener that used a transmission
* An event came in, which started a gooutine, that caused the transmission to try to write a metric, which meant reading the `counters` variable
* At the same time that the Start routine was trying to finish running and write to `counters`
* This caused the panic with read/write conflict

Adding the locks should prevent that conflict.

VinozzZ:
> I believe that what happened was this:
> 
> * Start was called on multimetrics (without locking)
> * Start was also called on something that started a listener that used a transmission
> * An event came in, which started a gooutine, that caused the transmission to try to write a metric, which meant reading the `counters` variable
> * At the same time that the Start routine was trying to finish running and write to `counters`
> * This caused the panic with read/write conflict
> 
> Adding the locks should prevent that conflict.

I see. Thanks for the clarification. The part I'm having a hard time with is the theory of concurrent `Start` calls. From what I see in the `startstop` library code, it seems like that shouldn't be the case https://github.com/facebookarchive/startstop/blob/bc158412526d4d11f91db720786fc07131b6fb4d/startstop.go#L69

kentquirk:
I wasn't saying concurrent Start -- it's an earlier Start() call starting its own goroutine that then sends a metric. I do think something's broken about the way things are initialized, but I don't exactly understand how.

I haven't tracked down *exactly* what's going on, and because stopstart deliberately randomizes the order, it's hard to reproduce. If you have an alternative theory, either of what's happening or a better way to fix it, I'd love to hear it! I don't like not being sure about the solution.

VinozzZ:
After looking into it more, the race condition could be happening from this code below. Currently, we don't have lock for `Register()` in metrics. If there are events coming in as soon as Refinery starts, the transmission code could be incrementing metrics while the code below tries to call `Register()`. https://github.com/honeycombio/refinery/blob/22d0ecae96ae7f55c43caf61fb609a22e219efd6/cmd/refinery/main.go#L280

Since we don't have a way to reproduce the error, it's hard to say this is THE cause. For this reason, we believe adding the lock to protect accesses for the map should help prevent the race condition happening again.

VinozzZ (approved):

PR #1062, opened 2024-03-20
feat: add redis connection pool stats in GetMetrics

VinozzZ:

Providing metrics for measuring contention on redis operations

- Return connection pool stats from redis store
- use prefix for different metrics group


kentquirk (approved):
This looks good as far as it goes. We are also going to need a way to Register() all of these metrics; that should probably be a RegisterMetrics function, and could be part of this PR or a different one.

kentquirk (approved):
Looks good!

PR #1066, opened 2024-03-22
fix: Add validation for ingest keys

### LINKED ISSUE #1065: Ingest keys don't work with Refinery's validation
The new ingest keys have a different format, and Refinery's key validation logic doesn't include this format. This only affects the keys that are used in the configuration file, which means only the keys that are used to send Refinery's own telemetry. There are two possible workarounds today: 
* Don't attempt to use an ingest key for refinery telemetry yet
* Use the `no-validate` flag when starting Refinery

**Steps to reproduce**

1. Put an ingest key into a config for something like `OTelMetricsConfig.APIKey`. It will fail validation

**Additional context**

kentquirk:

- Honeycomb's new ingest keys weren't being properly validated as part of the config; this means they can't be used for tracking Refinery's own telemetry. 

- Fix the regexp for key validation to include ingest keys
- Add tests for ingest key format

Fixes #1065


config/validate_test.go
```diff
@@ -276,8 +276,11 @@ func Test_validate(t *testing.T) {
 			`field Network.APIKey (****) must be a valid Honeycomb API key`},
 		{"bad format apikey long", mm("Network.APIKey", "abc123abc123whee"),
 			`field Network.APIKey (****whee) must be a valid Honeycomb API key`},
+		{"bad format ingest key long", mm("Network.APIKey", "xxxxx-abcdefgh12345678abcdefgh12345678abcdefgh12345678aabbccddee"),
+			`field Network.APIKey (****ddee) must be a valid Honeycomb API key`},
 		{"good format apikey", mm("Network.APIKey", "abc123abc123abc123abc123abc123ab"), ""},
 		{"good format apikey", mm("Network.APIKey", "NewStyleKeyWith22chars"), ""},
+		{"good format ingest key", mm("Network.APIKey", "hcaik_01hshz0tyh2fqa9wznx5a1jf4exbmsd3jj4p89k8c02eb7tx4mwgs7tf99"), ""},
```
VinozzZ, line 283: I think I don't fully understand the regex above for the ingest key.
`hc[a-z][a-z]{2}` is a string starts with `hc` and followed by two lowercase letters? 
If that's the case, why is the `hcaik_` considered a match here?

kentquirk: that's hc followed by 1 lowercase letter (the region) and then followed by 2 more (the key type). It's documented that way in a spec doc I was reading.  I was trying to be clearer than `hc[a-z]{3}` but maybe that's less clear.

VinozzZ: Ohhh, that makes sense now. Maybe we can add a comment about what this regex is matching? Something like
```
A 32-character hexadecimal string.
A 20 to 23-character alphanumeric string.
A specific pattern starting with "hc", followed by 3 lowercase letters, an underscore, and then 58 characters which can be lowercase letters or digits.
```

VinozzZ:
Thank you for adding the comment! LGTM

VinozzZ (approved):

PR #1067, opened 2024-03-22
feat: initial implementation of CentralCollector

VinozzZ:

This is the skeleton code for integrating SmartStore into refinery for dynamic scaling. There's quite bit of TODOs left for couple methods that we still need to work out but I think this is a good starting point.

- Created CentralCollector type
- Implemented AddSpan method
- Reworked the logic on how to extract key fields from each sampler


collect/central_collector.go
```diff
@@ -0,0 +1,250 @@
...
+	return c.eg.Wait()
+}
+
+// implement the Collector interface
+func (c *CentralCollector) AddSpan(span *types.Span) error {
+	return c.add(span, c.incoming)
+}
+
+func (c *CentralCollector) add(sp *types.Span, ch chan<- *types.Span) error {
+	if c.BlockOnAddSpan {
```
kentquirk, line 91: We don't need the add function any more because we only have one span channel.

I'm also tempted to suggest we can drop BlockOnAddSpan, or at least should consider removing it once we have this integrated. Maybe we need it to make sure tests don't flake, but I'm not convinced we do anymore; I don't think any of our tests even approach the number of spans we'd need before this becomes a problem.

kentquirk (approved):
A couple of things to think about but also we could fix them later if you prefer.

PR #1068, opened 2024-03-25
feat: implement span cache

kentquirk:
This contains two implementations of a span cache; `basic` is basically just a map of traceID to trace. The second, `complex`, trades off complexity and slightly slower performance for an implementation that tries to avoid allocating memory most of the time. This may prove valuable in a very high-throughput environment, but today I can't prove it -- in all the benchmarks I've written, the *only* advantage is low-allocation. So we have both implementations and we can play with them in a working refinery.

One thing that I tried was to avoid the big sort in GetOldest; I implemented an attempt to use a tdigest to find the threshold so that I could avoid the sort, but the tdigest was more expensive than sorting, even with 1 million items in the list, so I've abandoned that approach. Sorting a slice of ints is really fast even for big lists.

I'm now using the same cache impact calculation as was already in use for the previous span cache. 

The dynscaling system needs a new kind of cache for traces. This defines the interface and also has two implementations to choose from.

- Define an interface
- Basic implementation of span cache
- Complex implementation that avoids some memory allocations
- Tests and benchmarks

### DIFF
--- a/collect/cache/span_cache_basic.go
+++ b/collect/cache/span_cache_basic.go
@@ -1,6 +1,7 @@
 package cache
 
 import (
+	"sort"
 	"sync"
 
 	"github.com/facebookgo/startstop"
@@ -74,18 +75,24 @@ func (sc *spanCache_basic) Get(traceID string) *types.Trace {
 // appropriate fraction of traces
 func (sc *spanCache_basic) GetOldest(fract float64) []string {
 	n := int(float64(len(sc.cache)) * fract)
-	ret := make([]string, 0, n)
+	ids := make([]string, 0, len(sc.cache))
+
 	sc.mut.RLock()
-	i := 0
 	for traceID := range sc.cache {
-		ret = append(ret, traceID)
-		i++
-		if i >= n {
-			break
-		}
+		ids = append(ids, traceID)
 	}
+	sort.Slice(ids, func(i, j int) bool {
+		t1 := sc.cache[ids[i]]
+		t2 := sc.cache[ids[j]]
+		return t1.ArrivalTime.Before(t2.ArrivalTime)
+	})
 	sc.mut.RUnlock()
-	return ret
+
+	if len(ids) < n {
+		n = len(ids)
+	}
+	// truncate the slice to the desired length AND capacity without reallocating
+	return ids[:n:n]
 }
 
 // This gets a batch of up to n traceIDs from the cache; it's used to get a
--- a/collect/cache/span_cache_basic.go
+++ b/collect/cache/span_cache_basic.go
@@ -1,0 +1,122 @@
+package cache
+
+import (
+	"sync"
+
+	"github.com/facebookgo/startstop"
+	"github.com/honeycombio/refinery/types"
+	"github.com/jonboulle/clockwork"
+)
+
+// this is a naive implementation that uses a map
+// and doesn't try to be clever about memory usage or allocations
+type spanCache_basic struct {
+	Clock   clockwork.Clock `inject:""`
+	cache   map[string]*types.Trace
+	current []string
+	nextix  int
+	mut     sync.RWMutex
+}
+
+// ensure that spanCache implements SpanCache
+var _ SpanCache = &spanCache_basic{}
+
+// ensure that spanCache implements startstop.Starter
+var _ startstop.Starter = &spanCache_basic{}
+
+// ensure that spanCache implements startstop.Stopper
+var _ startstop.Stopper = &spanCache_basic{}
+
+func (sc *spanCache_basic) Start() error {
+	sc.cache = make(map[string]*types.Trace)
+	return nil
+}
+
+func (sc *spanCache_basic) Stop() error {
+	sc.cache = nil
+	return nil
+}
+
+func (sc *spanCache_basic) GetClock() clockwork.Clock {
+	return sc.Clock
+}
+
+func (sc *spanCache_basic) Set(sp *types.Span) error {
+	sc.mut.Lock()
+	defer sc.mut.Unlock()
+	traceID := sp.TraceID
+	trace, ok := sc.cache[traceID]
+	if !ok {
+		trace = &types.Trace{
+			APIHost:     sp.APIHost,
+			APIKey:      sp.APIKey,
+			Dataset:     sp.Dataset,
+			TraceID:     traceID,
+			ArrivalTime: sc.Clock.Now(),
+		}
+		sc.cache[traceID] = trace
+	}
+	trace.AddSpan(sp)
+	return nil
+}
+
+func (sc *spanCache_basic) Get(traceID string) *types.Trace {
+	sc.mut.RLock()
+	defer sc.mut.RUnlock()
+	trace, ok := sc.cache[traceID]
+	if !ok {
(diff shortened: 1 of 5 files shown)


collect/cache/span_cache_basic.go
```diff
@@ -0,0 +1,127 @@
...
+	defer sc.mut.RUnlock()
+	trace, ok := sc.cache[traceID]
+	if !ok {
+		return nil
+	}
+	return trace
+}
+
+func (sc *spanCache_basic) GetOldest(fract float64) []string {
+	n := int(float64(len(sc.cache)) * fract)
```
VinozzZ, line 75: I think we also need a read lock for the len(s.cache) call

collect/cache/span_cache_basic.go
```diff
@@ -0,0 +1,127 @@
...
+	"github.com/honeycombio/refinery/types"
+	"github.com/jonboulle/clockwork"
+)
+
+// this is a naive implementation that uses a map
+// and doesn't try to be clever about memory usage or allocations
+type spanCache_basic struct {
+	Clock   clockwork.Clock `inject:""`
+	cache   map[string]*types.Trace
+	current []string
```
VinozzZ, line 17: Can we add a comment for `current` and `nettix` clarifying that they are not protected by the mutex?

collect/cache/span_cache_basic.go
```diff
@@ -0,0 +1,127 @@
...
+	// truncate the slice to the desired length AND capacity without reallocating
+	return ids[:n:n]
+}
+
+// This gets a batch of up to n traceIDs from the cache; it's used to get a
+// batch of traceIDs to process in parallel. It snapshots the active map and
+// returns a slice of traceIDs that were current at the time of the call. It
+// will return successive slices of traceIDs until it has returned all of them,
+// then it will start over from a fresh snapshot.
+func (sc *spanCache_basic) GetTraceIDs(n int) []string {
```
VinozzZ, line 101: Since current and nextix are not protected by the mutex, that means `GetTraceIDs` is not concurrency-safe. Can we add a comment about that?

collect/cache/span_cache_test.go
```diff
@@ -1,106 +1,285 @@
...
+			assert.Equal(t, "trace1", trace.TraceID)
+			assert.Equal(t, "apihost", trace.APIHost)
+			assert.Equal(t, "apikey", trace.APIKey)
+			assert.Equal(t, "dataset", trace.Dataset)
+			// assert.Equal(t, c.Clock.Now(), trace.ArrivalTime)
+			assert.Equal(t, 1, c.Len())
+
+			// test that we can remove the span
+			c.Remove("trace1")
+			c.Get("trace1")
```
VinozzZ, line 68: Should we assert the result from this Get call?

collect/cache/span_cache_test.go
```diff
@@ -1,106 +1,285 @@
...
+			b.ResetTimer()
+			// we have numIDs IDs, and we'll iterate and for each
+			// ID we'll either:
+			// - add a span (80% of the time)
+			// - get a trace (10% of the time)
+			// - delete a trace (10% of the time)
+			for i := 0; i < b.N; i++ {
+				switch rand.Intn(10) {
+				case 0:
+					c.Remove(ids[i%numIDs])
```
VinozzZ, line 272: Should we put each call into a goroutine so we can benchmark the mix with contention?

collect/cache/span_cache_basic.go
```diff
@@ -0,0 +1,127 @@
...
+	current []string
+	nextix  int
+	mut     sync.RWMutex
+}
+
+// ensure that spanCache implements SpanCache
+var _ SpanCache = &spanCache_basic{}
+
+// ensure that spanCache implements startstop.Starter
+var _ startstop.Starter = &spanCache_basic{}
```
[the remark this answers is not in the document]
kentquirk: My rule of thumb is that if the object is not a leaf -- in other words, that it contains things in the injection graph -- it should be in the graph. In this case, it has clock and config.

collect/cache/span_cache_basic.go
```diff
@@ -0,0 +1,127 @@
...
+// ensure that spanCache implements startstop.Stopper
+var _ startstop.Stopper = &spanCache_basic{}
+
+func (sc *spanCache_basic) Start() error {
+	sc.cache = make(map[string]*types.Trace)
+	return nil
+}
+
+func (sc *spanCache_basic) Stop() error {
+	sc.cache = nil
```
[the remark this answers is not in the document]
kentquirk: By the time we call `Stop`, we should be completely done and already drained. So I think maybe we just don't make the cache a Stopper at all? I can remove that entirely.

collect/cache/span_cache_test.go
kentquirk: In my view, that's a different benchmark -- and I haven't come up with a good way to write that one yet.

collect/cache/span_cache_basic.go
```diff
@@ -72,32 +70,48 @@ func (sc *spanCache_basic) Get(traceID string) *types.Trace {
 }
 
 func (sc *spanCache_basic) GetOldest(fract float64) []string {
+	type tidWithImpact struct {
+		id     string
+		impact int
+	}
 	n := int(float64(len(sc.cache)) * fract)
```
VinozzZ, line 77: len(sc.cache) -> sc.Len()?

collect/cache/span_cache_complex.go
```diff
@@ -103,27 +99,38 @@ func (sc *spanCache_complex) Get(traceID string) *types.Trace {
 // because it has to sort the traces by arrival time, but I couldn't find a
 // faster way to do it.
 func (sc *spanCache_complex) GetOldest(fract float64) []string {
+	type tidWithImpact struct {
+		id     string
+		impact int
+	}
 	n := int(float64(len(sc.active)) * fract)
-	ids := make([]int, 0, len(sc.active))
+	ids := make([]tidWithImpact, 0, len(sc.active))
```
VinozzZ, line 107: len(sc.active) -> sc.Len()

collect/cache/span_cache_basic.go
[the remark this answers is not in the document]
VinozzZ: That makes sense. Then we need to make SpanCache an exported struct so that we can initialize it from outside of cache package

collect/cache/span_cache_basic.go
kentquirk: I get what you're going for, and it's technically safe here, but we have locking going on here, and using the public (locked) Len function in a routine that also locks is dangerous -- it could cause a deadlock if someone rearranges code. We could have an unlocked helper function that's used by both, but for something like this it doesn't seem worth it to me.

VinozzZ: Right. ~~My main concerns is the race condition here. Let's say the a call to `Remove(traceIDs)` happened right after the `len(sc.cache)` here. Now we may end up with a `n` that's bigger than the number of items in the new `sc.cache`.~~

I missed the check on line 98. Ok, I don't think it matters here that we get an accurate n now

VinozzZ (approved):

PR #1072, opened 2024-03-27
fix: race condition in cleanup logic in smartwrapper tests

VinozzZ:

When multiple tests are running using the same redis instance, their cleanup call may delete test data created from another running test depends on timing. 

- Run tests in each individual database index
- only cleanup test data that's created from the running test

### DIFF
--- a/centralstore/smartwrapper_test.go
+++ b/centralstore/smartwrapper_test.go
@@ -66,6 +66,8 @@ func getAndStartSmartWrapper(storetype string) (*SmartWrapper, func(), error) {
 			DroppedSize:       1000,
 			SizeCheckInterval: duration("1s"),
 		},
+		// by default redis only has 16 database index slots available
+		GetRedisDatabaseVal: rand.Intn(16),
 	}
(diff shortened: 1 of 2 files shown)


centralstore/smartwrapper_test.go
```diff
@@ -66,6 +66,8 @@ func getAndStartSmartWrapper(storetype string) (*SmartWrapper, func(), error) {
 			DroppedSize:       1000,
 			SizeCheckInterval: duration("1s"),
 		},
+		// by default redis only has 16 database index slots available
+		GetRedisDatabaseVal: rand.Intn(16),
```
kentquirk, line 70: Will this fail if two tests randomly choose the same db val?

VinozzZ: Yes. In the longer term, maybe we need to have better namespace mechanism for our test data or we use miniredis. As if right now, this is the easiest fix I could come up with

kentquirk (approved):

Issue #1037, opened 2024-03-05
Processor/Dispatcher: Refinery implementation


akvanhar:
Implement the PRocessor/dispatcher logic in Refinery
Key decision point:
Can we implement this in such a way that it works with existing Refinery logic as well as dynamic scaling work? If yes, then we do not require Dynamic Scaling work to be a breaking change.

PR #1075, opened 2024-04-02
feat: implement processor and decider in CentralCollector

VinozzZ:

Enable the CentralCollector to make sampling decisions and act on traces in its cache based on decisions retrieved from remote store.

https://github.com/honeycombio/refinery/issues/1037

- Implement processes with backoff for making decisions and process traces based on those decisions
- Fix an issue in SpanCache's GetTraceIDs
- Add `Metadata` field in CentralTraceStatus to communicate metadata attached to a given trace between refineries


collect/cache/span_cache_basic.go
```diff
@@ -127,6 +130,10 @@ func (sc *SpanCache_basic) GetTraceIDs(n int) []string {
 	if sc.nextix+n > len(sc.current) {
 		n = len(sc.current) - sc.nextix
 	}
+	defer func() {
```
kentquirk, line 133: Ohhhhhh, this is clever. Maybe too clever? It at least needs an explanatory comment, please, here and in the other place you use it.

collect/central_collector.go
```diff
@@ -71,10 +84,31 @@ func (c *CentralCollector) Start() error {
 		return nil
 	})
 
+	c.eg.Go(func() (err error) {
+		defer func() {
+			if r := recover(); r != nil {
+				err = fmt.Errorf("panic in central collector: %v\n%s", r, debug.Stack())
```
kentquirk, line 90: I like the idea of catching the panic and dumping the stack -- but I think that what this means is that if one of these processes panics, it will die and leave the rest running. 

I think instead we want to think about restarting the dead task (so that we can recover from a error that's driven by bad data) and put a counter limit on it (so we won't keep trying forever). If the counter expires, we should shut down refinery completely so that our orchestrator can restart the whole process.

(Or alternatively, never retry and just rely on restarts.)

collect/central_collector.go
```diff
@@ -119,18 +153,235 @@ func (c *CentralCollector) collect() {
...
+
+				case centralstore.DecisionDrop:
+					c.SpanCache.Remove(status.TraceID)
+				default:
+					c.Logger.Debug().Logf("trace %s is still pending", status.TraceID)
+				}
+			}
+		}
+
+		select {
```
kentquirk, line 186: I feel like everything from here down could be replaced by `time.Sleep(processTracesBackoffInterval)` since the interval is so short. But if we want this, I think we can use a Timer, not a Ticker (since we want it to fire only once) and skip this early test for c.done. It doesn't win us anything that I can tell.

The same is true in `decide()`.

collect/central_collector.go
```diff
@@ -22,7 +22,12 @@ import (
 )
 
 const (
-	cacheEjectBatchSize = 100
+	// TODO: these should be configurable
+	cacheEjectBatchSize          = 100
+	processTracesBatchSize       = 100
+	processTracesBackoffInterval = 200 * time.Microsecond
```
kentquirk, line 28: These aren't really a backoff since we don't adjust them, it's just a pause, maybe rename them to "pauseDuration" or something?

collect/central_collector.go
```diff
@@ -119,18 +153,235 @@ func (c *CentralCollector) collect() {
...
+		select {
+		case <-c.done:
+			return nil
+		default:
+		}
+
+		ticker := c.Clock.NewTicker(processTracesBackoffInterval)
+		select {
+		case <-c.done:
+			return nil
```
kentquirk, line 195: If you do need a ticker for reasons I don't understand yet, you need to Stop it here too.

collect/central_collector.go
```diff
@@ -119,18 +153,235 @@ func (c *CentralCollector) collect() {
...
-			return
+func (c *CentralCollector) processTraces() error {
+	for {
+		select {
+		case <-c.done:
+			return nil
+		default:
+			ids := c.SpanCache.GetTraceIDs(processTracesBatchSize)
+			if len(ids) == 0 {
+				continue
```
kentquirk, line 164: Let's create a metric histogram for the count returned from GetTraceIDs -- it will help in tuning the batch size.

collect/central_collector.go
```diff
@@ -119,18 +153,235 @@ func (c *CentralCollector) collect() {
...
+			statuses, err := c.Store.GetStatusForTraces(context.Background(), ids)
+			if err != nil {
+				c.Logger.Error().Logf("error getting statuses for traces: %s", err)
+				continue
+			}
+			for _, status := range statuses {
+				switch status.State {
+				case centralstore.DecisionKeep:
+					c.send(status)
+					c.SpanCache.Remove(status.TraceID)
```
kentquirk, line 176: Do these also need to get pushed into the decision caches?

collect/central_collector.go
```diff
@@ -119,18 +153,235 @@ func (c *CentralCollector) collect() {
...
+}
 
-	trace = c.SpanCache.Get(sp.TraceID)
+func (c *CentralCollector) decide() error {
+	for {
+		select {
+		case <-c.done:
+			return nil
+		default:
+			if c.BlockOnDecider {
```
kentquirk, line 209:
```suggestion
			// this is only ever true in test code
			if c.BlockOnDecider {
```

collect/central_collector.go
```diff
@@ -119,18 +153,235 @@ func (c *CentralCollector) collect() {
...
+				continue
+			}
+			ctx := context.Background()
+			tracesIDs, err := c.Store.GetTracesNeedingDecision(ctx, deciderBatchSize)
```
kentquirk, line 213: as above, let's do a histogram metric for the size actually returned.

collect/central_collector.go
```diff
@@ -119,18 +153,235 @@ func (c *CentralCollector) collect() {
...
+			for i, status := range statuses {
+				// make a decision on each trace
+				if status.State != centralstore.AwaitingDecision {
+					// someone else got to it first
+					continue
+				}
+				stateMap[status.TraceID] = status
+
+				wg.Add(1)
+				go func(status *centralstore.CentralTraceStatus, idx int) {
```
kentquirk, line 239: wondering if we want to use the errgroup functionality here too, so we can SetLimit the number of goroutines we start up here. That could also be a configurable value to keep from hammering on redis too hard. It's close to, but not identical, to `deciderBatchSize`. We probably only need one of them, but this is an alternative way to keep things under control.

collect/central_collector.go
```diff
@@ -119,18 +153,235 @@ func (c *CentralCollector) collect() {
...
+				if trace.Root != nil {
+					c.Metrics.Increment("trace_decision_has_root")
+				} else {
+					c.Metrics.Increment("trace_decision_no_root")
+				}
+
+				var sampler sample.Sampler
+				var found bool
+
+				tr := c.SpanCache.Get(trace.TraceID)
```
kentquirk, line 266: Oh, whoops, this is a mistake -- we need to be making the trace decision based on the centralspan information, NOT on the values the refinery has in cache. The rest of this function needs to be rewritten to use that central span data.

kentquirk (changes requested):
This is great stuff!

I had a bunch of comments. The biggest one is that the decider needs to use the information it gets from the cache to make the trace decisions.

collect/central_collector.go
VinozzZ: I think when the decider makes a decision about a trace, the trace decision gets put into the decision cache

collect/central_collector.go
VinozzZ: Good catch! A Timer is better for this. 
I was thinking a timer allows us to also monitor shutdown signal here without having to wait for the duration we set for the timer.

kentquirk: The timer shouldn't be named ticker anymore, and you still don't need the lines at 190-194.

collect/central_collector.go
```diff
@@ -229,13 +499,62 @@ func (c *CentralCollector) GetParentID(sp *types.Span) (string, bool) {
...
-func (c *CentralCollector) GetStressedSampleRate(traceID string) (rate uint, keep bool, reason string) {
-	return 0, false, ""
+	c.Metrics.Increment("trace_send_kept")
+	// This will observe sample rate decisions only if the trace is kept
+	c.Metrics.Histogram("trace_kept_sample_rate", float64(status.Rate))
+
+	c.Logger.Info().WithFields(logFields).Logf("Sending trace")
+	for _, sp := range trace.GetSpans() {
+		if c.Config.GetAddRuleReasonToTrace() {
+			sp.Data["meta.refinery.reason"] = status.Metadata["meta.refinery.reason"]
```
kentquirk, line 542: I think this should probably be a loop over Metadata so we can add new attributes without forgetting to update this.

collect/concurrency_limiter.go
```diff
@@ -0,0 +1,36 @@
...
+		done:  make(chan struct{}),
+	}
+}
+
+func (l *concurrencyLimiter) Go(fn func()) {
+
+	select {
+	case l.limit <- struct{}{}:
+	case <-l.done:
+		return
```
kentquirk, line 20: Do we want to block here instead of returning? Otherwise we're just going to accept work and not respond to it. Also, why use this instead of the library you had yesterday?

kentquirk (commented):
I haven't finished but I need to leave now. More later.

collect/concurrency_limiter.go
VinozzZ: The done channel is only closed when we are done with all the work. This is to prevent new work being started. Maybe we can return a boolean to indicate whether a work is started?

I don't think I have used other library for this. Do you mind to clarify which library you are referring to?

centralstore/centralstore.go
```diff
@@ -88,6 +94,46 @@ type CentralTrace struct {
 	Spans     []*CentralSpan
 }
 
+func (t *CentralTrace) GetSamplerKey() (string, bool) {
```
VinozzZ, line 97: add a samplerKey to centralTrace

sample/sample.go
```diff
@@ -12,8 +12,17 @@ import (
 	"github.com/honeycombio/refinery/types"
 )
 
+// KeyFielder is an interface for types that can return a key field
+// that's used to determine the sample rate for a trace
+type KeyInfoExtractor interface {
+	AllKeyFields() []types.KeyFielder
+	RootKeyFields() types.KeyFielder
```
VinozzZ, line 19: There has to be a better name for this

collect/concurrency_limiter.go
kentquirk: I meant errgroup.Group, which has a [SetLimit](https://pkg.go.dev/golang.org/x/sync/errgroup#Group.SetLimit) method.

collect/panic_handler.go
```diff
@@ -0,0 +1,16 @@
...
+
+import (
+	"fmt"
+	"runtime/debug"
+)
+
+func catchPanic(fn func()) (err error) {
+	defer func() {
+		if r := recover(); r != nil {
+			err = fmt.Errorf("panic in central collector: %v\n%s", r, debug.Stack())
```
kentquirk, line 11: I think we probably want to log this to the injected logger.

kentquirk (commented):
we talked about the KeyFielder stuff directly, so no other comments at this time.

collect/panic_handler.go
VinozzZ: it's logged on the caller side

collect/collect.go
```diff
@@ -297,6 +300,7 @@ func (i *InMemCollector) add(sp *types.Span, ch chan<- *types.Span) error {
 // structures.
 func (i *InMemCollector) collect() {
 	tickerDuration := i.Config.GetSendTickerValue()
+	fmt.Println("tickerDuration: ", tickerDuration)
```
kentquirk, line 303: Do we still need/want the printf?

collect/collect.go
```diff
@@ -603,15 +607,12 @@ func (i *InMemCollector) send(trace *types.Trace, sendReason string) {
 	var found bool
 
 	// get sampler key (dataset for legacy keys, environment for new keys)
-	samplerKey, isLegacyKey := trace.GetSamplerKey()
+	fmt.Println("i.datasetPrefix: ", i.datasetPrefix)
```
kentquirk, line 610: printf?

collect/retryLimiter.go
```diff
@@ -0,0 +1,39 @@
+package collect
+
+import "sync"
+
+type retryLimiter struct {
+	limit int
```
kentquirk, line 6: If we're using this, I am thinking the limit needs a lock -- or better yet, we could use the size of a channel as the lock. Fill up a channel, to acquire something you remove an item from the channel, and you put it back into the channel when you're done with it. You'll never block on adding something, but you might block on getting one.

internal/redis/test_harness.go
```diff
@@ -4,14 +4,16 @@ import (
...
+	"github.com/honeycombio/refinery/metrics"
 	"github.com/jonboulle/clockwork"
 )
 
 var _ Client = &TestService{}
 
 type TestService struct {
-	Config  config.Config // currently ignored by the test service
-	Clock   clockwork.Clock
+	Config  config.Config   `inject:""` // currently ignored by the test service
```
kentquirk, line 14: "currently ignored" isn't actually true any more

kentquirk (approved):
Removing my "request changes" - I do think it needs a bit more but it's quite close now.

PR #1077, opened 2024-04-07
feat: RedisMetricsRecorder

kentquirk:
This is the Refinery side of recording metrics to a central place in Redis so that we can make it available for autoscaling.

This will be followed up with the central metrics reporting app.

- We want to record metrics centrally so we can scale with them.

- Add GetAll to the MultiMetrics object so that we can fetch and send them all at once
- Send them to Redis with an expiration time so that they'll clean themselves up
- Use the hostname if it exists, otherwise the time, as a differentiator

### DIFF
--- a/centralstore/redis_metrics_recorder.go
+++ b/centralstore/redis_metrics_recorder.go
@@ -66,6 +66,7 @@ func (r *RedisMetricsRecorder) monitor() {
 			conn.SetHashTTL("Refinery_Metrics_"+r.prefix, allmetrics, r.reportingFreq*2)
 			conn.Close()
 		case <-r.done:
+			ticker.Stop()
 			return
 		}
 	}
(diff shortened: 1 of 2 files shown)


centralstore/redis_metrics_recorder.go
```diff
@@ -0,0 +1,72 @@
...
+	return nil
+}
+
+// We store the metrics in Redis as a hash with the prefix "Refinery_Metrics_" +
+// r.prefix, which should be made unique to this instance of refinery. The hash
+// is stored with an expiration time of 2 times the reporting frequency. The
+// hash will be removed automatically once the refinery instance stops. It is
+// expected that the metrics reporter will iterate the prefixes and aggregate them.
+func (r *RedisMetricsRecorder) monitor() {
+	ticker := r.Clock.NewTicker(r.reportingFreq)
```
VinozzZ, line 60: missing `ticker.Stop()` here

PR #1078, opened 2024-04-09
feat: make central collector injectable

VinozzZ:

The new CentralCollector should replace the old InMemoryCollector. 

- Implements the rest of the collector functionalities except stress relief
- Makes sure CentralCollector has access to all necessary information when making a decision for a trace as well as 
sending a kept trace.
- Since the `ParentID` is not currently used for anything but determine whether a span is root span, I replaced the `ParentID` field with `IsRoot` field on span objects
- Adapted applicable existing Refinery tests to use the new CentralCollector.


collect/central_collector.go
```diff
@@ -19,13 +22,38 @@ import (
...
+	TraceSendGotRoot        = "trace_send_got_root"
+	TraceSendExpired        = "trace_send_expired"
+	TraceSendEjectedFull    = "trace_send_ejected_full"
+	TraceSendEjectedMemsize = "trace_send_ejected_memsize"
+	TraceSendLateSpan       = "trace_send_late_span"
+)
+
 const (
 	// TODO: these should be configurable
 	cacheEjectBatchSize         = 100
```
VinozzZ, line 52: this is not needed anymore

centralstore/smartwrapper_test.go
```diff
@@ -83,12 +69,25 @@ func getAndStartSmartWrapper(storetype string) (*SmartWrapper, func(), error) {
...
-		{Value: store},
 		{Value: sw},
 	}
 	g := inject.Graph{Logger: dummyLogger{}}
+
+	var store BasicStorer
+	switch storetype {
+	case "local":
+		store = &LocalRemoteStore{}
+	case "redis":
```
VinozzZ, line 81: add another case "real-redis"

centralstore/smartwrapper_test.go
```diff
@@ -311,9 +311,8 @@ func TestSetTraceStatuses(t *testing.T) {
 
 	err = store.RecordMetrics(ctx)
 	require.NoError(t, err)
-	value, ok := store.Metrics.Get("redisstore_count_traces")
+	_, ok := store.Metrics.Get("redisstore_count_traces")
 	require.True(t, ok)
-	require.EqualValues(t, 0, value)
```
VinozzZ, line 322: capture this initial value and compare the result with the initial value

kentquirk (approved):
We've talked about this at length so I'm approving; we'll still have cleanup to do but it's just merging to a decision branch, so let's go.

PR #1079, opened 2024-04-11
maint: Move redis out of internal to a public folder

kentquirk:

- I want to reuse the redis code, and doing so requires making it public. 
- I also made it use a subset of the full config object so I can implement it differently in the other app

- Move redis out of internal
- Create RedisConfig interface
- Use it in the redis code


centralstore/redis_metrics_recorder.go
```diff
@@ -61,10 +61,13 @@ func (r *RedisMetricsRecorder) monitor() {
 	for {
 		select {
 		case <-ticker.Chan():
-			allmetrics := r.Metrics.GetAll()
-			conn := r.RedisClient.Get()
-			conn.SetHashTTL("Refinery_Metrics_"+r.prefix, allmetrics, r.reportingFreq*2)
-			conn.Close()
+			// if we don't have a redis client, we can't report metrics to it
```
kentquirk, line 64: Whoops, well, I had noticed this potential problem and it snuck into this PR. It technically doesn't belong here, but doesn't feel worth it to me to remove it.

VinozzZ (approved):

Issue #1082, opened 2024-04-11
Make integration tests more reliable


VinozzZ:
Add a channel to the collector, decider, and processor goroutine so that we can have more control when each goroutine runs during tests.

PR #1083, opened 2024-04-12
fix: make central collector tests more reliable

VinozzZ:

Since processor and decider run concurrently, it's hard to manage the timing in tests. 

#1082 

- implement a Cycle object to manage the goroutine so we can pause or restart it
- modify tests to use the new Cycle management interface

### DIFF
--- a/app/app_test.go
+++ b/app/app_test.go
@@ -96,7 +96,6 @@ func newStartedApp(
 	enableHostMetadata bool,
 ) (*App, inject.Graph, func()) {
 	c := &config.MockConfig{
-		GetSendDelayVal:          0,
 		GetTraceTimeoutVal:       10 * time.Millisecond,
 		GetMaxBatchSizeVal:       500,
 		GetSamplerTypeVal:        &config.DeterministicSamplerConfig{SampleRate: 1},
@@ -120,6 +119,7 @@ func newStartedApp(
 			BasicStoreType:  "redis",
 			SpanChannelSize: 10000,
 			SendDelay:       config.Duration(2 * time.Millisecond),
+			DecisionTimeout: config.Duration(100 * time.Millisecond),
 		},
 	}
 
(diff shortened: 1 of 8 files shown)


kentquirk (commented):
This feels like a lot of complexity. Do we need this much?

What I had in mind for stepping a test was:
* The ticker is already configurable, so the text fixture would set the ticker to a Long Time so that it wouldn't fire during the test
* Add a runOnce channel to the objects and add a case to the select statements that are already there. 
* Tests can inject a value into the runOnce channel as often as they need to.

If you think this is worth the added code, I'd like to see a bunch more comments please, particularly around the channel containing a struct that contains a channel. That's tricky stuff! BTW, you can have a channel that just contains a channel if you want; you don't need the struct in the middle.

VinozzZ:
> This feels like a lot of complexity. Do we need this much?
> 
> 
> 
> What I had in mind for stepping a test was:
> 
> * The ticker is already configurable, so the text fixture would set the ticker to a Long Time so that it wouldn't fire during the test
> 
> * Add a runOnce channel to the objects and add a case to the select statements that are already there. 
> 
> * Tests can inject a value into the runOnce channel as often as they need to.
> 
> 
> 
> If you think this is worth the added code, I'd like to see a bunch more comments please, particularly around the channel containing a struct that contains a channel. That's tricky stuff! BTW, you can have a channel that just contains a channel if you want; you don't need the struct in the middle.
> 
> 

Yes, I originally started with the design you described above. However, I quickly learned that the runOnce channel still needs a way to tell the caller that the run has finished. Therefore, I need to be able to send a signal once a run is completed. Otherwise, tests won't be reliable because we can't make decider and processor synchronous calls. 

I can add more comments into the abstraction for sure

kentquirk (approved):
Nice. You even added the thing I forgot to ask you to put in. :)

Issue #1041, opened 2024-03-05
Refinery Shutdown logic


akvanhar:
**Is your feature request related to a problem? Please describe.**
Refinery needs to know how to drain itself on shutdown.
The design has been done, but has not yet been implemented.

- [ ] We also need to make sure we mark the refinery instance as unavailable for LB when Refinery received shutdown signal

PR #1085, opened 2024-04-16
feat: implement graceful shutdown for central store

VinozzZ:

Drain in-flight spans during refinery shutdown
#1041

- Run `process()` to monitor trace decisions for in-flight traces
- Stop accepting new spans immediately after receive shutdown signal
- Send remaining spans without a decision to central store
- consolidate `SpanType` constant

### DIFF
--- a/collect/central_collector.go
+++ b/collect/central_collector.go
@@ -125,7 +125,7 @@ func (c *CentralCollector) Start() error {
 	c.eg = &errgroup.Group{}
 	c.eg.SetLimit(retryLimit)
 	c.eg.Go(func() error {
-		err := catchPanic(c.collect)
+		err := catchPanic(c.receive)
 		if err != nil {
 			c.Logger.Error().Logf("error collecting spans: %s", err)
 		}
@@ -133,7 +133,7 @@ func (c *CentralCollector) Start() error {
 	})
 
 	c.eg.Go(func() error {
-		err := catchPanic(c.processor)
+		err := catchPanic(c.process)
 		if err != nil {
 			c.Logger.Error().Logf("error processing traces: %s", err)
 		}
@@ -141,7 +141,7 @@ func (c *CentralCollector) Start() error {
 	})
 
 	c.eg.Go(func() error {
-		err := catchPanic(c.decider)
+		err := catchPanic(c.decide)
 		if err != nil {
 			c.Logger.Error().Logf("error making decision for traces: %s", err)
 		}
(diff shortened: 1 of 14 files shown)


collect/central_collector.go
```diff
@@ -151,11 +151,74 @@ func (c *CentralCollector) Start() error {
...
+}
+
+// shutdown implements the shutdown logic for the collector.
+// It will immediately stop the receiver and the decider.
+// The processor will finish its current cycle before stopping.
+// It will also upload all the remaining traces in the cache to
+// the central store.
+func (c *CentralCollector) shutdown(ctx context.Context) error {
+	// try to send the remaining traces to the cache
+	interval := 20 * time.Microsecond
```
kentquirk, line 182: This feels like a *really* short interval for an actual network. 

Also, I'm slightly confused here; the comment implies that the processor shuts down after its current run, but it looks like you're allowing the processor to try to send everything. Am I misreading this? Maybe some more comments?

VinozzZ: Oops, that's my bad. I hardcoded it to make it faster for testing purpose. My original thought was to give a separate `done` signal for the processor so we don't have to spin up a new one. I changed my mind since that would require us adding code to synchronize config change for the interval. I will update the comment to reflect the current design

collect/central_collector.go
```diff
@@ -151,11 +151,74 @@ func (c *CentralCollector) Start() error {
...
+		c.Logger.Error().Logf("error processing remaining traces: %s", err)
+	}
+	defer close(done)
+
+	fmt.Println("Sending remaining traces to central store")
+
+	// send the remaining traces to the central store
+	ids := c.SpanCache.GetTraceIDs(c.SpanCache.Len())
+	for _, id := range ids {
+		trace := c.SpanCache.Get(id)
```
VinozzZ, line 198: TODO: logs for how many we send

collect/central_collector.go
```diff
@@ -151,11 +151,74 @@ func (c *CentralCollector) Start() error {
...
+				IsRoot:    sp.IsRoot,
+			}
+
+			cs.SetSamplerKey(trace.GetSamplerKey(c.Config.GetDatasetPrefix()))
+			err := c.Store.WriteSpan(ctx, cs)
+			if err != nil {
+				c.Logger.Error().Logf("error sending span %s for trace %s during shutdown: %s", sp.ID, id, err)
+			}
+			if errors.Is(err, context.DeadlineExceeded) {
+				return err
```
VinozzZ, line 216: TODO: log this error

config/file_config.go
```diff
@@ -239,6 +240,13 @@ type SmartWrapperOptions struct {
 	ReaperRunInterval Duration `yaml:"ReaperRunInterval" default:"1h"`
 }
 
+func (c CollectionConfig) GetShutdownDelay() time.Duration {
```
kentquirk, line 243: Given that we have `GetCollectionConfig` and that you're using it, I'm leaning toward getting rid of the specific field accessors in favor of just accessing it. It has a default and we can put a "nonzero" validator on it, so we can be sure it's never 0. That does mean that tests have to be sure to set it, but I think that's acceptable in favor of getting rid of a few dozen accessors. 

(This is a request to get rid of this, not a demand.)

kentquirk (approved):
Approved -- one comment, but it's optional.

config/file_config.go
VinozzZ: I like the idea. We can make this change in another PR

PR #1086, opened 2024-04-16
fix: implement retry for central collector

VinozzZ:

Now that the trace processing logic are separated into three goroutines, we want to make sure only running refinery if all three goroutines are running.

- implement a goroutine runner for the three main goroutines in central collector


VinozzZ:
We have decided that the benefit of this design is not worth it for the added code complexity

PR #1091, opened 2024-04-18
fix: local store fixes

kentquirk:

- Local store wasn't being tested and it had some problems

- Rename localremotestore to localstore
- Fix tests in smartwrapper to test both kinds of store
- Fix tests in central_collector to test both kinds of store
- Rename "sampler key" to "sampler selector" because it's used to choose a sampler, not the key that's given to a sampler.
- Rewrite Clone so it copies the struct and then the interior of the map, rather than individual fields
- Skip a couple of tests that don't make sense without peers (I didn't delete them because we might want to add something that tests related behavior). 
- Rename "LocalRemoteStore" (which was never a good name but has become completely wrong) to "LocalStore"


centralstore/smartwrapper_test.go
```diff
@@ -28,7 +28,7 @@ func duration(s string) config.Duration {
 	return config.Duration(d)
 }
 
-var storeType = "redis"
+var storeType = "local"
```
VinozzZ, line 31: should we remove this global variable now?

VinozzZ (approved):
Just one small nit. LGTM

PR #1093, opened 2024-04-18
feat: record metrics on an interval in central store

VinozzZ:

Record central store metrics on an interval so we have metrics ready for stress level calculation

- added metrics for smart wrapper, central collectors
- added metrics cycle to run `RecordMetrics` in an interval
- add span counts to all spans' metadata instead of only on root span
- removed unnecessary code
- inject StressRelief to central collector

### DIFF
--- a/collect/central_collector.go
+++ b/collect/central_collector.go
@@ -28,8 +28,6 @@ type Collector interface {
 	// scheduled for transmission.
 	AddSpan(*types.Span) error
 	Stressed() bool
-	GetStressedSampleRate(traceID string) (rate uint, keep bool, reason string)
-	ProcessSpanImmediately(sp *types.Span, keep bool, sampleRate uint, reason string)
 }
 
 func GetCollectorImplementation(c config.Config) Collector {
@@ -45,6 +43,8 @@ const (
 	TraceSendEjectedFull    = "trace_send_ejected_full"
 	TraceSendEjectedMemsize = "trace_send_ejected_memsize"
 	TraceSendLateSpan       = "trace_send_late_span"
+
+	metricsCycleInterval = 1 * time.Second
 )
 
 type traceForDecision struct {
@@ -65,6 +65,7 @@ type CentralCollector struct {
 	Transmission   transmit.Transmission    `inject:"upstreamTransmission"`
 	Logger         logger.Logger            `inject:""`
 	Metrics        metrics.Metrics          `inject:"genericMetrics"`
+	StressRelief   StressReliever           `inject:"stressRelief"`
 	SamplerFactory *sample.SamplerFactory   `inject:""`
 	SpanCache      cache.SpanCache          `inject:""`
 
@@ -80,6 +81,7 @@ type CentralCollector struct {
 	eg             *errgroup.Group
 	processorCycle *Cycle
 	deciderCycle   *Cycle
+	metricsCycle   *Cycle
 
 	hostname string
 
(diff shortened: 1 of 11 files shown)


collect/central_collector.go
```diff
@@ -101,24 +103,48 @@ func (c *CentralCollector) Start() error {
...
+	c.metricsCycle = NewCycle(c.Clock, c.Config.GetSendTickerValue(), c.done)
 	c.processorCycle = NewCycle(c.Clock, collectorCfg.GetProcessTracesPauseDuration(), c.done)
 	c.deciderCycle = NewCycle(c.Clock, collectorCfg.GetDeciderPauseDuration(), c.done)
 
 	c.Metrics.Register("collector_processor_batch_count", "histogram")
 	c.Metrics.Register("collector_decider_batch_count", "histogram")
 	c.Metrics.Register("trace_send_kept", "counter")
+	c.Metrics.Register("trace_duration_ms", "histogram")
+	c.Metrics.Register("trace_span_count", "histogram")
+	c.Metrics.Register("trace_decision_ket", "counter")
```
kentquirk, line 115: what does "ket" mean?

collect/central_collector.go
```diff
@@ -630,6 +656,10 @@ func (c *CentralCollector) send(status *centralstore.CentralTraceStatus) {
 			sp.Data["meta.refinery.send_reason"] = TraceSendLateSpan
 
 		}
+		sp.Data["meta.span_event_count"] = int(status.SpanEventCount())
```
kentquirk, line 659: Did we decide to add counts to every span? And make it not optional? We'll have to change this in config too.

collect/central_collector_test.go
```diff
@@ -681,11 +679,11 @@ func TestCentralCollector_AddCountsToRoot(t *testing.T) {
 
 	transmission.Mux.RLock()
 	require.Equal(t, 5, len(transmission.Events), "adding a root span should send all spans in the trace")
-	assert.Equal(t, nil, transmission.Events[0].Data["meta.span_count"], "child span metadata should NOT be populated with span count")
-	assert.Equal(t, nil, transmission.Events[1].Data["meta.span_count"], "child span metadata should NOT be populated with span count")
-	assert.Equal(t, nil, transmission.Events[1].Data["meta.span_event_count"], "child span metadata should NOT be populated with span event count")
-	assert.Equal(t, nil, transmission.Events[1].Data["meta.span_link_count"], "child span metadata should NOT be populated with span link count")
-	assert.Equal(t, nil, transmission.Events[1].Data["meta.event_count"], "child span metadata should NOT be populated with event count")
+	assert.Equal(t, 2, transmission.Events[0].Data["meta.span_count"], "child span metadata should NOT be populated with span count")
```
kentquirk, line 682: The error messages here and below are now wrong

collect/central_collector_test.go
```diff
@@ -1322,16 +1320,15 @@ func duration(s string) config.Duration {
 
 func waitUntilReadyToDecide(t *testing.T, coll *CentralCollector, traceIDs []string) {
 	ctx := context.Background()
+	idMap := make(map[string]struct{})
```
kentquirk, line 1323: If you use generics.Set here you can call idset.Add(ids...).

kentquirk (approved):
A couple of nits, but feel free to merge after fixing the error messages.

collect/central_collector.go
VinozzZ: I believe so unless my memory is messing with me which is not a uncommon thing

collect/central_collector.go
VinozzZ: it means I had a typo 😆

PR #1094, opened 2024-04-18
fix: do not exit from the processor and decider loop on first error

VinozzZ:

We shouldn't exit on the first error returned from processor or decider since that can happen when redis is upgrading or in the process of leader election if redis is in cluster mode.

- added a `isTest` flag to make sure we can still catch the error in tests


kentquirk (approved):
Seems like what we talked about...

PR #1096, opened 2024-04-19
feat: implement stress relief

VinozzZ:

In order to share stress information across all instances in a cluster, this PR implements the new stress relief calculation for stress level. 

- Use Gossip to communicate stress level amongst peers
- Fix concurrency issue with the GossipRedis
- Add tests for the new stress relief calculcation


collect/stressRelief/stress_relief.go
```diff
@@ -0,0 +1,154 @@
...
+	return m.SampleRate, m.ShouldKeep, "mock"
+}
+func (m *MockStressReliever) ShouldSampleDeterministically(traceID string) bool {
+	return m.SampleDeterministically
+}
+
+// hashSeed is a random value to seed the hash generator for the sampler.
+// We want it to be a constant that's the same across all nodes so that they
+// all make the same sampling decisions during stress relief.
+const hashSeed = 34527861234
```
VinozzZ, line 43: FYI: All the code below is copied from the original file

route/route.go
```diff
@@ -518,7 +518,13 @@ func (r *Router) processEvent(ev *types.Event, reqID interface{}) error {
 	// if we are, then we want to make an immediate, deterministic trace decision
```
kentquirk, line 518: Please fix this comment so it says what we do now.

internal/gossip/gossip.go
```diff
@@ -60,6 +60,9 @@ func (g *InMemoryGossip) Subscribe(channel string, callback func(data []byte)) e
 
 func (g *InMemoryGossip) Start() error {
 	g.channel = make(chan []byte, 10)
+	g.eg = &errgroup.Group{}
```
kentquirk, line 63: if we're not using context or limit, why bother with errgroup? Is there some other benefit?

collect/stressRelief/stress_relief_test.go
```diff
@@ -0,0 +1,129 @@
...
+	}, 2*time.Second, 100*time.Millisecond, "stress relief should be false")
+}
+
+// TestStressRelief_Sample tests that traces are sampled deterministically
+// by traceID.
+// The test generates 10000 traceIDs and checks that the sampling rate is
+// within 10% of the expected value.
+func TestStressRelief_ShouldSampleDeterministically(t *testing.T) {
+	traceIDs := make([]string, 0, 100000)
+	for i := 0; i < 100; i++ {
```
kentquirk, line 181: I don't recall if this is old or new code, but we say 10K in the comments, and only generate 100 of them.

Also, if we did generate 10K, the random generator has only 1M possibilities, so there are going to be a lot of duplicates. 

OTel TraceIDs are supposed to be 32-char hex values; maybe something like:
`fmt.Sprintf("%016x%016x", rand.Int63(), rand.Int63())`

collect/stressRelief/stress_relief_redis.go
```diff
@@ -0,0 +1,365 @@
...
+	}
+
+	// All of the numerator metrics are gauges. The denominator metrics are constants.
+	s.calcs = []StressReliefCalculation{
+		{Numerator: "collector_incoming_queue_length", Denominator: "INCOMING_CAP", Algorithm: "sqrt", Reason: "CacheCapacity (incoming)"},
+		{Numerator: "libhoney_upstream_queue_length", Denominator: "UPSTREAM_BUFFER_SIZE", Algorithm: "sqrt", Reason: "UpstreamBufferSize"},
+		{Numerator: "memory_heap_allocation", Denominator: "MEMORY_MAX_ALLOC", Algorithm: "sigmoid", Reason: "MaxAlloc"},
+		// TODO: what's the denominator for this one?
+		// users need to tell us what's their redis memory limit
+		//{Numerator: "redisstore_memory_used_total", }
```
kentquirk, line 81: Let's add a configuration value for this; maybe just add a TODO since we'll need to come back to it anyway.

kentquirk (approved):
A couple of minor issues, but not enough to keep things from moving along.

collect/stressRelief/stress_relief_redis.go
VinozzZ: i added a TODO here

internal/gossip/gossip.go
VinozzZ: I changed to wait group. The reason why I'm using a control group for goroutines is because I want to make sure we are not leaking anything

kentquirk: OK; for what it's worth, I had missed that benefit of errgroup and would have been happy for you to leave it. But the waitgroup is good too.

VinozzZ: I was able to add it back cuz I had to do a rebase against the feature branch anyway 🥳

PR #1098, opened 2024-04-22
feat: implement Gossiper interface

VinozzZ:

When using Refinery with Redis, we need to let each Refinery in a cluster know about each other's stress level and config changes. This PR implements the communication protocol for such use case

- Define a `Gossiper` interface
- Implement the `Gossiper` interface using Redis pub/sub

### DIFF
--- a/internal/gossip/gossip.go
+++ b/internal/gossip/gossip.go
@@ -1,0 +1,24 @@
+package gossip
+
+import "github.com/facebookgo/startstop"
+
+type Gossiper interface {
+	// Gossip sends a message to all peers that are listening on the channel
+	Publish(channel string, message Message) error
+
+	// Register a handler for messages on a channel
+	// The handler will be called for each message received on the channel
+	Register(channel string, handler func(Message)) error
+
+	startstop.Starter
+	startstop.Stopper
+}
+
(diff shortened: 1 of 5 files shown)


kentquirk (commented):
What I had been thinking was that Message should be a map of k-v pairs, but I hadn't thought about getting that to go through the redis publish interface. If we're only going to have a single value, then I think that:
* we don't need a name for it (the key)
* we should just say that message is an arbitrary string (or []byte) and make the caller responsible for any encoding / decoding they need. You've invented a protocol here, but it seems like there are entire libraries designed to support this sort of thing.

Do you mind deleting some code?

internal/gossip/gossip.go
```diff
@@ -0,0 +1,101 @@
...
+	msg := message{
+		key:  channel,
+		data: value,
+	}
+
+	select {
+	case <-g.done:
+		return errors.New("gossip has been stopped")
+	case g.channel <- msg.ToBytes():
+	default:
```
kentquirk, line 45: I don't love the default case here; it could lead to us dropping messages a bit too easily. Can we please give the channel a length (like 10)?

kentquirk (approved):
Looks great.

PR #1099, opened 2024-04-23
feat: add health package

kentquirk:

- Adds an internal health monitor package that will be used for health checks

- Add health object
- Add tests for it

### DIFF
--- a/internal/health/health.go
+++ b/internal/health/health.go
@@ -1,7 +1,6 @@
 package health
 
 import (
-	"fmt"
 	"sync"
 	"time"
 
@@ -45,7 +44,7 @@ var TickerTime = 1 * time.Second
 // When services are registered, they will be expected to report in at least once every timeout interval.
 // If they don't, they will be marked as not alive.
 type Health struct {
-	Clock    clockwork.Clock
+	Clock    clockwork.Clock `inject:""`
 	timeouts map[string]time.Duration
 	timeLeft map[string]time.Duration
 	readies  map[string]bool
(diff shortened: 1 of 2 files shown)


internal/health/health.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+// TickerTime is the interval at which we will check the health of the system.
+var TickerTime = 1 * time.Second
+
+// The Health object is the main object that services will interact with.
+// When services are registered, they will be expected to report in at least once every timeout interval.
+// If they don't, they will be marked as not alive.
+type Health struct {
+	Clock    clockwork.Clock
```
VinozzZ, line 48: Do we want to inject this service?

internal/health/health.go
```diff
@@ -0,0 +1,143 @@
...
+	return nil
+}
+
+func (h *Health) Stop() error {
+	close(h.done)
+	return nil
+}
+
+func (h *Health) ticker() {
+	fmt.Println("starting ticker")
```
VinozzZ, line 75: do we still need the print lines here?

internal/health/health.go
```diff
@@ -0,0 +1,182 @@
...
+	h.mut.RLock()
+	defer h.mut.RUnlock()
+	// if any counter is not positive, we're not ready
+	for _, a := range h.timeLeft {
+		if a <= 0 {
+			return false
+		}
+	}
+
+	// if no one has registered yet, we're not ready
```
VinozzZ, line 171: If no one has registered, then `h.timeLeft` will be all `0` and that should be covered by line 166?

VinozzZ (approved):

internal/health/health.go
kentquirk: If no one has registered, these loops don't get run, so line 166 never gets executed.

internal/health/health.go
kentquirk: oops

internal/health/health.go
kentquirk: Yes!

PR #1100, opened 2024-04-24
feat: health checks now have alive/ready probes

kentquirk:

Refinery previously only had a simple `alive` health check, that just returned true all the time. In a world where we want to shut down cleanly and drain our traffic, we need both `alive` and `ready` checks so that we can turn off `ready` when we're shutting down. This way the load balancer will stop sending traffic while the server drains itself.

- Inject the new health system
- Call it from centralCollector in its different goroutines
- Call it with ready = false in the shutdown routine
- Rework alive endpoint to set the JSON content type
- Add the `/ready` endpoint (clone of alive) to deliver ready status
- Rewrite the grpc endpoints -- use the standard helper rather than implementing the interface manually (as was done in #498)
- Delete the existing grpc tests because they relied on calling the interface we no longer implement
- I haven't written new tests because I don't know how to do that in grpc-land


collect/central_collector.go
```diff
@@ -199,6 +212,10 @@ func (c *CentralCollector) shutdown(ctx context.Context) error {
 				return err
 			}
 		}
+		// we have to make sure the health check says we're alive but not accepting data during shutdown
+		c.Health.Ready(receiverHealth, false)
+		c.Health.Ready(processorHealth, true)
+		c.Health.Ready(deciderHealth, true)
```
VinozzZ, line 218: Do we want to keep the decider running during shutdown? I think we only want the processor running since that's the one sending traces in the SpanCache

VinozzZ: I see. That makes sense. And yes, currently as soon as we receive the shutdown signal, the decider stops asking for more work

kentquirk: Thanks, I've updated.

PR #1101, opened 2024-04-24
fix: persist keep reason to central store

VinozzZ:

If a trace is kept, we want to persist the reason in `KeepReason` field

- set keep reason during making decision
- retrieve reason from `KeepReason` before sending a trace


collect/central_collector.go
```diff
@@ -467,8 +468,6 @@ func (c *CentralCollector) makeDecision(ctx context.Context) error {
 			c.Logger.Info().WithFields(logFields).Logf("Dropping trace because of sampling")
 		}
 		c.Metrics.Increment("trace_decision_kept")
-		// This will observe sample rate decisions only if the trace is kept
-		c.Metrics.Histogram("trace_kept_sample_rate", float64(rate))
```
kentquirk, line 471: Is this being maintained somewhere else, or do we just lose this metric?

kentquirk (approved):

PR #1106, opened 2024-04-26
fix: remove peer from injection graph

VinozzZ:

We don't use peer in new refinery anymore

-

### DIFF
--- a/cmd/refinery/main.go
+++ b/cmd/refinery/main.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"context"
 	"fmt"
 	"net"
 	"net/http"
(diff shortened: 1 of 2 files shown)


TylerHelmuth (approved):

cmd/refinery/main.go
```diff
@@ -123,10 +121,7 @@ func main() {
 		os.Exit(1)
 	}
 
-	ctx, cancel := context.WithTimeout(context.Background(), cfg.GetPeerTimeout())
-	defer cancel()
 	done := make(chan struct{})
-	peers, err := peer.NewPeers(ctx, cfg, done)
 
 	if err != nil {
```
TylerHelmuth, line 126: Can remove this error check too

TylerHelmuth (approved):

PR #1107, opened 2024-04-28
feat: add tracing to store

kentquirk:

- We want to be able to watch performance of the store in detail; profiling is helpful but not enough. We had started to set up tracing already, this makes it work. It generates a lot of spans rather quickly, so it should definitely NOT be used in production.

- Add tracers to a few key types
- Use them to generate spans and decorate them with useful data

### DIFF
--- a/centralstore/local_store.go
+++ b/centralstore/local_store.go
@@ -8,9 +8,11 @@ import (
 
 	"github.com/honeycombio/refinery/collect/cache"
 	"github.com/honeycombio/refinery/config"
+	"github.com/honeycombio/refinery/internal/otelutil"
 	"github.com/honeycombio/refinery/metrics"
 	"github.com/honeycombio/refinery/types"
 	"github.com/jonboulle/clockwork"
+	"go.opentelemetry.io/otel/trace"
 )
 
 // LocalStore is a basic store that is local to the Refinery process. This is
@@ -19,6 +21,7 @@ type LocalStore struct {
 	Config        config.Config        `inject:""`
 	DecisionCache cache.TraceSentCache `inject:""`
 	Metrics       metrics.Metrics      `inject:"genericMetrics"`
+	Tracer        trace.Tracer         `inject:"tracer"`
 	Clock         clockwork.Clock      `inject:""`
 	// states holds the current state of each trace in a map of the different states
 	// indexed by trace ID.
@@ -121,6 +124,9 @@ func (lrs *LocalStore) changeTraceState(traceID string, fromState, toState Centr
 // during shutdown. WriteSpan is expecting to be called from an asynchronous
 // process and will only return an error if the span could not be written.
 func (lrs *LocalStore) WriteSpan(ctx context.Context, span *CentralSpan) error {
+	_, spanWrite := otelutil.StartSpan(ctx, lrs.Tracer, "LocalStore.WriteSpan")
+	defer spanWrite.End()
+
 	lrs.mutex.Lock()
 	defer lrs.mutex.Unlock()
 
@@ -191,6 +197,8 @@ func (lrs *LocalStore) WriteSpan(ctx context.Context, span *CentralSpan) error {
 // populated. Normally this call will be made after Refinery has been asked
 // to make a trace decision.
 func (lrs *LocalStore) GetTrace(ctx context.Context, traceID string) (*CentralTrace, error) {
+	_, span := otelutil.StartSpan(ctx, lrs.Tracer, "LocalStore.GetTrace")
+	defer span.End()
 	lrs.mutex.RLock()
 	defer lrs.mutex.RUnlock()
 	if trace, ok := lrs.traces[traceID]; ok {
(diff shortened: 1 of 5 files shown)


internal/otelutil/otel_tracing.go
```diff
@@ -27,30 +27,51 @@ func AddException(span trace.Span, err error) {
...
+	return attrs
+}
+
+// Starts a span with no extra fields.
+func StartSpan(ctx context.Context, tracer trace.Tracer, name string) (context.Context, trace.Span) {
+	return tracer.Start(ctx, name)
+}
+
+// Starts a span with a single field.
+func StartSpan1(ctx context.Context, tracer trace.Tracer, name string, field string, value interface{}) (context.Context, trace.Span) {
```
VinozzZ, line 68: Maybe this could be `StartSpanWith`

internal/otelutil/otel_tracing.go
```diff
@@ -27,30 +27,51 @@ func AddException(span trace.Span, err error) {
...
+	return tracer.Start(ctx, name, trace.WithAttributes(Attributes(map[string]interface{}{field: value})...))
+}
+
+// Starts a span with multiple fields.
+func StartSpanWith(ctx context.Context, tracer trace.Tracer, name string, fields map[string]interface{}) (context.Context, trace.Span) {
```
VinozzZ, line 73: This could be `StartSpanWithMulti`

kentquirk: I called it `StartSpanMulti`.

centralstore/local_store.go
```diff
@@ -121,6 +124,9 @@ func (lrs *LocalStore) changeTraceState(traceID string, fromState, toState Centr
 // during shutdown. WriteSpan is expecting to be called from an asynchronous
 // process and will only return an error if the span could not be written.
 func (lrs *LocalStore) WriteSpan(ctx context.Context, span *CentralSpan) error {
+	_, spanWrite := otelutil.StartSpan(ctx, lrs.Tracer, "LocalStore.WriteSpan")
```
TylerHelmuth, line 127: Should we add a field to the span to record the state?

centralstore/local_store.go
```diff
@@ -191,6 +197,8 @@ func (lrs *LocalStore) WriteSpan(ctx context.Context, span *CentralSpan) error {
 // populated. Normally this call will be made after Refinery has been asked
 // to make a trace decision.
 func (lrs *LocalStore) GetTrace(ctx context.Context, traceID string) (*CentralTrace, error) {
+	_, span := otelutil.StartSpan(ctx, lrs.Tracer, "LocalStore.GetTrace")
```
TylerHelmuth, line 200: Should we record whether or not the trace was found?

centralstore/local_store.go
```diff
@@ -210,6 +218,8 @@ func (lrs *LocalStore) GetTrace(ctx context.Context, traceID string) (*CentralTr
 // after this call (although kept spans will have counts updated when late
 // spans arrive).
 func (lrs *LocalStore) GetStatusForTraces(ctx context.Context, traceIDs []string) ([]*CentralTraceStatus, error) {
+	_, span := otelutil.StartSpan(ctx, lrs.Tracer, "LocalStore.GetStatusForTraces")
+	defer span.End()
```
TylerHelmuth, line 222: Unrelated to this PR, but in the below for loop, since we know the exact length of `statuses` and that we'll be completely filling it, I think it is faster to do `statuses[i] = ` than `append`.

centralstore/local_store.go
```diff
@@ -247,6 +259,8 @@ func (lrs *LocalStore) GetTracesForState(ctx context.Context, state CentralTrace
 // ReadyForDecision state. These IDs are moved to the AwaitingDecision state
 // atomically, so that no other refinery will be assigned the same trace.
 func (lrs *LocalStore) GetTracesNeedingDecision(ctx context.Context, n int) ([]string, error) {
+	_, span := otelutil.StartSpan(ctx, lrs.Tracer, "LocalStore.GetTracesNeedingDecision")
+	defer span.End()
```
TylerHelmuth, line 263: Same as above, we could do `traceids[i] = ` instead.

centralstore/local_store.go
```diff
@@ -225,6 +235,8 @@ func (lrs *LocalStore) GetStatusForTraces(ctx context.Context, traceIDs []string
 
 // GetTracesForState returns a list of trace IDs that match the provided status.
 func (lrs *LocalStore) GetTracesForState(ctx context.Context, state CentralTraceState) ([]string, error) {
+	_, span := otelutil.StartSpan(ctx, lrs.Tracer, "LocalStore.GetTracesForState")
+	defer span.End()
```
TylerHelmuth, line 239: Same as above, we could do `traceids[i] =`  instead.

centralstore/local_store.go
kentquirk: Because I'm That Guy, I decided to try to verify this claim with a couple of benchmarks, and I think that the modern Go compiler is just as good in either case. These literally run to within half a nanosecond of each other. Since I prefer the append model in case of an early exit, I'd rather do that.

```go
type testStruct struct {
	foo string
	bar int
}

func BenchmarkAssign(b *testing.B) {
	data := make([]*testStruct, b.N)
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		item := &testStruct{foo: "hi", bar: 1}
		data[i] = item
	}
}

func BenchmarkAppend(b *testing.B) {
	data := make([]*testStruct, 0, b.N)
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		item := &testStruct{foo: "hi", bar: 1}
		data = append(data, item)
	}
}
```

VinozzZ (approved):

PR #1109, opened 2024-04-29
feat: write span in batch to redis

VinozzZ:

To increase throughput for writing spans into redis, we should write spans in batches instead of one by one.

- Change `WriteSpan` to `WriteSpans`
- use transaction for independent redis calls for `addStatus`
- Defined a constant batch size of 10 as a starting point for WriteSpans


centralstore/local_store.go
```diff
@@ -123,69 +123,70 @@ func (lrs *LocalStore) changeTraceState(traceID string, fromState, toState Centr
...
 	lrs.mutex.Lock()
 	defer lrs.mutex.Unlock()
 
-	// first let's check if we've already processed and dropped this trace; if so, we're done and
-	// can just ignore the span.
-	if lrs.DecisionCache.Dropped(span.TraceID) {
-		otelutil.AddSpanField(spanWrite, "dropped", true)
-		return nil
-	}
+	for _, span := range spans {
```
VinozzZ, line 133: I just moved all the logic into a loop here

VinozzZ:
It might be good to check for performance before we merge this.

centralstore/smartwrapper.go
```diff
@@ -113,8 +115,24 @@ func (w *SmartWrapper) WriteSpan(ctx context.Context, span *CentralSpan) error {
...
-	for span := range w.spanChan {
-		err := w.BasicStore.WriteSpan(ctx, span)
+	for first := range w.spanChan {
+		var spans []*CentralSpan
+		spans = append(spans, first)
+
+	Remaining:
+		for len(spans) < WriteSpanBatchSize { // For control maximum size of batch
+			select {
+			case span, ok := <-w.spanChan:
```
TylerHelmuth, line 125: Will this have an opposite reaction where refinery will pause on processing spans when the rate of incoming spans is slow and it takes a while to reach `WriteSpanBatchSize`?

kentquirk:
I added my `exists` tweak to this PR.

kentquirk (approved):
I think this is ready to go now.

PR #1108, opened 2024-04-29
Check for traces before apply state change

TylerHelmuth:
## Which problem is this PR solving?

- Ensures that state change requests only happen if traceids are provided

## Short description of the changes

- Updates functions to check the length of the trace ids slice.
- Added unit tests

PR #1110, opened 2024-04-30
feat: performance tweaks

kentquirk:

- Addresses some performance issues with talking to Redis. Before these changes, some calls to GetStatusForTraces were taking 800ms when called with 5000 traceIDs. With this, 13000 traceIDs takes a little more than 100ms. I'm also seeing Refinery lasting quite a bit longer under heavy loads, although it still has problems; I have more ideas but I'd like to see this, plus #1108 and #1109 land first.
- This doesn't solve all performance issues, but it definitely helps. See me for links to HC queries that show it.

- GetStatusForTraces now takes a list of the statuses it cares about, and only returns traces that have those values. This signature change needed to be made in all the CentralStore interfaces, and implemented in both redis and local stores.
- getTraceStatuses (which is called by GetStatusForTraces) now splits up its redis requests into multiple goroutines that run in parallel, using a fan-out/fan-in mechanism.
- Defaults for redis pool sizes have been increased to 15.
- Lots of tests have been fixed.

### DIFF
--- a/centralstore/centralstore.go
+++ b/centralstore/centralstore.go
@@ -182,7 +182,7 @@ type SmartStorer interface {
 	// traces with a state of DecisionKeep or DecisionDrop should be considered
 	// to be final and appropriately disposed of; the central store will not
 	// change the state of these traces after this call.
-	GetStatusForTraces(ctx context.Context, traceIDs []string, statesToCheck []CentralTraceState) ([]*CentralTraceStatus, error)
+	GetStatusForTraces(ctx context.Context, traceIDs []string, statesToCheck ...CentralTraceState) ([]*CentralTraceStatus, error)
 
 	// GetTracesForState returns a list of trace IDs that match the provided status.
 	GetTracesForState(ctx context.Context, state CentralTraceState) ([]string, error)
@@ -278,7 +278,7 @@ type BasicStorer interface {
 	// traces with a state of DecisionKeep or DecisionDrop should be considered
 	// to be final and appropriately disposed of; the central store will not
 	// change the state of these traces after this call.
-	GetStatusForTraces(ctx context.Context, traceIDs []string, statesToCheck []CentralTraceState) ([]*CentralTraceStatus, error)
+	GetStatusForTraces(ctx context.Context, traceIDs []string, statesToCheck ...CentralTraceState) ([]*CentralTraceStatus, error)
 
 	// GetTracesForState returns a list of trace IDs that match the provided status.
 	GetTracesForState(ctx context.Context, state CentralTraceState) ([]string, error)
--- a/centralstore/centralstore.go
+++ b/centralstore/centralstore.go
@@ -175,15 +175,14 @@ type SmartStorer interface {
 	// its state will not be changed.
 	GetTrace(ctx context.Context, traceID string) (*CentralTrace, error)
 
-	// GetStatusForTraces returns the current state for a list of traces,
-	// including any reason information if the trace decision has been made and
-	// it was to keep the trace. If a trace is not found, it will be returned as
-	// Status:Unknown. This should be considered to be a bug in the central
-	// store, as the trace should have been created when the first span was
-	// added. Any traces with a state of DecisionKeep or DecisionDrop should be
-	// considered to be final and appropriately disposed of; the central store
-	// will not change the state of these traces after this call.
-	GetStatusForTraces(ctx context.Context, traceIDs []string) ([]*CentralTraceStatus, error)
+	// GetStatusForTraces returns the current state for a list of traces if they
+	// match any of the states passed in, including any reason information if
+	// the trace decision has been made and it was to keep the trace. If a trace
+	// is not found in any of the listed states, it will be not be returned. Any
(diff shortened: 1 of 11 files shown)


centralstore/redis_store.go
```diff
@@ -329,13 +331,12 @@ func (r *RedisBasicStore) GetStatusForTraces(ctx context.Context, traceIDs []str
 		if err != nil {
 			return nil, err
 		}
+		pendingTraceIDs = slices.DeleteFunc(pendingTraceIDs, func(id string) bool {
```
VinozzZ, line 334: Wow, I didn't know `slices` package contains this helper function. This is neat

centralstore/redis_store.go
```diff
@@ -745,29 +737,89 @@ func (t *tracesStore) keepTrace(ctx context.Context, conn redis.Conn, status []*
...
+					if errors.Is(err, redis.ErrKeyNotFound) {
+						continue
+					}
+					statusSpan.RecordError(err)
+					continue
+				}
+				faninChan <- normalizeCentralTraceStatusRedis(status)
+				found++
+				// just to make sure we give all the goroutines a chance to run
+				time.Sleep(1 * time.Microsecond)
```
VinozzZ, line 805: Wow, does the go scheduler not able to manage this appropriately?

centralstore/centralstore.go
```diff
@@ -272,17 +271,14 @@ type BasicStorer interface {
...
-	// spans arrive).
-	GetStatusForTraces(ctx context.Context, traceIDs []string) ([]*CentralTraceStatus, error)
+	// GetStatusForTraces returns the current state for a list of traces if they
+	// match any of the states passed in, including any reason information if
+	// the trace decision has been made and it was to keep the trace. If a trace
+	// is not found in any of the listed states, it will be not be returned. Any
+	// traces with a state of DecisionKeep or DecisionDrop should be considered
+	// to be final and appropriately disposed of; the central store will not
+	// change the state of these traces after this call.
+	GetStatusForTraces(ctx context.Context, traceIDs []string, statesToCheck []CentralTraceState) ([]*CentralTraceStatus, error)
```
VinozzZ, line 281: We could do so that we don't have to construct a `[]CentralTraceState` each time from the caller
```suggestion
	GetStatusForTraces(ctx context.Context, traceIDs []string, statesToCheck ...CentralTraceState) ([]*CentralTraceStatus, error)
```

centralstore/redis_store.go
kentquirk: Oh! I can remove this now; I was trying to figure out why I only got 3 goroutines running, and then realized it was because the redis connection pool was so limited.

centralstore/centralstore.go
kentquirk: OH, that's a great idea. Will fix!

VinozzZ (approved):

collect/central_collector.go
```diff
@@ -410,7 +408,7 @@ func (c *CentralCollector) processTraces(ctx context.Context) error {
 		return nil
 	}
 
-	statuses, err := c.Store.GetStatusForTraces(ctx, ids)
+	statuses, err := c.Store.GetStatusForTraces(ctx, ids, centralstore.DecisionKeep, centralstore.DecisionDrop)
```
TylerHelmuth, line 411: Should the default case below now error? This function should no longer ever return something that isn't Keep or Drop

collect/central_collector.go
```diff
@@ -459,7 +457,7 @@ func (c *CentralCollector) makeDecision(ctx context.Context) error {
 	if len(tracesIDs) == 0 {
 		return nil
 	}
-	statuses, err := c.Store.GetStatusForTraces(ctx, tracesIDs)
+	statuses, err := c.Store.GetStatusForTraces(ctx, tracesIDs, centralstore.AwaitingDecision)
```
TylerHelmuth, line 460: Should the check below now error? This function should no longer ever return something that isn't Awaiting

centralstore/local_store.go
```diff
@@ -212,27 +213,22 @@ func (lrs *LocalStore) GetTrace(ctx context.Context, traceID string) (*CentralTr
...
+// change the state of these traces after this call.
+func (lrs *LocalStore) GetStatusForTraces(ctx context.Context, traceIDs []string, statesToCheck ...CentralTraceState) ([]*CentralTraceStatus, error) {
 	_, span := otelutil.StartSpan(ctx, lrs.Tracer, "LocalStore.GetStatusForTraces")
 	defer span.End()
 	lrs.mutex.RLock()
 	defer lrs.mutex.RUnlock()
 	var statuses = make([]*CentralTraceStatus, 0, len(traceIDs))
 	for _, traceID := range traceIDs {
-		if state, status := lrs.findTraceStatus(traceID); state != Unknown {
+		if state, status := lrs.findTraceStatus(traceID); slices.Contains(statesToCheck, state) {
```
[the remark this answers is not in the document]
kentquirk: Good call.

TylerHelmuth (approved):

PR #1112, opened 2024-05-01
feat: Performance tuning, more telemetry

kentquirk:
## Which problems is this PR solving?

Along the route to tuning performance, this:
- Renames Pause->Cycle in a couple config values, as Pause was inaccurate
- Updates the Count function in metrics to work with updowns as well as counters, so that we can count up/down by more than 1
- Adds some metrics to the spanstore as well as to the redis store
- Uses config to define the number of goroutines used to talk to redis in a chatty subroutine

### DIFF
--- a/config/metadata/configMeta.yaml
+++ b/config/metadata/configMeta.yaml
@@ -1029,7 +1029,7 @@ groups:
         firstversion: v2.6
         type: int
         valuetype: nondefault
-        default: 3
+        default: 15
         reload: false
         validations:
           - type: minimum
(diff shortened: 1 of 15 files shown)


config/metadata/configMeta.yaml
```diff
@@ -1043,15 +1043,13 @@ groups:
         firstversion: v2.6
         type: int
         valuetype: nondefault
-        default: 5
+        default: 30
```
VinozzZ, line 1046: When I ran with 30 connections, the `redisstore_conn_wait` is still pretty high. Maybe we need even more connections available https://ui-dogfood.honeycomb.io/kent_test/environments/thenewworld/datasets/refinery_metrics/result/zGi8XYhTKpE

TylerHelmuth (approved):

PR #1113, opened 2024-05-01
fix: improve performance for getting trace state from redis

VinozzZ:

Previously, redis_store retrieves trace state by checking if a trace id exists in one of the state list. That means in worst case, refinery needs 6 round trips to redis in order to get trace state information.
Given `WriteSpans` is on a hot path, it's crucial to make sure refinery can write spans to redis as quick as possible.
Since refinery is able to update the trace status hash in redis during a trace state change through lua script, we can directly get the state information from the trace status hash instead. That will reduce 6 round trip to 1 per trace in the worst case scenario.

It does mean the state information may not be the most up-to-date. That's ok since the state machine is able to handle stale state change.

The commits are broken into individual changes I did in this PR:

- [add smartstore_span_queue_length to stress relief calculation](https://github.com/honeycombio/refinery/commit/240be13101ad22b123feb23ac4f6a32665f71ee2)

- [make sure to report the actual value for redis connection wait duration](https://github.com/honeycombio/refinery/commit/8e1fa6c9f12c83d6a0853a59bc9c216d10c37478)

- [Get trace state from trace status instead of the trace state sorted s…](https://github.com/honeycombio/refinery/commit/09f5f59ce1091ce4ac66c39ed3a4778c5ff550e1) 

- [Fix cluster stress level calculation when no refinery is ready to acc…](https://github.com/honeycombio/refinery/commit/adbc2714ea2f9f8825ee61d4b4d29a20e53aad97) 

- [add metrics for stress levels](https://github.com/honeycombio/refinery/commit/be37eaa6629cf684ccd1b6cbce169f545e9bc2bc)


centralstore/redis_store.go
```diff
@@ -131,7 +131,7 @@ func (r *RedisBasicStore) RecordMetrics(ctx context.Context) error {
 	r.Metrics.Gauge(metricsPrefixConnection+"active", connStats.ActiveCount)
 	r.Metrics.Gauge(metricsPrefixConnection+"idle", connStats.IdleCount)
 	r.Metrics.Gauge(metricsPrefixConnection+"wait", connStats.WaitCount)
-	r.Metrics.Histogram(metricsPrefixConnection+"wait_duration", connStats.WaitDuration)
+	r.Metrics.Histogram(metricsPrefixConnection+"wait_duration_ms", connStats.WaitDuration.Microseconds())
```
TylerHelmuth, line 134: if we use the `ms` suffix we should use `Milliseconds`.

centralstore/redis_store.go
```diff
@@ -183,6 +183,18 @@ func (r *RedisBasicStore) WriteSpans(ctx context.Context, spans []*CentralSpan)
 	conn := r.RedisClient.Get()
 	defer conn.Close()
 
+	traceIDs := make([]string, 0, len(spans))
+	for _, span := range spans {
+		traceIDs = append(traceIDs, span.TraceID)
+
```
TylerHelmuth, line 189:
```suggestion
```

centralstore/redis_store.go
```diff
@@ -183,6 +183,18 @@ func (r *RedisBasicStore) WriteSpans(ctx context.Context, spans []*CentralSpan)
 	conn := r.RedisClient.Get()
 	defer conn.Close()
 
+	traceIDs := make([]string, 0, len(spans))
+	for _, span := range spans {
+		traceIDs = append(traceIDs, span.TraceID)
```
TylerHelmuth, line 188: Is it possible for 2 spans in the span list to have the same TraceID? Should we use a map/set instead of a list to ensure traceid is unique?

VinozzZ: Yes! Good catch! I will make the change

centralstore/redis_store.go
```diff
@@ -183,6 +183,16 @@ func (r *RedisBasicStore) WriteSpans(ctx context.Context, spans []*CentralSpan)
 	conn := r.RedisClient.Get()
 	defer conn.Close()
 
+	states := make(map[string]CentralTraceState, len(spans))
+	for _, span := range spans {
+		states[span.TraceID] = Unknown
+	}
```
TylerHelmuth, line 189: Should we do this setup before we ask for a connection?

centralstore/redis_store.go
```diff
@@ -558,52 +508,43 @@ func (r *RedisBasicStore) RecordTraceDecision(ctx context.Context, trace *Centra
...
 
-	if r.states.exists(ctx, conn, AwaitingDecision, traceID) {
-		return AwaitingDecision
+	if cacheHitCount == len(states) {
+		return nil
 	}
 
-	if r.states.exists(ctx, conn, ReadyToDecide, traceID) {
-		return ReadyToDecide
+	results, err := r.traces.getTraceStates(ctx, conn, notFound)
```
TylerHelmuth, line 538: We end up looping through all the traceids in `notFound` inside `r.traces.getTraceStates`.  Instead of building a second list and looping through it again, can we do the logic in `r.traces.getTraceStates` in the above for loop if the traceid was not found?

TylerHelmuth: Or, even better, is it possible to collect all the not found traceIDs like you are and then make a single request to redis for all the states of thoses traceids in 1 query?

VinozzZ: I pipelined them into one request

redis/redis.go
```diff
@@ -847,6 +860,45 @@ func (c *DefaultConn) MemoryStats() (map[string]any, error) {
...
+		if err != nil {
+			return err
+		}
+		val, err := redis.String(reply, nil)
+		if errors.Is(err, redis.ErrNil) {
+			replies = append(replies, "")
+			return nil
+		}
+		if err != nil {
+			return err
```
TylerHelmuth, line 864: How to errors affect the pipelining? If any command in the pipe has an error will it ruin the whole batch?  Will it ruin the count that `c.receive` depends on?

redis/redis.go
```diff
@@ -847,6 +860,45 @@ func (c *DefaultConn) MemoryStats() (map[string]any, error) {
...
 
+func (c *DefaultConn) ReceiveStrings(n int) ([]string, error) {
+	replies := make([]string, 0, n)
+	err := c.receive(n, func(reply any, err error) error {
+		if err != nil {
+			return err
+		}
+		val, err := redis.String(reply, nil)
+		if errors.Is(err, redis.ErrNil) {
+			replies = append(replies, "")
```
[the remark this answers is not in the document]
VinozzZ: Unfortunately, we can't add the centralstore information here since this is in the lower layer redis package

redis/redis.go
VinozzZ: This is a callback function to c.receive. It doesn't modify the `count`, `c.receive` depends on.

The error here means something is wrong with redis and I'm think it's safer to bail on this whole batch then continue.

redis/redis.go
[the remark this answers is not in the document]
TylerHelmuth: If we set the string `"unknown"` can it be automatically like the other return string values?

redis/redis.go
TylerHelmuth: I am ok bailing on the whole batch.

redis/redis.go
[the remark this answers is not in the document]
VinozzZ: It could, but that means we are leaking the centralstore information into the redis package. It feels a bit weird to me

redis/redis.go
```diff
@@ -725,6 +726,18 @@ func (c *DefaultConn) GetStructHash(key string, val interface{}) error {
 	return redis.ScanStruct(values, val)
 }
 
+func (c *DefaultConn) GetStructsHash(key string, val interface{}) error {
```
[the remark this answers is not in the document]
VinozzZ: I was sure I removed this at some point lol It must came back when I was merging with the latest code from the feature branch

redis/redis.go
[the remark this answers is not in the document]
TylerHelmuth: Ok lets leave it as empty string

TylerHelmuth (approved):

PR #1114, opened 2024-05-01
fix: Clarify what has-root-span does

cartermp:
Noted in pollinators: https://honeycombpollinators.slack.com/archives/C01BZ3J7RMJ/p1714601065818559

### DIFF
--- a/rules_complete.yaml
+++ b/rules_complete.yaml
@@ -108,6 +108,8 @@ Samplers:
                 - Name: drop incomplete traces from the buggy service
                   Drop: true
                   Conditions:
+                    # NOTE: has-root-span only checks if the trace has a root span.
+                    # It does NOT act as a short-hand for evaluating the root span.
                     - Operator: has-root-span
                       Value: false
                     - Field: service.name
--- a/rules_complete.yaml
+++ b/rules_complete.yaml
@@ -109,7 +109,7 @@ Samplers:
                   Drop: true
                   Conditions:
                     # NOTE: has-root-span only checks if the trace has a root span.
-                    # It does NOT act as a short-hand for evaluating the root span.
+                    # It does NOT act as a shorthand for evaluating the root span; use the `root.` construct for that.
                     - Operator: has-root-span
                       Value: false
                     - Field: service.name
--- a/rules_conditions.md
+++ b/rules_conditions.md
@@ -207,8 +207,8 @@ Tests if the trace as a whole has a root span.
 
 The `Value` parameter can either be `true` or `false`.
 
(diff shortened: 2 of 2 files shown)


rules_conditions.md
```diff
@@ -201,6 +201,15 @@ For most cases, use `not-exists` in a rule with a scope of "span".
 WARNING: Rules can have `Scope: trace` or `Scope: span`; `not-exists` used with `Scope: trace` will be true if **any** single span in the entire trace matches the negative condition.
 This is almost never desired behavior.
 
+### has-root-span
+
+Tests if the trace as a whole has a root span.
+
+The `Value` parameter can either be `true` or `false`.
+
+**NOTE:** `has-root-span` does not check if a given span is a root span.
```
VinozzZ, line 210: nit - it looks like other places uses `WARNING` for calling out special characteristic of rule conditions. Maybe we can change `**NOTE**` to `WARNING:` here as well

cartermp:
@VinozzZ done!

rules_conditions.md
```diff
@@ -201,6 +201,15 @@ For most cases, use `not-exists` in a rule with a scope of "span".
...
 This is almost never desired behavior.
 
+### has-root-span
+
+Tests if the trace as a whole has a root span.
+
+The `Value` parameter can either be `true` or `false`.
+
+WARNING: `has-root-span` does not check if a given span is a root span.
+It only checks if a trace has a root span.
```
kentquirk, line 211:
```suggestion
NOTE: `has-root-span` does not check if a given span **is** a root span,
it checks if the containing trace **has** a root span.
```

rules_complete.yaml
```diff
@@ -108,6 +108,8 @@ Samplers:
                 - Name: drop incomplete traces from the buggy service
                   Drop: true
                   Conditions:
+                    # NOTE: has-root-span only checks if the trace has a root span.
+                    # It does NOT act as a short-hand for evaluating the root span.
```
kentquirk, line 112:
```suggestion
                    # It does NOT act as a shorthand for evaluating the root span; use the `root.` construct for that.
```

kentquirk (commented):
I tweaked it a bit. What do you think?

rules_conditions.md
kentquirk: Huh, my suggestion includes the opposite of Yingrong's -- I feel like "WARNING" is too strident and want this to be a note.

kentquirk (approved):

danielcompton:
Thank you! 😊

PR #1115, opened 2024-05-03
fix: prevent data race when retrieve information about cuckoo cache

VinozzZ:

In order to record metrics about dropped traces in cuckoo cache, we need to access the cuckoofilter object periodically in a goroutine. Since cuckoofilter itself is not concurrent safe, we need to add locks around accessing it.

Add mutex when accessing cuckoofilter for current count and load factor


kentquirk (approved):
🔐

PR #1116, opened 2024-05-03
feat: generic fanout function, used in place of existing fanout mess

kentquirk:

I had written a fanout/fanin system to improve the speed of talking to redis. It was pretty complex. I realized I could genericize it and
- make it easier to understand by separating the technique from the details of the work
- probably reuse it elsewhere if it stood alone

- Create Fanout, FanoutToMap, EasyFanout, and EasyFanoutToMap functions
- Write some tests
- Use FanoutToMap in getTraceStatuses

### DIFF
--- a/centralstore/redis_store.go
+++ b/centralstore/redis_store.go
@@ -5,14 +5,13 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"math/rand"
 	"slices"
 	"sort"
-	"sync"
 	"time"
 
 	"github.com/honeycombio/refinery/collect/cache"
 	"github.com/honeycombio/refinery/config"
+	"github.com/honeycombio/refinery/generics"
 	"github.com/honeycombio/refinery/internal/otelutil"
 	"github.com/honeycombio/refinery/metrics"
 	"github.com/honeycombio/refinery/redis"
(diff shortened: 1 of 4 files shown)


generics/fanout.go
```diff
@@ -0,0 +1,163 @@
...
+// The factory takes an integer (the worker number) and constructs a function of
+// type func(T) U that processes a single input and produces a single output. It
+// also constructs a cleanup function, which may be nil. The cleanup function is
+// called once for each worker, after the worker has completed processing all of
+// its inputs. It is given the same index as the corresponding worker factory.
+//
+// If predicate is not nil, it will only add the output to the result slice if
+// the predicate returns true. It will fan out the input to the worker function
+// in parallel, and fan in the results to the output slice.
+func Fanout[T, U any](input []T, parallelism int, workerFactory func(int) (worker func(T) U, cleanup func(int)), predicate func(U) bool) []U {
```
TylerHelmuth, line 20: I don't see this function used anywhere, should we hold off adding it until we have a use case?

centralstore/redis_store.go
```diff
@@ -762,84 +761,61 @@ func (t *tracesStore) getTraceStatuses(ctx context.Context, client redis.Client,
...
 				"num_queries": queries,
 				"num_found":   found,
 			})
-		}(i)
+			span.End()
+		}
+		return worker, cleanup
+	}
+	// this filters out the nil statuses
+	predicate := func(s *CentralTraceStatus) bool {
```
TylerHelmuth, line 810: I don't see the previous implementation filtering nil statuses, what is the impact of dropping these? Is a nil status an error or does it mean redis didnt have that trace id anymore?

generics/fanout.go
kentquirk: It's not yet, but this is in a generics folder intended for broader use, AND I expect to find other uses for it within Refinery.

centralstore/redis_store.go
kentquirk: The previous implementation did it differently -- the worker had a loop that it `continue`d if the result was meaningless. In this one, it adds nil and we filter it outside the worker.

PR #1117, opened 2024-05-03
fix: add sender hostname in span metadata

VinozzZ:

add metadata to every span regarding which refinery send it to upstream

- add `meta.refinery.sender.local_hostname`

### DIFF
--- a/collect/central_collector.go
+++ b/collect/central_collector.go
@@ -797,6 +797,10 @@ func (c *CentralCollector) send(status *centralstore.CentralTraceStatus) {
 			sp.Data[k] = v
 		}
 
+		if c.hostname != "" && c.Config.GetAddHostMetadataToTrace() {
+			sp.Data["meta.refinery.sender.local_hostname"] = c.hostname
+		}
+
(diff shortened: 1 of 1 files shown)


TylerHelmuth (approved):

collect/central_collector.go
```diff
@@ -797,6 +797,10 @@ func (c *CentralCollector) send(status *centralstore.CentralTraceStatus) {
 			sp.Data[k] = v
 		}
 
+		if c.hostname != "" && c.Config.GetAddHostMetadataToTrace() {
+			sp.Data["meta.refinery.sender.local_hostname"] = c.hostname
```
TylerHelmuth, line 801: nit: using some flavor of `host.name` conforms better to the otel host semconv: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/resource/host.md

VinozzZ: ohh, so like `meta.refinery.sender.host.name` ?

TylerHelmuth: Ya I like that better

TylerHelmuth (approved):

PR #1122, opened 2024-05-07
feat: Add Parallelism to config, getTraceStatuses uses it

kentquirk:

- getTraceStatuses needs to know how many goroutines to use; trying to be smart about it based on pool size has failed. Let's just do a config value.

- Add new config value
- fix a bunch of tests that were setting pool size to set parallelism


VinozzZ (approved):
LGTM!

PR #1123, opened 2024-05-07
feat: add WriteSpanBatchSize to central store options as a configurable value

VinozzZ:

Enable quick experiment with the batch size for writing spans to redis store

- fixed an incorrect metric name for collector_send_trace
- add a new config value for WriteSpanBatchSize and default it to 20

### DIFF
--- a/config/mock.go
+++ b/config/mock.go
@@ -569,5 +569,9 @@ func (f *MockConfig) GetCentralStoreOptions() SmartWrapperOptions {
 	f.Mux.RLock()
 	defer f.Mux.RUnlock()
 
+	if f.StoreOptions.WriteSpanBatchSize == 0 {
+		f.StoreOptions.WriteSpanBatchSize = 20
+	}
+
 	return f.StoreOptions
 }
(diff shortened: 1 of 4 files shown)


config/metadata/configMeta.yaml
```diff
@@ -1837,6 +1837,17 @@ groups:
...
+      - name: WriteSpanBatchSize
+        firstVersion: v2.6
+        type: int
+        valuetype: nondefault
+        default: 20
+        reload: true
+        summary: is the maximum number of spans to be included for writing to basic store in a single request.
+        description: >
+          This setting controls the number of spans sent to the basic store in
+          central store when Refinery is processing incoming spans.
```
kentquirk, line 1849:
```suggestion
          This setting controls the number of spans sent to the basic store at
          the same time when Refinery is processing incoming spans.
```

config/mock.go
```diff
@@ -569,5 +569,9 @@ func (f *MockConfig) GetCentralStoreOptions() SmartWrapperOptions {
 	f.Mux.RLock()
 	defer f.Mux.RUnlock()
 
+	if f.StoreOptions.WriteSpanBatchSize == 0 {
```
kentquirk, line 572: This is a good idea. I should have done this with Parallelism, etc.

PR #1124, opened 2024-05-07
feat: record decision and send rates

kentquirk:

- Record how fast the decider and the sender are actually operating

- for each invocation of the decider and sender, record histograms of the number of traces actually considered (and for the decider, actually handled)

### DIFF
--- a/collect/central_collector.go
+++ b/collect/central_collector.go
@@ -146,6 +146,9 @@ func (c *CentralCollector) Start() error {
 	c.Metrics.Register("kept_from_stress", "counter")
 	c.Metrics.Register("collector_send_trace", "counter")
 	c.Metrics.Register("collector_decide_trace", "counter")
+	c.Metrics.Register("decider_decided_per_second", "histogram")
+	c.Metrics.Register("decider_considered_per_second", "histogram")
+	c.Metrics.Register("sender_considered_per_second", "histogram")
 
 	if c.Config.GetAddHostMetadataToTrace() {
 		if hostname, err := os.Hostname(); err == nil && hostname != "" {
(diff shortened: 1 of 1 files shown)


collect/central_collector.go
```diff
@@ -423,11 +433,16 @@ func (c *CentralCollector) sendTraces(ctx context.Context) error {
...
+			tracesConsidered++
 
 		case centralstore.DecisionDrop:
 			c.SpanCache.Remove(status.TraceID)
+			tracesConsidered++
 		default:
-			return fmt.Errorf("unexpected state %s for trace %s", status.State, status.TraceID)
+			// we don't want to send traces that are in any other state;
+			// it's also an error, but ending the loop is not what we want
+			c.Logger.Error().Logf("unexpected state %s for trace %s", status.State, status.TraceID)
```
VinozzZ, line 444: maybe we can make the state and trace id into log fields here?

collect/central_collector.go
```diff
@@ -146,6 +146,9 @@ func (c *CentralCollector) Start() error {
 	c.Metrics.Register("kept_from_stress", "counter")
 	c.Metrics.Register("collector_send_trace", "counter")
 	c.Metrics.Register("collector_decide_trace", "counter")
+	c.Metrics.Register("decider_decided_per_second", "histogram")
```
VinozzZ, line 149: I'm torn between `sender` v.s. `send` in our metrics. It feels more straightforward to use `send` and `decide` but I can go either way. Just want to make sure we have a consistent pattern here

kentquirk: I was distinguishing between the noun and the verb here, but I agree we should settle on a pattern. I've put it in the notes, let's decide on it together and not in this PR

VinozzZ (approved):

PR #1125, opened 2024-05-07
fix: clean up metrics and logs

VinozzZ:

Logs that includes trace ids in its message body is noisy.
We are also missing stress activation metric

- move trace/span IDs to log fields
- add stress_relief_activated metric

### DIFF
--- a/collect/central_collector.go
+++ b/collect/central_collector.go
@@ -533,9 +533,9 @@ func (c *CentralCollector) makeDecision(ctx context.Context) error {
 		// get sampler key (dataset for legacy keys, environment for new keys)
 		selector := stateMap[trace.TraceID].SamplerSelector
 		logFields := logrus.Fields{
-			"trace_id": trace.TraceID,
+			"trace_id":         trace.TraceID,
+			"sampler_selector": selector,
 		}
-		logFields["sampler_selector"] = selector
 
 		// use sampler key to find sampler; create and cache if not found
 		c.mut.RLock()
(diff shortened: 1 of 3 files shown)


collect/central_collector.go
```diff
@@ -619,9 +624,12 @@ func (c *CentralCollector) processSpan(sp *types.Span) error {
 		c.Metrics.Increment("span_processed")
 		c.Metrics.Down("spans_waiting")
 	}()
+	logFields := logrus.Fields{
+		"trace_id": sp.TraceID,
+	}
 	err := c.SpanCache.Set(sp)
 	if err != nil {
-		c.Logger.Error().Logf("error adding span with trace ID %s to cache: %s", sp.TraceID, err)
+		c.Logger.Error().WithFields(logFields).Logf("error adding span to cache: %s", err)
```
kentquirk, line 632: why not just "WithField" here?

collect/central_collector.go
```diff
@@ -545,7 +550,7 @@ func (c *CentralCollector) makeDecision(ctx context.Context) error {
 
 		status, ok := stateMap[trace.TraceID]
 		if !ok {
-			c.Logger.Error().Logf("trace %s not found in state map", trace.TraceID)
+			c.Logger.Error().WithFields(logFields).Logf("trace not found in state map")
```
kentquirk, line 553: I can't comment on line 538 but that should go inside the logFields declaration

kentquirk (approved):

TylerHelmuth (approved):

PR #1128, opened 2024-05-08
feat: test sampler key stuff with mock sampler

kentquirk:

- Trying to write a test to check sampler behavior and sampler keys
- Doing this with a new MockSampler and a test that uses it

- Create a MockSampler; it always returns true and adds the sampler key it calculated
- Create a MockSamplerConfig
- Teach the SamplerFactory about it
- Write a test that uses it -- but the test isn't working so it's not finished.

### DIFF
--- a/app/app_test.go
+++ b/app/app_test.go
@@ -105,7 +105,7 @@ func newStartedApp(
 		GetTraceTimeoutVal:       10 * time.Millisecond,
 		GetMaxBatchSizeVal:       500,
 		GetSamplerTypeVal:        &config.DeterministicSamplerConfig{SampleRate: 1},
-		SendTickerVal:            200 * time.Microsecond,
+		SendTickerVal:            20 * time.Millisecond,
 		PeerManagementType:       "file",
 		GetUpstreamBufferSizeVal: 10000,
 		GetRedisDatabaseVal:      redisDB,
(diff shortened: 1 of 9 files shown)


app/app_test.go
```diff
@@ -354,6 +365,65 @@ func TestHostMetadataSpanAdditions(t *testing.T) {
...
+
+	require.Eventually(t, func() bool {
+		events := sender.Events()
+		return len(events) == 5
+	}, 5*time.Second, 100*time.Millisecond)
+
+	require.EventuallyWithT(t, func(collect *assert.CollectT) {
+		events := sender.Events()
+
+		// these are the wrong asserts for this test, but I don't care as long as they fail; will fix
```
VinozzZ, line 415: Oops, I forgot to delete this comment

centralstore/local_store.go
```diff
@@ -168,6 +168,7 @@ spanLoop:
 		}
 
 		// Add the span to the trace; this works even if the trace has no spans yet
+		fmt.Printf("Adding span %s to trace %s\n", span.SpanID, span.TraceID)
```
VinozzZ: do we want to keep this print line?

types/event.go
```diff
@@ -296,8 +296,10 @@ func IsLegacyAPIKey(apiKey string) bool {
...
 }
 
-func GenerateSpanID(traceID string) string {
-	ts := wyhash.Hash([]byte("next"), hashSeed)
-	h := wyhash.Hash([]byte(traceID), uint64(ts))
-	return fmt.Sprintf("%016x", h)
+func GenerateSpanID() string {
+	id, err := uuid.NewV7()
+	if err != nil {
+		return ""
```
VinozzZ, line 302: I don't know what would happen to cause this function to return an error, but maybe it's worth logging it out in the caller so that we don't mysteriously setting all span Id to `""`

kentquirk: A logger isn't available here but I can fall back to returning a random value as a string.

VinozzZ (approved):

PR #1129, opened 2024-05-08
fix: Root span test was inverted!

kentquirk:

- Traces were making weird decisions because we were marking root spans backwards.

- Invert the test so we're marking the root span properly.


TylerHelmuth (approved):
Programming is way too hard sometimes, I should just grow cacti for a living

TylerHelmuth:
Is there a test we can update for this?

kentquirk:
> Is there a test we can update for this?

There's not an obvious one since this is in the middle of the highly complex router function. I haven't found one yet but I'll look again in the morning.

PR #1132, opened 2024-05-08
feat: add MetricsCycleRate config option for redis

VinozzZ:

Through testing, we realized that getting metrics from redis causes network usage to go up. Adding a config value to allow different interval for retrieving redis metrics as we currently are not calculating stress level using any of these metrics.

- add MetricsCycleRate option in config

### DIFF
--- a/collect/central_collector.go
+++ b/collect/central_collector.go
@@ -121,7 +121,7 @@ func (c *CentralCollector) Start() error {
 	c.samplersByDestination = make(map[string]sample.Sampler)
 
 	// The cycles manage a periodic task and also provide some test hooks
-	c.metricsCycle = NewCycle(c.Clock, c.Config.GetSendTickerValue(), c.done)
+	c.metricsCycle = NewCycle(c.Clock, 2*c.Config.GetSendTickerValue(), c.done)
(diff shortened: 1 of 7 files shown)


collect/central_collector.go
```diff
@@ -121,7 +121,7 @@ func (c *CentralCollector) Start() error {
 	c.samplersByDestination = make(map[string]sample.Sampler)
 
 	// The cycles manage a periodic task and also provide some test hooks
-	c.metricsCycle = NewCycle(c.Clock, c.Config.GetSendTickerValue(), c.done)
+	c.metricsCycle = NewCycle(c.Clock, 2*c.Config.GetSendTickerValue(), c.done)
```
TylerHelmuth, line 124: Whats this multiplication doing?

VinozzZ: Maybe this is too confusing, I will remove it now that we have a separate config for redis metrics

kentquirk (approved):
Looks good!

collect/central_collector.go
kentquirk: I removed the comment in the PR description too

PR #1133, opened 2024-05-08
fix: make sure send_reason is attached to span metadata

VinozzZ:

`span.Metadata` is a `map[string]interface{}` type. We were checking the return value against an empty string, which is not always correct. That means if the `meta.refinery.send_reason` does not exist in the metadata map, we will never set the send_reason.

- Properly check empty value returned from metadata map and set `send_reason`

### DIFF
--- a/centralstore/redis_store.go
+++ b/centralstore/redis_store.go
@@ -636,6 +636,9 @@ func (t *tracesStore) addStatuses(ctx context.Context, conn redis.Conn, cspans [
 
 	commands := make([]redis.Command, 0, 3*len(cspans))
 	for _, span := range cspans {
+		if span.SpanID == "" {
+			continue
+		}
 
 		trace := &centralTraceStatusInit{
 			TraceID:    span.TraceID,
--- a/centralstore/redis_store.go
+++ b/centralstore/redis_store.go
@@ -646,6 +646,8 @@ func (t *tracesStore) addStatuses(ctx context.Context, conn redis.Conn, cspans [
 
 	commands := make([]redis.Command, 0, 3*len(cspans))
 	for _, span := range cspans {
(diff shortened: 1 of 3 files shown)


collect/central_collector.go
```diff
@@ -738,7 +735,7 @@ func (c *CentralCollector) checkAlloc() {
 		}
 		totalDataSizeSent += trace.DataSize
 		numOfTracesSent++
-		err := c.Store.WriteSpan(ctx, &centralstore.CentralSpan{TraceID: id})
+		err := c.Store.WriteSpan(ctx, &centralstore.CentralSpan{TraceID: id, IsRoot: true})
```
TylerHelmuth, line 738: How do we know this is always a root span?

VinozzZ: The idea here is that we want to force refinery to eject traces. The ejection mechanism is to pretend a root span has arrived and force traces starts to go through the state machine and have decision made for them

kentquirk: Please add that as a code comment. It's important.

kentquirk (approved):
I have one question, but otherwise looks good.

centralstore/redis_store.go
```diff
@@ -646,6 +646,9 @@ func (t *tracesStore) addStatuses(ctx context.Context, conn redis.Conn, cspans [
 
 	commands := make([]redis.Command, 0, 3*len(cspans))
 	for _, span := range cspans {
+		if span.SpanID == "" {
```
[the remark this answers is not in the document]
VinozzZ: This is to prevent we store the fake root span centra store sent because that's how we differentiate a signal span vs an actual span. I will add a comment here for this

PR #1135, opened 2024-05-09
feat: distribute workload for trace state management more efficiently

VinozzZ:

Currently, each refinery will retrieve all traces within a state to process them through the state machine in an interval. This approach creates a lot of duplicated work since multiple refineries can end up competing with each other on completing the task for the same trace ID.

- Randomize the trace IDs handed to each refinery
- Process 20% of the total incoming traces on each state ticker
- make `GetTracesNeedingDecision` atomic through lua script

### DIFF
--- a/centralstore/centralstore.go
+++ b/centralstore/centralstore.go
@@ -185,7 +185,7 @@ type SmartStorer interface {
 	GetStatusForTraces(ctx context.Context, traceIDs []string, statesToCheck ...CentralTraceState) ([]*CentralTraceStatus, error)
 
 	// GetTracesForState returns a list of trace IDs that match the provided status.
-	GetTracesForState(ctx context.Context, state CentralTraceState) ([]string, error)
+	GetTracesForState(ctx context.Context, state CentralTraceState, n int) ([]string, error)
 
 	// GetTracesNeedingDecision returns a list of up to n trace IDs that are in the
 	// ReadyForDecision state. These IDs are moved to the AwaitingDecision state
@@ -281,7 +281,7 @@ type BasicStorer interface {
 	GetStatusForTraces(ctx context.Context, traceIDs []string, statesToCheck ...CentralTraceState) ([]*CentralTraceStatus, error)
 
 	// GetTracesForState returns a list of trace IDs that match the provided status.
-	GetTracesForState(ctx context.Context, state CentralTraceState) ([]string, error)
+	GetTracesForState(ctx context.Context, state CentralTraceState, n int) ([]string, error)
 
 	// GetTracesNeedingDecision returns a list of up to n trace IDs that are in the
 	// ReadyForDecision state. These IDs are moved to the AwaitingDecision state
--- a/centralstore/centralstore.go
+++ b/centralstore/centralstore.go
@@ -184,7 +184,8 @@ type SmartStorer interface {
 	// change the state of these traces after this call.
 	GetStatusForTraces(ctx context.Context, traceIDs []string, statesToCheck ...CentralTraceState) ([]*CentralTraceStatus, error)
 
-	// GetTracesForState returns a list of trace IDs that match the provided status.
+	// GetTracesForState returns a list of up to n trace IDs that match the provided status.
+	// If n is -1, return all matching traces.
 	GetTracesForState(ctx context.Context, state CentralTraceState, n int) ([]string, error)
 
 	// GetTracesNeedingDecision returns a list of up to n trace IDs that are in the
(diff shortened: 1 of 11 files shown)


centralstore/smartwrapper.go
```diff
@@ -157,7 +157,10 @@ func (w *SmartWrapper) manageTimeouts(ctx context.Context, timeout time.Duration
 	})
 	defer span.End()
 
-	st, err := w.BasicStore.GetTracesForState(ctx, fromState)
+	// process up to 20% of the channel size
+	n := w.Config.GetCentralStoreOptions().SpanChannelSize * 20 / 100
```
VinozzZ, line 161: I'm not sure if this is the right math but I don't think this value should be a user configured value

centralstore/centralstore.go
```diff
@@ -185,7 +185,7 @@ type SmartStorer interface {
 	GetStatusForTraces(ctx context.Context, traceIDs []string, statesToCheck ...CentralTraceState) ([]*CentralTraceStatus, error)
 
 	// GetTracesForState returns a list of trace IDs that match the provided status.
```
kentquirk, line 187:
```suggestion
	// GetTracesForState returns a list of up to n trace IDs that match the provided status.
	// If n is -1, returns all matching traces.
```

centralstore/local_store.go
```diff
@@ -297,7 +297,7 @@ func (lrs *LocalStore) GetStatusForTraces(ctx context.Context, traceIDs []string
 }
 
 // GetTracesForState returns a list of trace IDs that match the provided status.
```
kentquirk, line 299:
```suggestion
	// GetTracesForState returns a list of up to n trace IDs that match the provided status.
	// If n is -1, returns all matching traces.
```

centralstore/redis_store.go
```diff
@@ -349,7 +351,7 @@ func (r *RedisBasicStore) GetStatusForTraces(ctx context.Context, traceIDs []str
 
 }
 
-func (r *RedisBasicStore) GetTracesForState(ctx context.Context, state CentralTraceState) ([]string, error) {
+func (r *RedisBasicStore) GetTracesForState(ctx context.Context, state CentralTraceState, n int) ([]string, error) {
```
kentquirk, line 354:
```suggestion
// GetTracesForState returns a list of up to n trace IDs that match the provided status.
// If n is -1, returns all matching traces.
func (r *RedisBasicStore) GetTracesForState(ctx context.Context, state CentralTraceState, n int) ([]string, error) {
```

centralstore/redis_store.go
```diff
@@ -376,21 +383,20 @@ func (r *RedisBasicStore) GetTracesNeedingDecision(ctx context.Context, n int) (
...
 	if err != nil {
+		if errors.Is(err, redis.ErrKeyNotFound) {
+			return nil, nil
+		}
 		return nil, err
 	}
 
 	if len(traceIDs) == 0 {
-		return nil, nil
+		return nil, errors.New("failed to get traces for needing decisions")
```
kentquirk, line 396:
```suggestion
		return nil, errors.New("failed to get traces needing decisions")
```

centralstore/smartwrapper.go
kentquirk: I don't think this is the value we want to use. We'd like to grab a fraction of all the traces stored; the size of the input channel doesn't really affect how many we have stored. If we make this configurable we could play with it, but if we don't want a config let's try a number like, say, 400.

kentquirk (approved):
A couple minor things.

Out of curiosity, did you actually run any tests on the ZRANDOM behavior?

centralstore/smartwrapper.go
VinozzZ: Since we do get the metrics of the total keys in a given state from redis, maybe we can use that to calculate the size here?

VinozzZ:
> A couple minor things.
> 
> Out of curiosity, did you actually run any tests on the ZRANDOM behavior?

I did! 
I generated 10k record in a hash in redis. 
When the size is 400
amount of iteration to retrieve all keys in a hash: [5244, 4662, 4643, 3775, 4856, 4796, 4134, 4393, 6566, 3877, 5416, 4318, 4338, 5356, 4083, 3937, 4350]
1294145177 ns/op	137932273 B/op

When the size is 1000
amount of iteration to retrieve all keys in a hash: [1575, 1815, 1720, 2272, 1662, 1613, 2185, 1798, 1629, 2135, 1786, 1665, 1710, 2173, 1560, 3118, 1606, 2862]
1365432953 ns/op	158076788 B/op

centralstore/smartwrapper.go
VinozzZ: I made it to be configurable for now

kentquirk: OK; I think that's a better choice. We do use metrics in a load-bearing way for Stress Relief, but I feel differently about that because its job is to monitor the state of the system, and that's also what metrics do.

kentquirk (approved):
Looks good!

PR #1137, opened 2024-05-10
fix: make sure we can retrieve trace decisions made during stress relief

VinozzZ:

When a late span comes in after stress relief made a decision for the trace it belongs to, we should still be able to make the same decision for that late span. 
This will only works if there's no restart between the stress relief decision is made and the late span arrival. 

- Retrieve trace decision from decision cache in `GetStatusForTraces`
- Correctly record trace decision during stress relief
- add tests

### DIFF
--- a/centralstore/redis_store.go
+++ b/centralstore/redis_store.go
@@ -329,9 +329,7 @@ func (r *RedisBasicStore) GetStatusForTraces(ctx context.Context, traceIDs []str
 	for _, state := range statesToCheck {
 		validStates[state] = struct{}{}
 		// is any of the states we are looking for a decision state?
-		if (state == DecisionKeep || state == DecisionDrop) && !decisionMade {
-			decisionMade = true
-		}
+		decisionMade = decisionMade || state == DecisionKeep || state == DecisionDrop
 	}
 
 	statuses := make([]*CentralTraceStatus, 0, len(statusMapFromRedis))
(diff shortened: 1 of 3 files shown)


centralstore/redis_store.go
```diff
@@ -325,19 +325,52 @@ func (r *RedisBasicStore) GetStatusForTraces(ctx context.Context, traceIDs []str
 	}
 
 	validStates := make(map[CentralTraceState]struct{}, len(statesToCheck))
+	var decisionMade bool
 	for _, state := range statesToCheck {
 		validStates[state] = struct{}{}
+		if (state == DecisionKeep || state == DecisionDrop) && !decisionMade {
+			decisionMade = true
+		}
```
kentquirk, line 333:
```suggestion
		// is any of the states we're looking for a Decision state?
		decisionMade |= (state == DecisionKeep || state == DecisionDrop)
```
I find this clearer. Do you?

centralstore/redis_store.go
```diff
@@ -325,19 +325,52 @@ func (r *RedisBasicStore) GetStatusForTraces(ctx context.Context, traceIDs []str
...
 	}
 
 	statuses := make([]*CentralTraceStatus, 0, len(statusMapFromRedis))
 	for _, status := range statusMapFromRedis {
-		// only include statuses that are in the statesToCheck list
+		// only include statuses that are in the statesToCheck list with one
+		// exception: if a trace decision is made during stress relief, we need to
+		// find the trace state from the decision cache instead of redis
```
kentquirk, line 340:
```suggestion
		// only include statuses that are in the statesToCheck list.
		// exception: if a trace decision was made during stress relief, we need to
		// find the trace state from the decision cache instead of redis
```

kentquirk (approved):
Minor tweaks, but this is good!

PR #1138, opened 2024-05-13
fix: several bugs

kentquirk:
## Which problems is this PR solving?

This is fixing several issues that would be in separate PRs if we weren't pushing to get this out but we really just need to fix them all.

- Localstore wasn't giving the correct results because it wasn't handling the sampler selector correctly.
- There was an overly-chatty healthcheck
- Timing on the receiver healthcheck was wrong
- Bug in metadata over config validation
- Add some tracing to the receiver
- Fix status construction code
- Turn off the manual GC call because it seems like it's no longer needed with modern Go GC

### DIFF
--- a/centralstore/local_store.go
+++ b/centralstore/local_store.go
@@ -105,6 +105,8 @@ func (lrs *LocalStore) cleanup() {
 				lrs.mutex.Lock()
 				for _, traceID := range deletes {
 					delete(lrs.states[DecisionKeep], traceID)
+					// also remove it from the current traces list
+					delete(lrs.traces, traceID)
 				}
 				lrs.mutex.Unlock()
 			}
@@ -125,11 +127,21 @@ func (lrs *LocalStore) findTraceStatus(traceID string) (CentralTraceState, *Cent
 				return DecisionKeep, status
 			}
 			// we don't have more information, so we return what we have
-			status := NewCentralTraceStatus(traceID, DecisionKeep, lrs.Clock.Now())
+			// status := NewCentralTraceStatus(traceID, DecisionKeep, lrs.Clock.Now())
+			status := &CentralTraceStatus{
+				TraceID:   traceID,
+				State:     DecisionKeep,
+				Timestamp: lrs.Clock.Now(),
+			}
+
 			status.KeepReason = reason
 			return DecisionKeep, status
 		} else {
-			return DecisionDrop, NewCentralTraceStatus(traceID, DecisionDrop, lrs.Clock.Now())
+			return DecisionDrop, &CentralTraceStatus{
+				TraceID:   traceID,
+				State:     DecisionDrop,
+				Timestamp: lrs.Clock.Now(),
+			}
 		}
 	}
 	// wasn't in the cache, look in all the other states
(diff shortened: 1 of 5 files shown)


centralstore/local_store.go
```diff
@@ -125,11 +127,21 @@ func (lrs *LocalStore) findTraceStatus(traceID string) (CentralTraceState, *Cent
 				return DecisionKeep, status
 			}
 			// we don't have more information, so we return what we have
-			status := NewCentralTraceStatus(traceID, DecisionKeep, lrs.Clock.Now())
+			// status := NewCentralTraceStatus(traceID, DecisionKeep, lrs.Clock.Now())
```
akvanhar, line 130: Did you mean to save this for later?

kentquirk: Nope! Thanks.

centralstore/local_store.go
```diff
@@ -391,8 +408,6 @@ func (lrs *LocalStore) KeepTraces(ctx context.Context, statuses []*CentralTraceS
 			// use the one we have in the statuses list
 			lrs.states[DecisionKeep][status.TraceID] = status
 			delete(lrs.states[AwaitingDecision], status.TraceID)
-			// also remove it from the current traces list
```
VinozzZ, line 394: Why do we want to keep the span list after a decision has been made?

internal/health/health.go
```diff
@@ -162,8 +168,9 @@ func (h *Health) IsAlive() bool {
 func (h *Health) checkAlive() bool {
 	// if any counter is 0, we're dead
 	for source, a := range h.timeLeft {
-		if a == 0 {
+		if a == 0 && h.alives[source] {
 			h.Logger.Error().WithField("source", source).Logf("IsAlive: source dead due to timeout")
+			h.alives[source] = false
```
VinozzZ, line 173: Should we change the mutex on line 161 to a write lock since we are changing `h.alives` here?

centralstore/local_store.go
kentquirk: I moved this to the cleanup stage because we need it for late spans.

internal/health/health.go
```diff
@@ -146,24 +148,29 @@ func (h *Health) Ready(source string, ready bool) {
...
 }
 
 // checkAlive returns true if all registered services are alive
-// only call with the lock held
+// only call with a write lock held
 func (h *Health) checkAlive() bool {
 	// if any counter is 0, we're dead
 	for source, a := range h.timeLeft {
-		if a == 0 {
+		if a == 0 && h.alives[source] {
```
TylerHelmuth, line 171: It looks like we will only return false the first time we transition from alive to not alive.  If we are non alive twice in a row we wont return false.

VinozzZ: Do you mean if `checkAlive` is called twice and `Ready` is not called at all?

TylerHelmuth: If `checkAlive` is called twice without `h.alives[source]` being set back to `true`.

kentquirk: Thanks. Fixed.

TylerHelmuth (approved):

VinozzZ (approved):

PR #1139, opened 2024-05-13
feat: Allow sample rate to be configured for Refinery's sampling

kentquirk:

- Refinery now has tracing, but it's a multiplier; we need to be able to specify head sampling to keep it under control.

- Set up sampling in config, use it in setting up OTelTracing

### DIFF
--- a/internal/otelutil/otel_tracing.go
+++ b/internal/otelutil/otel_tracing.go
@@ -10,6 +10,7 @@ import (
 	"github.com/honeycombio/refinery/config"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
+	samplers "go.opentelemetry.io/otel/sdk/trace"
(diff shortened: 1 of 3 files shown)


internal/otelutil/otel_tracing.go
```diff
@@ -81,12 +82,20 @@ func SetupTracing(cfg config.OTelTracingConfig, resourceLibrary string, resource
 		cfg.APIHost = strings.TrimSuffix(cfg.APIHost, "/")
 		apihost := fmt.Sprintf("%s:443", cfg.APIHost)
 
+		sampleRate := cfg.SampleRate
+		if sampleRate == 0 {
```
TylerHelmuth, line 86:
```suggestion
		if sampleRate < 1 {
```

TylerHelmuth (approved):

PR #1140, opened 2024-05-13
feat: Add new ticker for memory check cycle

kentquirk:

- The memory checker is running runtime.GC() very frequently (many times per second). We think this may be causing problems.

- Create a new configuration to split off the memory checker's cycle time to a new config value
- Set it to 10s by default
- Document the setting
- Make sure it's nonzero

### DIFF
--- a/collect/central_collector.go
+++ b/collect/central_collector.go
@@ -363,6 +363,9 @@ func (c *CentralCollector) AddSpan(span *types.Span) error {
 
 func (c *CentralCollector) receive() error {
 	tickerDuration := time.Duration(c.Config.GetCollectionConfig().MemoryCycleDuration)
+	if tickerDuration <= 0 {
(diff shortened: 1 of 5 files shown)


collect/central_collector.go
```diff
@@ -362,9 +362,12 @@ func (c *CentralCollector) AddSpan(span *types.Span) error {
 }
 
 func (c *CentralCollector) receive() error {
-	tickerDuration := c.Config.GetSendTickerValue()
```
TylerHelmuth, line 365: Do we need to update anything about SendTicker's config docs?

kentquirk: I think we were overusing it conveniently for other things. We do need to update its docs for Refinery 3, but that's already planned work.

TylerHelmuth (approved):
This has an impact on stress level calculations that we'll revisit after some testing

PR #1141, opened 2024-05-13
feat: expire keys in redis based on TraceTimeout

VinozzZ:

Redis memory usage was growing faster than the rate of cleanup from refinery.

- Set the expiration time for a key to be minimum 15 minutes or 10 times of user defined TraceTimeout

### DIFF
--- a/centralstore/redis_store.go
+++ b/centralstore/redis_store.go
@@ -21,10 +21,9 @@ import (
 )
 
 const (
-	expirationForTraceStatus   = 24 * time.Hour
-	expirationForTraceState    = 24 * time.Hour
-	defaultPendingWorkCapacity = 10000
-	redigoTimestamp            = "2006-01-02 15:04:05.999999 -0700 MST"
+	defaultTraceRetention  = 15 * time.Minute
+	defaultReaperBatchSize = 400
+	redigoTimestamp        = "2006-01-02 15:04:05.999999 -0700 MST"
 
 	metricsKey          = "refinery:metrics"
 	traceStatusCountKey = "refinery:trace_status_count"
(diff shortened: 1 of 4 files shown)


centralstore/redis_store.go
```diff
@@ -1091,32 +1096,6 @@ func (t *traceStateProcessor) traceIDsByState(ctx context.Context, conn redis.Co
 	return results, nil
 }
 
-func (t *traceStateProcessor) traceIDsWithTimestamp(ctx context.Context, conn redis.Conn, state CentralTraceState, traceIDs []string) (map[string]time.Time, error) {
```
TylerHelmuth, line 1094: How come we don't need this anymore, was it not being called?

config/metadata/configMeta.yaml
```diff
@@ -1949,19 +1949,6 @@ groups:
           happen is if a Refinery crashes or is forcibly shut down. TODO: not
           sure we need to expose this.
 
-      - name: MaxTraceRetention
```
TylerHelmuth, line 1952: Lets update TraceTime's configuration to state that it is also the value used for Redis retention

centralstore/redis_store.go
VinozzZ: it was a left over code from a previous change that i forgot to remove

kentquirk (approved):
I think we have the following scenario:

* TraceTimeout 30 sec, expiration 15 minutes
* A trace is processed normally and a decision is made
* A scale up occurs and a new refinery enters the cluster
* A late span arrives 30 minutes later and is routed to the new refinery
* The new refinery hasn't seen this trace before, so it adds it to Redis
* After a trace timeout, this trace ID will be given to a refinery to make a decision.
  * If it's an old refinery, the decision will be found in the local cache and the same decision will be made
  * But if it's the new refinery, a new decision will be made which might be different from the old one

Am I correct?

I think this is OK (it's a lot of low-probability occurrences stacked up), but maybe we need to explain it at some point.

VinozzZ:
> I think we have the following scenario:
> 
> * TraceTimeout 30 sec, expiration 15 minutes
> * A trace is processed normally and a decision is made
> * A scale up occurs and a new refinery enters the cluster
> * A late span arrives 30 minutes later and is routed to the new refinery
> * The new refinery hasn't seen this trace before, so it adds it to Redis
> * After a trace timeout, this trace ID will be given to a refinery to make a decision.
>   
>   * If it's an old refinery, the decision will be found in the local cache and the same decision will be made
>   * But if it's the new refinery, a new decision will be made which might be different from the old one
> 
> Am I correct?
> 
> I think this is OK (it's a lot of low-probability occurrences stacked up), but maybe we need to explain it at some point.

You are correct. One clarification I want to add is the retention period is reset after each state transition. That means the 15min expiration starts after a decision has been made. The late arriving span will encounter the scenario you described above if it showed up 15min after a decision has been made.

Issue #1144, opened 2024-05-14
Panic! presumably in /route/route.go "fatal error: concurrent map read and map write"


senojj:
**Versions**

- Go: 1.22.3
- Refinery: 2.5.0

**Steps to reproduce**

1. This is caused by a race condition, but is most likely to occur when the application is starting under heavy load.
2. Observe the application crash when a fatal error occurs due to a concurrent map read and write.

**Additional context**

I would provide the trace of the panic, however, it was stored as multiple individual logs lines within CloudWatch, and there were many active routines. I just wanted to log this so that it has some visibility.

TylerHelmuth:
This issue has been merged to both main (for Refinery 3) and release in Refinery 2.5.1

PR #1145, opened 2024-05-14
fix: Update cache lookup to use read lock

senojj:

- #1144 

- Within the `get` method of an `environmentCache`, a read lock is added around read access to the `items` map within the environment cache. It is possible for one routine to hold a write lock within this function while another routine enters the function and attempts to read from the map while the other routine is writing to it.

### DIFF
--- a/route/route.go
+++ b/route/route.go
@@ -769,11 +769,17 @@ type cacheItem struct {
 // get queries the cached items, returning cache hits that have not expired.
 // Cache missed use the configured getFn to populate the cache.
 func (c *environmentCache) get(key string) (string, error) {
+	// get read lock so that we don't attempt to read from the map
+	// while another routine has a write lock and is actively writing
+	// to the map.
+	c.mutex.RLock()
 	if item, ok := c.items[key]; ok {
 		if time.Now().Before(item.expiresAt) {
+			c.mutex.RUnlock()
 			return item.value, nil
 		}
 	}
+	c.mutex.RUnlock()
 
 	// get write lock early so we don't execute getFn in parallel so the
 	// the result will be cached before the next lock is acquired to prevent
--- a/route/route.go
+++ b/route/route.go
@@ -769,17 +769,20 @@ type cacheItem struct {
 // get queries the cached items, returning cache hits that have not expired.
 // Cache missed use the configured getFn to populate the cache.
 func (c *environmentCache) get(key string) (string, error) {
+	var val string
 	// get read lock so that we don't attempt to read from the map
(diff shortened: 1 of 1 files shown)


kentquirk (changes requested):
Thank you for catching this! Please change the two calls to RUnlock to a single call to
`defer c.mutex.RLock()` just after line 775. That's the pattern we prefer whenever possible.

senojj:
> Thank you for catching this! Please change the two calls to RUnlock to a single call to `defer c.mutex.RLock()` just after line 775. That's the pattern we prefer whenever possible.

I may have misunderstood the ask, but I believe that would create a deadlock when the code attempts to acquire the write lock on line 787.

route/route.go
```diff
@@ -769,11 +769,17 @@ type cacheItem struct {
...
+	// while another routine has a write lock and is actively writing
+	// to the map.
+	c.mutex.RLock()
 	if item, ok := c.items[key]; ok {
 		if time.Now().Before(item.expiresAt) {
+			c.mutex.RUnlock()
 			return item.value, nil
 		}
 	}
+	c.mutex.RUnlock()
```
MikeGoldsmith, line 782: I think @senojj is correct regarding defer'ing `RUnlock` because it's in the same func and it wouldn't activate before `Lock` is called.

An alternative might look like this:

```suggestion
	val := ""
	// get read lock so that we don't attempt to read from the map
	// while another routine has a write lock and is actively writing
	// to the map.
	c.mutex.RLock()
	if item, ok := c.items[key]; ok {
		if time.Now().Before(item.expiresAt) {
			val = item.value
		}
	}
	c.mutex.RUnlock()
	if val != "" {
		return val, nil
	}
```

The above suggestion wouldn't allow empty (`""`) cache items, but not sure that's valid anyway.

senojj: I like that suggestion, and would additionally suggest that you use the zero value of the string type, as opposed to explicitly initializing it to `""`.
```suggestion
	var val string
	// get read lock so that we don't attempt to read from the map
	// while another routine has a write lock and is actively writing
	// to the map.
	c.mutex.RLock()
	if item, ok := c.items[key]; ok {
		if time.Now().Before(item.expiresAt) {
			val = item.value
		}
	}
	c.mutex.RUnlock()
	if val != "" {
		return val, nil
	}
```

MikeGoldsmith: Yeah, I think that's a good change 👍🏻

MikeGoldsmith (approved):

PR #1148, opened 2024-05-15
fix: make sure information for state machine is present before state changes

VinozzZ:

On Refinery startup, state machine information is loaded into Redis. However, redis can lose that information during a restart while Refinery continues to run. In this scenario, Refinery is unable to process any incoming traces due to the state machine being broken.

- Before each state transition request, we now check if the valid-state-change-events table exist. If it doesn't exist, we will repopulate it before process a state transition request

### DIFF
--- a/centralstore/redis_store.go
+++ b/centralstore/redis_store.go
@@ -1244,8 +1244,7 @@ func (t *traceStateProcessor) applyStateChange(ctx context.Context, conn redis.C
 }
 
 const tracesNeedingDecisionScriptKey = 1
-const tracesNeedingDecisionScript = `
-		local possibleStateChangeEvents = KEYS[1]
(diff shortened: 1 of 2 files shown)


centralstore/redis_store.go
```diff
@@ -1378,6 +1377,24 @@ func ensureValidStateChangeEvents(client redis.Client) error {
 	)
 }
 
+// validStateChangeEventsScript is a lua script that ensures the valid state change events are present in redis
+// If the valid state change events set doesn't exist, it will be created and populated with the valid state change events
+// NOTE: the events value should match the string value from stateChangeEvent struct
```
TylerHelmuth, line 1382: Is there any way to write a test that would enforce this?

TylerHelmuth (approved):

TylerHelmuth (approved):

PR #1149, opened 2024-05-16
fix: decrease network traffic from redis

VinozzZ:

Turns out transmitting trace IDs back and forth between redis and refinery is very expensive.

- Use `ZCOUNT` to get the amount of traces in a given state list instead of `ZRANGE` so that we are only returning a integer value from redis
- Use lua script for deleting expired traces in state lists to avoid returning trace IDs from redis to refinery during cleanup


kentquirk (approved):
Yes, please.

TylerHelmuth (approved):

TylerHelmuth (approved):

PR #1151, opened 2024-05-16
fix: no need to store decisions in a list in redis

VinozzZ:

A trace state is stored in two data structures in redis currently, one in a hash table with the rest of the `CentralTraceStatus` information, another one in a sorted set. The sorted set is used for `GetTracesForState` so that Refinery can quickly get all trace IDs within a state. Refinery never retrieves a list of trace IDs that has their sampling decisions already made. Therefore, we actually don't need to store than in a sorted set form anymore

This should further reduce our Redis memory usage.

- don't store trace IDs in a state if it's decision is already made

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,1 +1,1 @@
-golang 1.22
+golang 1.22.3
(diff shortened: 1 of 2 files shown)


kentquirk (approved):
Had to think about it for a while, but looks good!

PR #1153, opened 2024-05-17
fix: health check shutdown behavior improved

kentquirk:

- On shutdown, the health check could toggle aliveness when it shouldn't be because the health Ready() reports might not be timely.

- Add health.Unregister()
- Use it for receiver and decider
- Increase timeout during shutdown for sender


TylerHelmuth (approved):

internal/health/health.go
```diff
@@ -130,14 +131,34 @@ func (h *Health) Register(source string, timeout time.Duration) {
 	}
 }
 
+// Unregister a service with the health system. This marks the service as not
+// ready and removes it from the alive tracking. It also means that it no longer
+// needs to report in. If it does report in, the report will be ignored.
+func (h *Health) Unregister(source string) {
```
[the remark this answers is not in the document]
kentquirk: I don't think it matters; when stress relief hears it's shutting down, it sets Ready to false and never calls it with true again, so the right thing happens.

collect/central_collector.go
```diff
@@ -278,8 +278,10 @@ func (c *CentralCollector) shutdown(ctx context.Context) error {
 			}
 		}
 		// we have to make sure the health check says we're alive but not accepting data during shutdown
-		c.Health.Ready(receiverHealth, false)
-		c.Health.Ready(deciderHealth, false)
+		c.Health.Unregister(receiverHealth)
```
VinozzZ, line 281: Now think about it, this probably should be in `Stop()` instead of this goroutine?

collect/central_collector.go
```diff
@@ -239,6 +239,13 @@ func (c *CentralCollector) Stop() error {
 	ctx, cancel := context.WithTimeout(ctx, c.Config.GetCollectionConfig().GetShutdownDelay())
 	defer cancel()
 
+	// we have to make sure the health check says we're alive but not accepting data during shutdown
+	c.Health.Unregister(receiverHealth)
+	c.Health.Unregister(deciderHealth)
+	// reregister the sender health check to a much longer time so we can finish sending traces
+	c.Health.Register(senderHealth, 5*time.Second)
+	c.Health.Ready(senderHealth, true)
```
VinozzZ, line 247: I'm sorry I probably wasn't clear. I think line 247 still needs to be in the `sendCycle.Run` so that the sender keeps checking in with the health check system

kentquirk: You're absolutely right -- I pulled out a "shutdownMonitor" function into a local variable -- I think it's clearer because passing it inline to Run made the logic confusing to me.

PR #1157, opened 2024-05-18
maint: Update to Go 1.22

cartermp:


kentquirk (commented):
We generally leave off the patch number for a given Go release, which means it will build with latests on that release. Could you remove the `.3` from these, please?

cartermp:
@kentquirk done

kentquirk (approved):

PR #1158, opened 2024-05-19
fix: more efficient sender with gossiped decisions

kentquirk:

- The current sender is inefficient; it keeps sending a large list of trace IDs to Redis to ask for decisions; most of them aren't current. The result is a lot of traffic with Redis, most of which is useless.

- Create two new gossip channels (keep and send)
- When a decision is made, gossip the decision
- The gossip listeners accumulate batches of keep or drop decisions
    - drop is dropped immediately with no other traffic
    - keep gets statuses and acts the same way the existing sender does
- The accumulator uses an aggregation algorithm that collects messages for a max of time or number of items, then hands them off to the sender for dispatch in a goroutine.
- If the publish outraced the propagation of the trace status, it tries again next time.
- Renamed GetOldest to GetHighImpact to clarify what it does.
- Implement SpanCache.GetOldest (with test) that returns the items in the cache that have been hanging around longer than 2*(TraceTimeout+SendDelay). These traces that should have been decided by now but no gossip message was received.
- There's now a cleanup task that runs once every TraceTimeout and explicitly asks only about "Oldest" traces. This might happen after a refinery gets added to a cluster, or when the decider loop is very busy. It also acts as a hedge against issues in pub/sub, which doesn't have very robust delivery guarantees.
- Add a (temporary) config flag to control the use of this feature.

The hope is that this ends up with much less redis traffic, and less memory usage overall since decided traces are dispatched more quickly (soon after decision was made). It should also reduce latency for kept traces.

Since this task required modifying the SpanCache, it also deletes the complex span cache, which was an experiment that failed. It is both slower and more complicated than the basic span cache.

It passes existing tests because we haven't deleted the sender stuff yet, but it needs tests in CentralCollector that depend on it. 

In local testing, it appears to be noticeably faster and more consistent than the sender implementation.

The hope is to test this, and if it does what's desired, to delete the existing sender implementation and its config items. This will also have an effect on the health check and stress relief.


internal/gossip/gossip.go
```diff
@@ -43,6 +43,7 @@ func (g *InMemoryGossip) Publish(channel string, value []byte) error {
 		return errors.New("gossip has been stopped")
 	case g.channel <- msg.ToBytes():
 	default:
+		// drop the message if the channel is full
```
TylerHelmuth, line 46: Logging when we do this would be helpful

collect/central_collector.go
```diff
@@ -662,8 +870,10 @@ func (c *CentralCollector) makeDecisions(ctx context.Context) error {
 		if shouldSend {
 			state = centralstore.DecisionKeep
 			status.KeepReason = reason
+			c.Gossip.Publish(gossip_keep, []byte(trace.TraceID))
```
TylerHelmuth, line 873: Since we still set the status below, is there a possible race condition between the status in redis and the new `keepTraces`?

kentquirk: Yes, which is why KeepTraces returns the list of things that didn't have a status available from Redis yet, so that it can be retried quickly.

Without it, it still works, but it has to wait for the cleanup task to find it, which might take several seconds. 

Noticing that delay is, in fact, why I added the retry logic.

internal/gossip/gossip.go
kentquirk: I will, but prefer to wait until #1159 is merged before I touch gossip code. This was me just wanting to note that this could drop data.

collect/central_collector.go
```diff
@@ -478,12 +515,171 @@ func (c *CentralCollector) sendTraces(ctx context.Context) error {
...
+		c.cleanupTraces(ctx)
+		c.Metrics.Increment("collector_cleanup_runs")
+		return nil
+	})
+}
+
+// Cleanup traces asks for old (expired) trace IDs from the cache and then gets the status
+// (either keep or drop) from the central store, and then dispatches them appropriately.
+// In a stable refinery cluster, this should be a no-op.
+func (c *CentralCollector) cleanupTraces(ctx context.Context) {
```
TylerHelmuth, line 641: Does this cleanup process have an effect on the spans processed metrics? Should we be incrementing that metric since we are "reprocessing" the span?

TylerHelmuth (dismissed):

collect/central_collector.go
```diff
@@ -478,12 +528,163 @@ func (c *CentralCollector) sendTraces(ctx context.Context) error {
...
+		}
+	}
+}
+
+// keepTraces needs to retrieve the status of the traces from the central store
+// so that it can attach the metadata as needed
+func (c *CentralCollector) keepTraces(ids []string) []string {
+	ctx, span := otelutil.StartSpanWith(context.Background(), c.Tracer, "CentralCollector.keepTraces", "num_ids", len(ids))
+	defer span.End()
+	statuses, err := c.Store.GetStatusForTraces(ctx, ids, centralstore.DecisionKeep)
```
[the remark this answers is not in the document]
kentquirk: See if you think this is a good solution?

TylerHelmuth (approved):

PR #1159, opened 2024-05-20
fix: gossip into channels instead of callbacks

kentquirk:

- Gossip uses callbacks. A callback approach adds extra code we don't need if we use channels. This is a refactoring to change it to channels. Each commit is standalone.

- [x] Split the gossip interface file from the inmem implementation
- [x] Add a SubscribeChan function that lives beside the Subscribe function
- [x] Add test code for chan version
- [x] Convert stress relief to use the channel model
- [x] Remove callback code
- [x] Rename remaining functions to simpler names again

My desire is to merge this before #1158 and then fix that one (which does a bit more gossiping) to use channels.

### DIFF
--- a/internal/gossip/gossip_inmem.go
+++ b/internal/gossip/gossip_inmem.go
@@ -32,7 +32,7 @@ func (g *InMemoryGossip) Publish(channel string, value []byte) error {
 		return errors.New("gossip has been stopped")
 	case g.gossipCh <- msg.ToBytes():
 	default:
-		g.Logger.Debug().WithFields(map[string]interface{}{
+		g.Logger.Warn().WithFields(map[string]interface{}{
 			"channel": channel,
 			"msg":     string(value),
 		}).Logf("Unable to publish message")
@@ -73,7 +73,7 @@ func (g *InMemoryGossip) Start() error {
 					select {
 					case ch <- msg.data:
 					default:
-						g.Logger.Debug().WithFields(map[string]interface{}{
+						g.Logger.Warn().WithFields(map[string]interface{}{
 							"channel": msg.key,
 							"msg":     string(msg.data),
 						}).Logf("Unable to forward message")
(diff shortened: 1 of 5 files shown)


internal/gossip/gossip_inmem.go
```diff
@@ -0,0 +1,84 @@
...
+			select {
+			case <-g.done:
+				return nil
+			case value := <-g.gossipCh:
+				msg := newMessageFromBytes(value)
+				g.mut.RLock()
+				for _, ch := range g.subscriptions[msg.key] {
+					select {
+					case ch <- msg.data:
+					default:
```
TylerHelmuth, line 69: We should log if a message is dropped

internal/gossip/gossip_redis.go
```diff
@@ -46,10 +47,15 @@ func (g *GossipRedis) Start() error {
...
+					chans := g.subscriptions[msg.key]
 					g.lock.RUnlock()
-					for _, cb := range callbacks {
-						cb(msg.data)
+					// we never block on sending to a subscriber; if it's full, we drop the message
+					for _, ch := range chans {
+						select {
+						case ch <- msg.data:
+						default:
+							fmt.Println("dropped message", msg.key, string(msg.data))
```
TylerHelmuth, line 57: We have a logger, can we use it here?

kentquirk: Heh, this was just debugging code I forgot to delete; but I'll make it a log.

TylerHelmuth: I think it is a useful log since it'll show when our value of 100 isn't sufficient

internal/gossip/gossip_redis.go
```diff
@@ -54,7 +53,10 @@ func (g *GossipRedis) Start() error {
 						select {
 						case ch <- msg.data:
 						default:
-							fmt.Println("dropped message", msg.key, string(msg.data))
+							g.Logger.Debug().WithFields(map[string]interface{}{
```
TylerHelmuth, line 56: I feel these should be error logs.  Dropping a message here would be detrimental to Refinery's operations.

kentquirk: Because of cleanup, it will add latency but you won't lose data. I do agree that debug is not right -- but what do you think of Warn instead of Error?

TylerHelmuth (approved):

PR #1160, opened 2024-05-21
feat: add health check for GossipRedis

VinozzZ:

If the gossip channel isn't working, refinery is not able to function correctly. This PR adds the GossipRedis to the health check system

- adds a callback function to the pubsub goroutine so that we can execute a function whenever refinery receives a redis health check message
- remove the redisstore_count_decision_keep and redisstore_count_decision_drop metrics as they are no longer valid


centralstore/redis_store.go
```diff
@@ -96,6 +96,9 @@ func (r *RedisBasicStore) Start() error {
 
 	// register metrics for each state
 	for _, state := range r.states.states {
+		if state == DecisionDrop || state == DecisionKeep {
```
TylerHelmuth, line 99: isn't this from https://github.com/honeycombio/refinery/pull/1151?

VinozzZ: Yes, and this is the metrics registering code which I missed in that PR

kentquirk (approved):

PR #1169, opened 2024-05-23
fix: get trace decision from cache before asking redis

VinozzZ:

During stress relief, more trace decisions are made through the deterministic sampler and their decisions are stored only in memory. It's wasteful to asking redis for those decisions that redis has no idea of. Refinery should check with the in-memory decision cache first before asking redis for a trace status

- move the logic to check with the decision cache to the top of the `GetStatusForTraces` call
- make `RecordTraceDecision` signature to be more clear

### DIFF
--- a/centralstore/redis_store.go
+++ b/centralstore/redis_store.go
@@ -331,52 +331,54 @@ func (r *RedisBasicStore) GetStatusForTraces(ctx context.Context, traceIDs []str
 	conn := r.RedisClient.Get()
 	defer conn.Close()
 
-	statusMapFromRedis, err := r.traces.getTraceStatuses(ctx, r.RedisClient, traceIDs)
-	if err != nil {
-		return nil, err
-	}
-
 	validStates := make(map[CentralTraceState]struct{}, len(statesToCheck))
 	var decisionMade bool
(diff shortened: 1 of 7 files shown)


centralstore/redis_store.go
```diff
@@ -567,10 +602,22 @@ func (r *RedisBasicStore) KeepTraces(ctx context.Context, statuses []*CentralTra
...
+	var keep bool
+	switch trace.State {
+	case DecisionKeep:
+		keep = true
+	case DecisionDrop:
+		keep = false
+	default:
+		return fmt.Errorf("invalid trace state: %s", trace.State)
+	}
+	reason := trace.KeepReason + decisionKeepInMemoryOnly
```
TylerHelmuth, line 619: Do we have the ability to record a new flag on the status/trace? The temporary suffix feels too magical to me.

VinozzZ: Yes, I completely agree that it's a bit too magical. However, only redis_store cares about whether a decision is only stored in memory. I would prefer not have to change the cuckoo cache public interface because of this. We have plans to eventually move this cache into redis entirely. Once we do that, we won't need this magic anymore

PR #1170, opened 2024-05-23
maint: remove redis peer related code

VinozzZ:

 We no longer use redis table as a way to communicate peer membership

remove unused code


TylerHelmuth (approved):
🫗

PR #1171, opened 2024-05-23
feat: add peer service for sharing peer information

VinozzZ:

To share peer information in a centralized place so that we can take advantage of existing peer messages to manage active peers in a centralized place

- create a new PeerStore that's responsible for gossiping peer information
- inject the PeerStore into stress relief and the sampler factory


internal/peer/store.go
```diff
@@ -0,0 +1,215 @@
...
+
+// publish sends peer info from this host to the rest of the cluster.
+// If there is no new peer info to send, it will send a heartbeat.
+func (p *PeerStore) publish() {
+	defer p.wg.Done()
+
+	tk := p.Clock.NewTicker(keepAliveInterval)
+	defer tk.Stop()
+
+	for {
```
kentquirk, line 145: This select has a default, which means this is calling Publish as fast as it possibly can. I don't think that's what we want, is it?

internal/peer/peers_service.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"bytes"
+	"fmt"
+
+	"github.com/facebookgo/startstop"
+)
+
+// Peers provides information about all peers in a Refinery cluster.
+type Peers interface {
+	Subscribe(func(PeerInfo))
```
kentquirk, line 12: Why callbacks instead of channels?

Also, please add documentation in the interfaces used by external clients. That's what shows up in hovers in the editor, so named parameters and comments are better put here instead of (in addition to) the implementation.

VinozzZ: I added the comments.
I personally find the callback pattern more readable but I can go either way. It feels like it is less change to make to keep the channel pattern so I made the change

kentquirk (approved):
A few notes, but this is great!!

PR #1172, opened 2024-05-24
fix: Gossip channels are now bytes, no dropped messages

kentquirk:

- Gossip channels are arbitrary strings and don't really need to be; we have only a few of them. Instead, you can use your string name to create a numeric channel and put that on the messages. Because we need different instances to use the same channel IDs, we still use constants for the string IDs. We can build IDs on the fly (and do when we're standalone) but in Redis we expect the constants to be used. 

- Define Channel type
- Adjust how messages are built and used
- Fix the places they were used
- Use constants
- Debug tests

### DIFF
--- a/internal/gossip/gossip_inmem.go
+++ b/internal/gossip/gossip_inmem.go
@@ -12,7 +12,8 @@ import (
 type InMemoryGossip struct {
 	Logger        logger.Logger `inject:""`
 	gossipCh      chan []byte
-	subscriptions map[string][]chan []byte
+	subscriptions map[Channel][]chan []byte
+	channels      map[string]Channel
 
 	done chan struct{}
 	mut  sync.RWMutex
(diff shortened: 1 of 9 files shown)


internal/gossip/gossip_inmem.go
```diff
@@ -67,19 +60,19 @@ func (g *InMemoryGossip) Start() error {
...
-					select {
-					case ch <- msg.data:
-					default:
-						g.Logger.Warn().WithFields(map[string]interface{}{
-							"channel": msg.key,
-							"msg":     string(msg.data),
-						}).Logf("Unable to forward message")
-					}
-				}
+				chans := g.subscriptions[msg.Channel()]
```
VinozzZ, line 65: Should we copy the slice here as well?

internal/gossip/gossip_redis.go
```diff
@@ -137,14 +136,15 @@ func (g *GossipRedis) GetChannel(name string) Channel {
...
 	}
-	ch := Channel('A')
-	for _, v := range g.channels {
-		if v == ch {
-			ch++
-		}
+
+	var ch Channel
+	if strings.HasPrefix(name, "test") {
+		ch = Channel(name[4]) // yes, this will panic if you don't specify a 5th character
```
VinozzZ, line 142: Clever!

VinozzZ (approved):

PR #1173, opened 2024-05-24
fix: make sure we clean up data in redis for late spans

VinozzZ:

This PR fixes multiple issues when dealing with late spans.
- updating late span counts in decision cache
- make sure late span data in redis gets cleaned up
- make sure key span list in redis is cleaned up if the trace decision is made through stress relief
-  make sure expired traces in span cache is cleaned up regardless their state

- create `SpanRecorder` interface so that both `types.Span` and `centralstore.Span` can be recorded in cuckoo cache
- add expiration time for `addStatus`, `storeSpans`, and `incrementSpanCount` queries
- make sure `cleanupTraces` retrieves state for a trace in spancache regardless which state the trace is in

-

### DIFF
--- a/centralstore/centralstore.go
+++ b/centralstore/centralstore.go
@@ -24,7 +24,7 @@ type CentralSpan struct {
 	IsRoot          bool
 }
 
-// can you come up with another name for this fucntion?
+// Trace returns the trace ID of the span.
 func (s *CentralSpan) Trace() string {
 	return s.TraceID
 }
--- a/centralstore/centralstore.go
+++ b/centralstore/centralstore.go
@@ -47,6 +47,10 @@ func (s *CentralSpan) SetSamplerSelector(key string) {
 
 type CentralTraceState string
 
+// AllTraceStates is a list of all possible trace states.
+var AllTraceStates = []CentralTraceState{DecisionKeep, DecisionDrop, Unknown, Collecting,
+	DecisionDelay, ReadyToDecide, AwaitingDecision}
+
 const (
 	Unknown          CentralTraceState = "unknown"
 	Collecting       CentralTraceState = "collecting"
--- a/centralstore/centralstore.go
+++ b/centralstore/centralstore.go
@@ -24,6 +24,15 @@ type CentralSpan struct {
 	IsRoot          bool
 }
 
+// can you come up with another name for this fucntion?
+func (s *CentralSpan) Trace() string {
+	return s.TraceID
+}
+
+func (s *CentralSpan) SpanType() types.SpanType {
+	return s.Type
+}
+
 func (s *CentralSpan) Fields() map[string]interface{} {
 	if s.KeyFields == nil {
 		return s.AllFields
--- a/centralstore/redis_store.go
+++ b/centralstore/redis_store.go
@@ -623,8 +623,8 @@ func (r *RedisBasicStore) getTraceStates(ctx context.Context, conn redis.Conn, s
 	defer span.End()
 
 	var cacheHitCount int
-	notFound := make([]string, 0)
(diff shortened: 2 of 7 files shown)


centralstore/redis_store.go
```diff
@@ -984,12 +999,14 @@ func (t *tracesStore) incrementSpanCounts(ctx context.Context, conn redis.Conn,
 	})
 	defer spanInc.End()
 
-	cmds := make([]redis.Command, len(spans))
-	for i, s := range spans {
-		cmds[i] = t.incrementSpanCountsCMD(s.TraceID, s.Type)
+	commands := make([]redis.Command, 2*len(spans))
+	for i := 0; i < len(commands); i += 2 {
+		span := spans[i/2]
```
kentquirk, line 1004: I don't think i/2 is right -- don't you want i+1?

collect/central_collector.go
```diff
@@ -668,6 +668,17 @@ func (c *CentralCollector) cleanupTraces(ctx context.Context) {
...
 	}
+	for _, t := range ids {
+		trace := c.SpanCache.Get(t)
+		if trace == nil {
+			continue
+		}
+		data := make(map[string]interface{}, 0)
+		data["trace_id"] = t
+		data["arrival_time"] = trace.ArrivalTime
+		data["span_coutn"] = len(trace.GetSpans())
```
kentquirk, line 679: spelling of `span_count`

And should these be `meta` fields?

collect/central_collector_test.go
```diff
@@ -186,7 +186,7 @@ func TestCentralCollector_ProcessTraces(t *testing.T) {
 				assert.Equal(t, numberOfTraces*10, len(transmission.Events))
 				assert.Equal(t, "aoeu", transmission.Events[0].Dataset)
 				assert.Equal(t, "test", transmission.Events[0].Environment)
-				assert.Equal(t, TraceSendGotRoot, transmission.Events[0].Data["meta.refinery.send_reason"])
+				assert.Equal(t, TraceSendGotRoot, transmission.Events[0].Data["meta.refinery.send_reason"], transmission.Events[0])
```
kentquirk, line 189: maybe this wants to be a format string with a message instead of  just an extra parameter?

centralstore/centralstore.go
```diff
@@ -24,6 +24,15 @@ type CentralSpan struct {
 	IsRoot          bool
 }
 
+// can you come up with another name for this fucntion?
+func (s *CentralSpan) Trace() string {
```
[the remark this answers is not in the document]
VinozzZ: Unfortunately, `CentralSpan` already has an exported field named `TraceID`

centralstore/redis_store.go
VinozzZ: This is so that each span in the spans list is looped through. For each span, there are two commands to add to the commands list. I believe `i/2` will allow me to loop through each span in the spans list

collect/central_collector.go
VinozzZ: Ooops, this is my debugging code lol

collect/central_collector_test.go
VinozzZ: I didn't mean to commit this

centralstore/redis_store.go
kentquirk: Ah, now I see it. Thanks.

centralstore/centralstore.go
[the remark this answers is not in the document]
kentquirk: Maybe GetTraceID? Or leave it as Trace, it's fine. In either case, please remove the comment.

kentquirk (approved):
Just clean up that one thing and it's good.

PR #1175, opened 2024-05-29
fix: return when gossip channel is closed

VinozzZ:

Test are flaky when using in-memory gossiper due to race condition on shutdown.  

- If the `g.gossipCh` is closed, Refinery should return instead of keep processing the value


kentquirk (approved):
Ohhhhh, now I see how message could be empty - it's a slice of bytes so it could be nil!

PR #1176, opened 2024-05-29
fix: allow sending otel tracing to honeycomb classic environment

VinozzZ:

Refinery's traces wasn't getting send to honeycomb classic environment

- add dataset header to otel tracing config

### DIFF
--- a/internal/otelutil/otel_tracing.go
+++ b/internal/otelutil/otel_tracing.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"regexp"
 	"strings"
 
 	"github.com/honeycombio/otel-config-go/otelconfig"
@@ -14,6 +15,9 @@ import (
 	"go.opentelemetry.io/otel/trace"
 	"go.opentelemetry.io/otel/trace/noop"
 )
+
+var classicKeyRegex = regexp.MustCompile(`^[a-f0-9]*$`)
+var classicIngestKeyRegex = regexp.MustCompile(`^hc[a-z]ic_[a-z0-9]*$`)
 
 // telemetry helpers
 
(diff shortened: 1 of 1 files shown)


internal/otelutil/otel_tracing.go
```diff
@@ -97,7 +97,8 @@ func SetupTracing(cfg config.OTelTracingConfig, resourceLibrary string, resource
 			otelconfig.WithTracesEnabled(true),
 			otelconfig.WithSampler(samplers.TraceIDRatioBased(sampleRatio)),
 			otelconfig.WithHeaders(map[string]string{
-				"x-honeycomb-team": cfg.APIKey,
+				"x-honeycomb-team":    cfg.APIKey,
+				"x-honeycomb-dataset": cfg.Dataset,
```
MikeGoldsmith, line 101: We should probably check the API key length and only set this header when it's a classic key. We did this in our distros - https://github.com/honeycombio/honeycomb-opentelemetry-go/blob/f93016685160d16fd11b0831db80396038fb1b1c/honeycomb.go#L215

MikeGoldsmith (approved):

internal/otelutil/otel_tracing.go
```diff
@@ -108,3 +118,12 @@ func SetupTracing(cfg config.OTelTracingConfig, resourceLibrary string, resource
 	pr := noop.NewTracerProvider()
 	return pr.Tracer(resourceLibrary, trace.WithInstrumentationVersion(resourceVersion)), func() {}
 }
+
+func isClassicKey(key string) bool {
```
kentquirk, line 122: instead of writing this, you should use `types.IsLegacyAPIKey` please

kentquirk (approved):
Approved, but we already have a key checker; let's not write another one.

kentquirk (approved):

PR #1177, opened 2024-05-29
perf: gossip traceIDs in batches instead of one at a time

kentquirk:

- We were gossiping traceIDs one at a time. This is a lot of messages and a lot of goroutines to process them.

- Batch them up and compress them using snappy compression.
- Test and benchmark the compression code and fix existing tests


collect/central_collector.go
```diff
@@ -900,6 +936,24 @@ func (c *CentralCollector) makeDecisions(ctx context.Context) error {
 		span.End()
 	}
 
+	if len(keptIDs) > 0 {
+		data, err := compress(keptIDs)
```
VinozzZ, line 940: I wonder if we should put this into a goroutine or something so that this is not blocking decider to work on the next batch of traces. I'm mostly worried that it's going to reduce the decider's throughput

kentquirk: The snappy compressor is quite fast -- 1000 ids takes 1 msec on my desktop. I don't think this is going to be the biggest problem.

PR #1180, opened 2024-05-31
feat: Dispatch late spans asap if they’re in the decision cache

VinozzZ:

Late spans currently have to wait for the cleanup goroutine in central collector to be processed. That means Refinery is holding on to late spans for longer than we need to

- move decision cache into central collector
- enable tests for the decision gossip code
- fix a race condition in the decision gossip code


collect/central_collector.go
```diff
@@ -512,6 +530,19 @@ func (c *CentralCollector) sendTraces(ctx context.Context) error {
 		c.Metrics.Histogram("sender_considered_per_second", tracesConsidered/sendTime.Seconds())
 	}()
 
+	// if a trace is dropped, we don't want to send it to the central store
+	ids = slices.DeleteFunc(ids, func(id string) bool {
+		if c.DecisionCache.Dropped(id) {
+			c.SpanCache.Remove(id)
```
kentquirk, line 536: oh, clever. I like this.

collect/central_collector.go
```diff
@@ -217,15 +227,17 @@ func (c *CentralCollector) Start() error {
 	if maxConcurrency <= 0 {
 		maxConcurrency = 4
 	}
-	c.egAgg = &errgroup.Group{}
-	c.egAgg.SetLimit(maxConcurrency) // we want to limit the number of goroutines that are aggregating trace IDs
+	egKeepAgg := &errgroup.Group{}
+	egDropAgg := &errgroup.Group{}
+	egKeepAgg.SetLimit(maxConcurrency) // we want to limit the number of goroutines that are aggregating trace IDs
```
kentquirk, line 232: I'm not sure I think that these different concurrency limits should all be the same. I'm especially worried that the keepTraces one might want to be higher, or it's going to block our ability to process traces.

I guess we can mess with it, but it's something to watch out for.

kentquirk (approved):
I'm worried about the concurrency limit (I deliberately didn't add one before) but we should just experiment.

PR #1183, opened 2024-06-03
feat: Get central trace in one round trip to redis

VinozzZ:

GetTrace was costing us a roundtrip to redis for each traces. It's not as performant as we want it to be.

- Use pipelining to send all GetTrace request in a single batch
- change encoding method from json to snappy for trace metadata and span data


collect/central_collector.go
```diff
@@ -834,53 +835,33 @@ func (c *CentralCollector) makeDecisions(ctx context.Context) error {
...
 
 	ctxTraces, spanTraces := otelutil.StartSpanWith(ctx, c.Tracer, "CentralCollector.makeDecision.traceLoop", "num_traces", len(traces))
 	defer spanTraces.End()
 
-	gossip_keep_channel := c.Gossip.GetChannel(gossip.ChannelKeep)
-	gossip_drop_channel := c.Gossip.GetChannel(gossip.ChannelDrop)
 	keptIDs := make([]string, 0)
 	droppedIDs := make([]string, 0)
 
+	c.Logger.Error().Logf("deciding on %d traces", len(statuses))
```
TylerHelmuth: I think `Info` or `Debug` feels more appropriate than `Error`

cmd/test_stores/fake_refinery.go
```diff
@@ -132,12 +132,14 @@ func (fri *FakeRefineryInstance) runDecider(opts CmdLineOptions, nodeIndex int,
...
 
-						trace, err := fri.Store.GetTrace(getCtx, status.TraceID)
+						traces, err := fri.Store.GetTraces(getCtx, status.TraceID)
 						if err != nil {
 							otelutil.AddException(span3, err)
 						}
-						otelutil.AddSpanField(span3, "span_count", len(trace.Spans))
-						traces = append(traces, trace)
+						for _, trace := range traces {
+							otelutil.AddSpanField(span3, "span_count", len(trace.Spans))
```
kentquirk, line 140: You're trying to add `span_count` once for each trace; they're going to overwrite each other. Are you trying to record the total number of spans in the trace?

cmd/test_stores/fake_refinery.go
```diff
@@ -220,16 +222,21 @@ func (fri *FakeRefineryInstance) runDecider(opts CmdLineOptions, nodeIndex int,
 						span3.End()
 						continue
 					}
-					trace, err := fri.Store.GetTrace(getCtx, status.TraceID)
+					traces, err := fri.Store.GetTraces(getCtx, status.TraceID)
 					if err != nil {
 						otelutil.AddException(span3, err)
 					}
+					if len(traces) == 0 {
+						otelutil.AddSpanField(span3, "decision", "no trace")
```
kentquirk, line 230: Is "decision" the right field here? That seems weird, but maybe it's fine.

centralstore/redis_store.go
```diff
@@ -264,48 +264,57 @@ func (r *RedisBasicStore) WriteSpans(ctx context.Context, spans []*CentralSpan)
...
-		Span   []byte
+	for _, traceID := range traceIDs {
+		cmd := redis.NewGetAllValuesHashCommand(spansHashByTraceIDKey(traceID))
+		if err := cmd.Send(conn); err != nil {
+			return nil, err
+		}
 	}
 
-	err := conn.GetSliceOfStructsHash(spansHashByTraceIDKey(traceID), &tmpSpan)
+	data, err := conn.ReceiveByteSlice(len(traceIDs))
```
kentquirk, line 284: Given that this returns `[][][]byte`, I feel like I would like to see a lot more comments in the loops here explaining the logic here. 

Also, shouldn't this be called `ReceiveByteSlices`?

(Really, I wish there were a better way to describe that type, but you're stuck with what redigo did.)

kentquirk (approved):
I think there's an issue in the telemetry, and a couple of other notes, but no need to block it.

cmd/test_stores/fake_refinery.go
VinozzZ: I changed it to `not_found`

PR #1185, opened 2024-06-04
feat: use pipelining for get trace status from redis

VinozzZ:

Optimize redis query performance by using pipelining as suggested by redis doc https://redis.io/docs/latest/develop/use/pipelining/

- Implement pipelining for `HGETALL` command
- Implement `FanoutChunksToMap` to allow using fanout pattern for batch processing


generics/fanout.go
```diff
@@ -161,3 +161,74 @@ func EasyFanoutToMap[T comparable, U any](input []T, parallelism int, worker fun
 		return worker, nil
 	}, nil)
 }
+
+// FanoutChunksToMap takes a slice of input, a chunk size, a maximum parallelism factor, and a worker factory.
+// It calls the generated worker on every chunk of the input, and returns a (possibly filtered) map of the inputs to the outputs.
+// Only the outputs that pass the predicate (if it is not nil) will be added to the map.
+//
+// The maximum parallelism factor is the maximum number of workers that will be run in parallel. The actual number of workers
+// will be the minimum of the maximum parallelism factor and the number of chunks in the input.
```
kentquirk, line 170:
```suggestion
// FanoutChunksToMap takes a slice of input, a chunk size, a maximum parallelism
// factor, and a worker factory. It calls the generated worker on every chunk of
// the input, and returns a (possibly filtered) map of the inputs to the
// outputs. Only the outputs that pass the predicate (if it is not nil) will be
// added to the map.
//
// The maximum parallelism factor is the maximum number of workers that will be
// run in parallel. The actual number of workers will be the minimum of the
// maximum parallelism factor and the number of chunks in the input.
```

kentquirk (approved):
We went over this in a pairing session. 

I just wordwrapped a comment.

PR #1186, opened 2024-06-04
fix: Revert OTel to avoid semconv version clash

kentquirk:

- OTel updates pulled in a new semconv in some of the deps. Refinery currently depends on Honeycomb's otel-config-go, which doesn't have the latest, so we got a version conflict. Tried to update the honeycomb one, and IT, in turn, depends on lambda, which doesn't have the latest...

- Revert to otel 1.25 for now. We'll deal with the above issue by dropping our dep on Honeycomb's helper lib.


go.mod
```diff
@@ -10,6 +10,7 @@ require (
 	github.com/dgryski/go-wyhash v0.0.0-20191203203029-c4841ae36371
 	github.com/facebookgo/inject v0.0.0-20180706035515-f23751cae28b
 	github.com/facebookgo/startstop v0.0.0-20161013234910-bc158412526d
+	github.com/golang/snappy v0.0.4
```
VinozzZ, line 13: Do we currently have code for snappy?

PR #1168, opened 2024-05-23
feat: Add OTLP log endpoints (gRPC & HTTP)

MikeGoldsmith:
## Which problem is this PR solving?
Adds support for OTLP logs. Only logs with a trace ID associated are handled and are sampled along with spans.

## Short description of the changes
- Add OTLP endpoints gRPC & HTTP (http/proto & http/json) including support for gzip & zstd
- Add unit tests to verify endpoints and different encodings are handled
- Rename `processTraceRequest` to ``processOTLPRequest` and move otlp_trace to router
- Rename OTLP trace endpoint func `postOTLP` to `postOTLPTraces`

PR #1187, opened 2024-06-05
feat: Add OTLP log endpoints (gRPC & HTTP)

MikeGoldsmith:

Adds support for receving OTLP logs. Only logs with a trace ID associated are handled and are sampled along with spans. Logs without a trace ID are forwarded to Honeycomb immediately like events without TraceIDs are.

Further work to add support for configurable correlation IDs will be done in later PRs.

**Note**: This is the 2.x version of adding OTLP Logs, see the following PR that tracks adding to 3.x on main branch:
- #1168 

- Add OTLP endpoints gRPC & HTTP (http/proto & http/json) including support for gzip & zstd
- Adds missing API key check for trace gRPC handler with tests
- Add unit tests to verify endpoints and different encodings are handled
- Add unit tests to verify logs with a Trace ID are added to collector and logs without a Trace ID are sent to libhoney for dispatch
- Rename processTraceRequest to `processOTLPRequest` and use in otlp_trace
- Rename OTLP trace endpoint func postOTLP to postOTLPTraces
- Add MockCollector to aid with testing


route/route.go
```diff
@@ -915,6 +962,9 @@ func (r *Router) AddOTLPMuxxer(muxxer *mux.Router) {
 	otlpMuxxer := muxxer.PathPrefix("/v1/").Methods("POST").Subrouter()
 
 	// handle OTLP trace requests
-	otlpMuxxer.HandleFunc("/traces", r.postOTLP).Name("otlp")
-	otlpMuxxer.HandleFunc("/traces/", r.postOTLP).Name("otlp")
+	otlpMuxxer.HandleFunc("/traces", r.postOTLPTrace).Name("otlp")
```
kentquirk, line 965: We don't use `.Name` for its intended purpose -- it's basically just a comment for us -- but I think it would be worth using different values for the logs and traces muxxer names; they might show up in a log or debugging session sometime.

kentquirk (approved):
This looks really good! One very minor nit.

PR #1188, opened 2024-06-05
feat: store trace kept record with pipelining instead of lua script

VinozzZ:

We found the keep trace lua script in redis' slow query log. Turns out, we actually don't need to store those fields separately since they don't change once they are set. By storing them together, we can eliminate the need for using lua script.

- Define a `KeepRecord` field in the data structure we use to store trace status in redis
- encode `centralTraceStatusReason` into one json blob
- change redis to allow easier pipelining commands

### DIFF
--- a/centralstore/redis_store.go
+++ b/centralstore/redis_store.go
@@ -87,7 +87,7 @@ func (r *RedisBasicStore) Start() error {
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


kentquirk (approved):
We talked about changing the interface this uses, but that can be a separate PR.

PR #1199, opened 2024-06-13
feat: Allow URL encoded dataset in libhoney endpoint paths

MikeGoldsmith:

Allow URL encoded dataset names in /event and /batch endpoint.

For example, to send data to a dataset with the name `foo/bar`, you can now send data like this:

```
curl -i -H "x-honeycomb-team: <api-key>" -X POST 'http://localhost:8080/1/events/foo%2Fbar' -d '{"name":"mike"}'
```

The dataset name `foo%2Fbar` is URL decoded into `foo/bar` and used as the dataset name. When forwarding events to the Honeycomb API, the dataset name is URL encoded again to ensure it's received into HNY correctly.

- Update mux router to allow encoded paths
- Add utility function to get the dataset from mux vars and url escape the value
- Add tests for utility function
- Simplify batch event processing by removing `batchedEventToEvent`, adding `getSampleRate` batchEvent member func and re-using apiHost and dataset

### DIFF
--- a/route/route.go
+++ b/route/route.go
@@ -450,6 +450,15 @@ func (r *Router) batch(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 
+	dataset, err := getDatasetFromRequest(req)
+	if err != nil {
+		r.handlerReturnWithError(w, ErrReqToEvent, err)
+	}
+	apiHost, err := r.Config.GetHoneycombAPI()
+	if err != nil {
+		r.handlerReturnWithError(w, ErrReqToEvent, err)
+	}
+
 	apiKey := req.Header.Get(types.APIKeyHeader)
 	if apiKey == "" {
 		apiKey = req.Header.Get(types.APIKeyHeaderShort)
@@ -463,7 +472,7 @@ func (r *Router) batch(w http.ResponseWriter, req *http.Request) {
 
 	batchedResponses := make([]*BatchResponse, 0, len(batchedEvents))
 	for _, bev := range batchedEvents {
-		ev, err := r.batchedEventToEvent(req, bev, apiKey, environment)
+		ev, err := r.batchedEventToEvent(req, bev, apiKey, environment, dataset, apiHost)
 		if err != nil {
 			batchedResponses = append(
 				batchedResponses,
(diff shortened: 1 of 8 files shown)


MikeGoldsmith:
Waiting on the following PR:
- https://github.com/honeycombio/refinery/pull/1200

route/route.go
```diff
@@ -969,3 +974,16 @@ func (r *Router) AddOTLPMuxxer(muxxer *mux.Router) {
...
+func getDatasetFromRequest(req *http.Request) (string, error) {
+	dataset := mux.Vars(req)["datasetName"]
+	if dataset == "" {
+		return "", fmt.Errorf("missing dataset name")
+	}
+	dataset, err := url.PathUnescape(dataset)
+	if err != nil {
+		return "", err
+	}
+	fmt.Println("dataset", dataset)
```
VinozzZ, line 987: Is this print line intended?

MikeGoldsmith: Nope, it can be removed. 

'''suggestion
'''

VinozzZ (approved):

route/route_test.go
```diff
@@ -612,3 +613,109 @@ func TestGRPCHealthProbeWatch(t *testing.T) {
...
+			datasetName:         "foo%5Fbar",
+			expectedDatasetName: "foo_bar",
+		},
+		{
+			name:                "dataset name with URL encoded tilde",
+			datasetName:         "foo%7Ebar",
+			expectedDatasetName: "foo~bar",
+		},
+		{
+			name:                "dataset name with URL encoded period",
```
kentquirk, line 680: just for completeness, how about a dataset name with a `%` symbol?

route/route.go
```diff
@@ -678,8 +681,10 @@ func (r *Router) batchedEventToEvent(req *http.Request, bev batchedEvent, apiKey
 	eventTime := bev.getEventTime()
 	// TODO move the following 3 lines outside of this loop; they could be done
 	// once for the entire batch instead of in every event.
-	vars := mux.Vars(req)
```
kentquirk, line 681: Is this the time where we actually implement this TODO given that we're messing with this code?

kentquirk (approved):
I don't need to block it, but given that there's an old TODO there and this code is being touched...

route/route_test.go
MikeGoldsmith: Added extra tests in https://github.com/honeycombio/refinery/pull/1199/commits/492e57250ae9b58d5e18d64c93ac24187755bf3b

route/route.go
MikeGoldsmith: Moved apihost and dataset lookup outside of the loop in https://github.com/honeycombio/refinery/pull/1199/commits/458d27f3c97ddbfa44c59771a4ea17c875bfafd0

PR #1205, opened 2024-06-15
feat: Working, tested, but unused pubsub system

kentquirk:

- In order to implement a gossip protocol, we need a functional pubsub system. I tried the go-redis library instead of redigo and it was pretty simple to use.

- Create a standardized pubsub interface
- Implement go-redis version of it
- Implement local version of it
- Write tests
- Write benchmarks
- Implement startstop
- Set up based on our redis peer config (I have not used all of the pool size parameters, since they don't  seem to matter anymore)

On my local machine (without going through an external network), it runs at about 20K messages per second, with average latency of about 200uS and max latency of about 5mS. On CI, the first test had avg latency of 6mS.

I also tried implementing a version using the reuidis library; see comments in the code. Basically, for pubsub it was no faster and the implementation was a bit ugly so I did not include it.


pubsub/pubsub_goredis.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+}
+
+// assert that GoRedisPubSub implements PubSub
+var _ PubSub = (*GoRedisPubSub)(nil)
+
+// when a topic is created, it is stored in the topics map and a goroutine is
+// started to listen for messages on the topic; each message is sent to all
+// subscribers to the topic.
+func (ps *GoRedisPubSub) NewTopic(ctx context.Context, topic string) Topic {
```
VinozzZ, line 49: I love the simplicity of the pub/sub interface from go-redis 💅🏽

pubsub/pubsub_goredis.go
```diff
@@ -0,0 +1,132 @@
...
+	ch := make(chan string)
+	t.mut.Lock()
+	t.subscribers = append(t.subscribers, ch)
+	t.mut.Unlock()
+	return ch
+}
+
+// Close shuts down the topic and unsubscribes all subscribers
+func (t *GoRedisTopic) Close() {
+	t.mut.Lock()
```
VinozzZ, line 125: I think we can use `sync.Once` for closing the `t.done` channel here to simplify the code. That way, `t.Close()` also doesn't need to race against other code paths for accessing the lock

kentquirk: Oh, great idea!

pubsub/pubsub_goredis.go
```diff
@@ -0,0 +1,129 @@
...
```
VinozzZ, line 124: I don't think we need this mutex here anymore?

pubsub/pubsub_local.go
```diff
@@ -0,0 +1,116 @@
...
+	mut    sync.RWMutex
+}
+
+type LocalTopic struct {
+	topic       string
+	pubChan     chan string
+	subscribers []chan string
+	done        chan struct{}
+	mut         sync.RWMutex
+	closed      bool
```
VinozzZ, line 23: This can use the same `sync.Once` pattern as the GoRedis implementation

pubsub/pubsub_goredis.go
```diff
@@ -83,88 +85,58 @@ func (ps *GoRedisPubSub) Stop() error {
...
 				}
-			case <-t.done:
-				t.redisSub.Close()
-				t.mut.RLock()
-				subscribers := t.subscribers
-				for _, sub := range subscribers {
-					close(sub)
+				select {
+				case sub.ch <- msg.Payload:
+				default:
```
VinozzZ, line 127: Can we log something here indicating a message has been dropped?

pubsub/pubsub_goredis.go
```diff
@@ -83,88 +85,58 @@ func (ps *GoRedisPubSub) Stop() error {
...
 }
 
-// Close shuts down the topic and unsubscribes all subscribers
-func (t *GoRedisTopic) Close() {
-	t.once.Do(func() {
-		close(t.done)
-	})
+func (s *GoRedisSubscription) Close() {
+	s.pubsub.Close()
+	close(s.done)
```
VinozzZ, line 141: We probably need a `sync.Once` here for `close(s.done)` in case a client called `Close` and another goroutine called `GoRedisPubSub.Close()`

pubsub/pubsub_test.go
```diff
@@ -33,24 +36,24 @@ func TestPubSubBasics(t *testing.T) {
...
-				ch := topic.Subscribe(ctx)
-				for msg := range ch {
-					require.Contains(t, msg, "message")
+				time.Sleep(100 * time.Millisecond)
+				for i := 0; i < 10; i++ {
+					err := ps.Publish(ctx, "topic", fmt.Sprintf("message %d", i))
+					require.NoError(t, err)
 				}
+				time.Sleep(100 * time.Millisecond)
+				ps.Close()
```
VinozzZ, line 56: I can comment on line 60. Do we need to call `ps.Close()` twice?

pubsub/pubsub_test.go
```diff
@@ -67,34 +70,34 @@ func TestPubSubMultiTopic(t *testing.T) {
...
-						err := topics[j].Publish(ctx, fmt.Sprintf("%d", (i+1)*(j+1)))
+						err := ps.Publish(ctx, topics[j], fmt.Sprintf("%d", (i+1)*(j+1)))
 						require.NoError(t, err)
 					}
 				}
 				time.Sleep(100 * time.Millisecond)
-				for i := 0; i < topicCount; i++ {
-					topics[i].Close()
-				}
+				ps.Close()
```
VinozzZ, line 90: Same as above. Do we need the second close call on line 110?

pubsub/pubsub_test.go
```diff
@@ -231,37 +225,43 @@ func BenchmarkPubSub(b *testing.B) {
...
-			topics := make([]pubsub.Topic, topicCount)
+			topics := make([]string, topicCount)
 			for i := 0; i < topicCount; i++ {
-				topics[i] = ps.NewTopic(ctx, fmt.Sprintf("topic%d", i))
+				topics[i] = fmt.Sprintf("topic%d", i)
 			}
-			time.Sleep(100 * time.Millisecond)
+			mut := sync.RWMutex{}
+			count := 0
+			counts := make([]int, topicCount)
```
VinozzZ, line 239: I can't seem to find there `counts` are being used

pubsub/pubsub_goredis.go
```diff
@@ -83,88 +85,58 @@ func (ps *GoRedisPubSub) Stop() error {
...
+}
+
+func (ps *GoRedisPubSub) Publish(ctx context.Context, topic, message string) error {
+	ps.mut.RLock()
+	defer ps.mut.RUnlock()
+	return ps.client.Publish(ctx, topic, message).Err()
+}
 
+func (ps *GoRedisPubSub) Subscribe(ctx context.Context, topic string) Subscription {
+	ps.mut.Lock()
```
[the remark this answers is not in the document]
kentquirk: We're accessing ps.client when we build the Subscription object. 

I could move the unlock to above the goroutine, but I don't think it makes a difference since the only other thing that happens is the return.

pubsub/pubsub_test.go
kentquirk: This was a vestige of previous tests, but we don't need it any more. Thanks.

pubsub/pubsub_goredis.go
kentquirk: I really didn't want to do that for a number of reasons, but also it's probably the cheapest way to detect a problem here, so I did it anyway.

pubsub/pubsub_goredis.go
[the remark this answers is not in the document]
VinozzZ: I believe the redis.Client is concurrent safe based on their doc

VinozzZ (approved):
💯

PR #1206, opened 2024-06-16
feat: Add some useful generics

kentquirk:

- Some features I plan to import from Refinery 3 will likely use the Fanout generic that was added there
- The gossip-based peer mechanism will need SetWithTTL

- Create and test SetWithTTL
- Import code and tests for Fanout


generics/fanout_test.go
```diff
@@ -0,0 +1,196 @@
...
+			mut.Lock()
+			cleanupTotal += i
+			mut.Unlock()
+		}
+		return worker, cleanup
+	}
+
+	result := Fanout(input, parallelism, workerFactory, nil)
+	assert.ElementsMatch(t, []int{2, 4, 6, 8, 10}, result)
+	assert.Equal(t, 6, cleanupTotal) // 0 + 1 + 2 + 3
```
VinozzZ, line 64: Would it be easier to reason about to do `cleanupTotal += 1` and assert that `assert.Eqaul(t, parallelism, cleanupTotal)`?

generics/fanout_test.go
```diff
@@ -0,0 +1,196 @@
...
+			mut.Lock()
+			cleanupTotal += i
+			mut.Unlock()
+		}
+		return worker, cleanup
+	}
+
+	result := FanoutToMap(input, parallelism, workerFactory, nil)
+	assert.EqualValues(t, expected, result)
+	assert.Equal(t, 6, cleanupTotal) // 0 + 1 + 2 + 3
```
VinozzZ, line 125: Same as above

generics/setttl.go
```diff
@@ -0,0 +1,75 @@
...
+		TTL:   ttl,
+	}
+	s.Add(es...)
+	return s
+}
+
+// Add adds elements `es` to the SetWithTTL.
+func (s SetWithTTL[T]) Add(es ...T) {
+	for _, e := range es {
+		s.Items[e] = time.Now().Add(s.TTL)
```
VinozzZ, line 31: If we still want to use `clockwork` package as a unified way to management time in Refinery, we will need to inject that into this generic implementation of SetWithTTL

generics/setttl.go
```diff
@@ -0,0 +1,75 @@
...
+	"cmp"
+	"sort"
+	"time"
+
+	"golang.org/x/exp/maps"
+)
+
+// SetWithTTL is a unique set of items with a TTL (time to live) for each item.
+// After the TTL expires, the item is automatically removed from the set.
+type SetWithTTL[T cmp.Ordered] struct {
```
VinozzZ, line 13: Should we make this implementation concurrent safe? If not, maybe we can add a note about it in the comments

generics/setttl.go
```diff
@@ -0,0 +1,75 @@
...
+	maps.DeleteFunc(s.Items, func(k T, exp time.Time) bool {
+		return exp.Before(time.Now())
+	})
+}
+
+// Members returns the unique elements of the SetWithTTL in sorted order.
+// It also removes any items that have expired.
+func (s SetWithTTL[T]) Members() []T {
+	s.cleanup()
+	members := make([]T, 0, len(s.Items))
```
VinozzZ, line 61: nit - we can use `maps.Keys` here

generics/setttl.go
kentquirk: We can't really do injection here; this is not a Start-able component. I also didn't want to make it dependent on passing a Clock in whenever you need one. 

What do you think of giving it a Clock member that defaults to a Real clock, and you have to manually override it in tests?

VinozzZ: Yes, that's actually what I was thinking too

VinozzZ (approved):

PR #1208, opened 2024-06-17
fix: Don’t consider log events as root spans

MikeGoldsmith:

Log events should never be considered root spans even though they do not contain a `trace.parent_id`. A log event that is part of a trace would mark the trace complete as if it's considered a root span.

This change updates the check for whether a span is considered a root span to also check for the meta field `meta.signal_type`. If the field exists and is set to "log", we return false.

- Update `InMemoryCollector.isRootSpan` to check if `meta.signal_type` and has a value of "log"
- Add unit tests for InMemoryCollector to verify expected behaviour for isRootSpan


kentquirk (approved):
This is good.

PR #1214, opened 2024-06-20
feat: implement peers on top of pubsub

kentquirk:

- A basic test 

Still to do:
- more testing
- make it more idiomatically injectable, including the old model


kentquirk:
It was easier to reimplement than to rebase this, so closing in favor of #1215.

PR #1215, opened 2024-06-21
feat: Implement peering with pubsub 

### LINKED ISSUE #1201: Implement Gossip protocol
Based on the experiment we have done in `3.x-work-branch` branch, we learned that using pub/sub model provides multiple benefits to Refinery's overall throughput.

Therefore, we decide to introduce the Gossip system that's backed by a pub/sub abstraction into the existing Refinery.

The first step is to add the gossip abstraction implemented through Redis.

kentquirk:

- A basic test 

Still to do:
- more testing
- make it more idiomatically injectable, including the old model

closes #1201

### DIFF
--- a/app/app_test.go
+++ b/app/app_test.go
@@ -85,17 +85,6 @@ func (w *countingWriterSender) waitForCount(t testing.TB, target int) {
 	}
 }
 
-type testPeers struct {
-	peers []string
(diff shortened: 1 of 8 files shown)


kentquirk:
Superceded by #1220

PR #1217, opened 2024-07-01
fix: Change pubsub interface to use callbacks.

kentquirk:

- The channel-based pubsub interface is really hard to reason about and it's difficult to shut down without race conditions.

- Change to a callback model
- Rewrite the tests and benchmarks


pubsub/pubsub_local.go
```diff
@@ -44,59 +46,52 @@ func (ps *LocalPubSub) Stop() error {
...
-	case ch <- message:
-	case <-ctx.Done():
-		return ctx.Err()
+	ps.mut.Lock()
+	defer ps.mut.Unlock()
+	ps.ensureTopic(topic)
+	for _, sub := range ps.topics[topic] {
+		// don't wait around for slow consumers
+		if sub.cb != nil {
+			go sub.cb(message)
```
VinozzZ, line 71: I think currently if a caller calls `Publish` after `Close`, `sub.cb(message)` will panic. We probably need to check whether `ps.done` is closed before calling the callbacks?

pubsub/pubsub_test.go
```diff
@@ -31,81 +32,126 @@ func newPubSub(typ string) pubsub.PubSub {
...
+	l.lock.Lock()
+	defer l.lock.Unlock()
+	return l.msgs
+}
+
 func TestPubSubBasics(t *testing.T) {
 	ctx := context.Background()
 	for _, typ := range types {
 		t.Run(typ, func(t *testing.T) {
+			fmt.Println("testing", typ)
```
VinozzZ, line 56: remove the print line?

pubsub/pubsub_goredis.go
```diff
@@ -124,30 +125,27 @@ func (ps *GoRedisPubSub) Subscribe(ctx context.Context, topic string) Subscripti
 		for {
 			select {
 			case <-sub.done:
-				close(sub.ch)
+				sub.mut.Lock()
+				sub.cb = nil
```
VinozzZ, line 129: Why do we want to make sure the `cb` is cleared when `sub.Close` is called?
If we don't clear `sub.cb` here, we could get rid of the mutex all together

pubsub/pubsub_goredis.go
```diff
@@ -109,11 +110,11 @@ func (ps *GoRedisPubSub) Publish(ctx context.Context, topic, message string) err
 	return ps.client.Publish(ctx, topic, message).Err()
 }
 
-func (ps *GoRedisPubSub) Subscribe(ctx context.Context, topic string) Subscription {
+func (ps *GoRedisPubSub) Subscribe(ctx context.Context, topic string, callback func(string)) Subscription {
 	sub := &GoRedisSubscription{
 		topic:  topic,
 		pubsub: ps.client.Subscribe(ctx, topic),
```
VinozzZ, line 116: It looks like goredis doesn't keep track of existing topics. That means if the same topic is subscribed multiple times, the same piece of message will be delivered through the network multiple times. For Refinery's use case, that probably is ok since we don't really have multiple subscribers for the same topic as far as I know. Maybe we can add a comment here so that it's recorded?

pubsub/pubsub_local.go
```diff
@@ -9,29 +9,31 @@ import (
 
 // LocalPubSub is a PubSub implementation that uses local channels to send messages; it does
 // not communicate with any external processes.
+// subs are individual channels for each subscription
 type LocalPubSub struct {
 	Config *config.Config `inject:""`
-	subs   []*LocalSubscription
-	topics map[string]chan string
+	topics map[string][]*LocalSubscription
+	done   chan struct{}
```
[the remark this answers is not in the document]
kentquirk: Good catch.

pubsub/pubsub_local.go
kentquirk: That's why I set sub.cb to nil in the Close() functions.

pubsub/pubsub_goredis.go
kentquirk: The point of the sub.cb stuff is to make it possible to close a subscription and not receive any more messages. If we don't care to ever let someone do that, all of this could get simpler.

pubsub/pubsub_goredis.go
kentquirk: I've added the comment, but this is why I had originally created the Topic concept as a separate entity, and we decided to remove it.

pubsub/pubsub_goredis.go
VinozzZ: I see. Once Close is called, the goroutine started in Subscribe will exit when the sub.done case is hit in the select statement. That will also set the sub.cb to nil. I'm trying to understand how sub.cb would get called again after the goroutine has existed

VinozzZ (approved):

PR #1218, opened 2024-07-01
feat: add otel tracing support for Refinery internal operations

VinozzZ:

Enable otel tracing for debugging Refinery internal operations.

- Add otel tracing configurations
- inject a tracer on Refinery startup
- Update `github.com/honeycombio/opentelemetry-proto-go/otlp` to v1.3.1


config/cmdenv.go
```diff
@@ -40,6 +40,7 @@ type CmdEnv struct {
 	HoneycombLoggerAPIKey string     `long:"logger-api-key" env:"REFINERY_HONEYCOMB_LOGGER_API_KEY" description:"Honeycomb logger API key"`
 	LegacyMetricsAPIKey   string     `long:"legacy-metrics-api-key" env:"REFINERY_HONEYCOMB_METRICS_API_KEY" description:"API key for legacy Honeycomb metrics"`
 	OTelMetricsAPIKey     string     `long:"otel-metrics-api-key" env:"REFINERY_OTEL_METRICS_API_KEY" description:"API key for OTel metrics if being sent to Honeycomb"`
+	OTelTracesAPIKey      string     `long:"otel-traces-api-key" env:"REFINERY_OTEL_TRACES_API_KEY" description:"API key for OTel metrics if being sent to Honeycomb"`
```
kentquirk, line 43:
```suggestion
	OTelTracesAPIKey      string     `long:"otel-traces-api-key" env:"REFINERY_OTEL_TRACES_API_KEY" description:"API key for OTel traces if being sent to Honeycomb"`
```
```suggestion
	OTelTracesAPIKey      string     `long:"otel-traces-api-key" env:"REFINERY_OTEL_TRACES_API_KEY" description:"API key for OTel metrics if being sent to Honeycomb"`
```

internal/otelutil/otel_tracing.go
```diff
@@ -0,0 +1,117 @@
...
+		apihost := fmt.Sprintf("%s:443", cfg.APIHost)
+
+		sampleRate := cfg.SampleRate
+		if sampleRate < 1 {
+			sampleRate = 1
+		}
+
+		var sampleRatio float64 = 1.0 / float64(sampleRate)
+
+		headers := map[string]string{
```
kentquirk, line 93: I know this is straight from R3, but I just noticed this is unconditionally adding the API Key headers, which is not what we say in the docs. We should be adding this APIKeyHeader only if APIKey is non-blank. That way we don't put the header if a user is sending the traces to some other OTel endpoint.

kentquirk (approved):
This looks great to enable tracing. Now, of course, we have to start using it, but at least this way it's available.

We should start using it in some places before we release a system that has it in the config, or people will think it's broken. That can be a separate PR, though.

internal/otelutil/otel_tracing.go
VinozzZ: There's a check for APIKey on line 80 to only set up a tracer if APIKey is not empty

PR #1220, opened 2024-07-02
feat: peer management on pubsub via callbacks

### LINKED ISSUE #1201: Implement Gossip protocol

kentquirk:

- Uses the current pubsub system that uses callbacks

- Fix the mock and existing implementations
- Add a few tests
- Update config metadata
- delete legacy redis peering
- make it more idiomatically injectable
- do all the things to make injection work everywhere

closes #1201


internal/peer/pubsub.go
```diff
@@ -0,0 +1,269 @@
...
+// NewPubsubPeers returns a peers collection backed by a pubsub system.
+// It expects members to publish their own presence to the pubsub system every so often
+// as determined by refreshCacheInterval. If they fail to do after peerEntryTimeout, they
+// will be removed from the list of peers.
+// They can also remove themselves from the list of peers by publishing an "unregister" message.
+// The register message is just "register address" and the unregister message is "unregister address".
+func newPubsubPeers(c config.Config) (Peers, error) {
+	return &pubsubPeers{
+		Config: c,
+		Clock:  clockwork.NewRealClock(),
```
VinozzZ, line 95: I don't think we want to instantiate a new clock instance if `Clock` is injected

internal/peer/pubsub.go
```diff
@@ -0,0 +1,269 @@
...
+	case Unregister:
+		p.peers.Remove(cmd.peer)
+	case Register:
+		p.peers.Add(cmd.peer)
+	}
+	p.checkHash()
+}
+
+func (p *pubsubPeers) Start() error {
+	p.peers = generics.NewSetWithTTL[string](peerEntryTimeout)
```
VinozzZ, line 126: This line can be removed since `p.peers` is initialized on line 132

internal/peer/pubsub_test.go
```diff
@@ -0,0 +1,51 @@
...
+				t.Errorf("publicAddr() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+			if got != tt.want {
+				t.Errorf("publicAddr() = %v, want %v", got, tt.want)
+			}
+		})
+	}
+}
+
```
VinozzZ, line 37: How about a test with multiple peers that can get correct number of peers and the callback functions are called when a peer joins or leaves the cluster?

VinozzZ (approved):
I'm more inclined to remove the existing redis peer management logic. Couple reasons that comes to mind:
- since the pub/sub implementation is still using redis, having it as a separate config option feels a bit confusing for me
- does that mean `stress_relief` also needs a `type` option so that customers can choose to stay with the existing stress relief mechanism?
- I can't think of a scenario that customers will want to stay with the existing redis implementation if pub/sub system ultimately enables scaling without loosing data.
- it would be nice that we can just have one redis package in refinery instead of maintaining two implementation

internal/peer/pubsub_test.go
kentquirk: I've been trying to write that test, and it's not real easy to run multiple peers on the same machine. I *think* it's possible, but it might need some gymnastics to make it happen.

Because I'm going on vacation now, I'm going to merge this so it doesn't get stale while I'm gone. We can look at that test another time.

PR #1221, opened 2024-07-03
feat: use pub/sub for stress relief

VinozzZ:

Enable stress relief management for Refinery cluster as a whole rather than only for individual instances.

- broadcast individual stress level through pub/sub to other peers
- calculate cluster stress level when individual cluster level changes or on a timer
- during stress relief, deterministic sampling for a portion of traces based on current cluster stress level
- add tests


collect/collect.go
```diff
@@ -469,23 +463,42 @@ func (i *InMemCollector) processSpan(sp *types.Span) {
...
-		Dataset:     sp.Dataset,
-		TraceID:     sp.TraceID,
-		ArrivalTime: now,
-		SendBy:      now,
-	}
-	// we do want a record of how we disposed of traces in case more come in after we've
-	// turned off stress relief (if stress relief is on we'll keep making the same decisions)
-	i.sampleTraceCache.Record(trace, keep, reason)
+func (i *InMemCollector) ProcessSpanImmediately(sp *types.Span) (processed bool, keep bool) {
+	_, span := otelutil.StartSpanWith(context.Background(), i.Tracer, "collector.ProcessSpanImmediately", "trace_id", sp.TraceID)
```
kentquirk, line 467: Given that we're trying to shed load here because we're stressed, I'm reluctant to generate a span in this routine since it's being called on every span. Can you convince me that it's worth it?

collect/stressRelief.go
```diff
@@ -46,26 +62,41 @@ const (
 	Always
 )
 
+type stressReport struct {
+	key   string
+	level uint
+	// we need to expire these reports after a certian amount of time
```
kentquirk, line 68:
```suggestion
	// we need to expire these reports after a certain amount of time
```

collect/stressRelief.go
```diff
@@ -99,6 +135,26 @@ func (s *StressRelief) Start() error {
 		{Numerator: "memory_heap_allocation", Denominator: "MEMORY_MAX_ALLOC", Algorithm: "sigmoid", Reason: "MaxAlloc"},
 	}
 
+	var hostID string
```
kentquirk, line 138: I just had an idea -- since peers have to generate an identifier anyway, is there any value in adding an interface to Peers to GetCurrentID()? It wouldn't be the hostname, but the IP address. It would correlate with the peer traffic, but no longer would match the k8s name. 

Do you have an opinion?

collect/stressRelief.go
```diff
@@ -119,25 +179,62 @@ func (s *StressRelief) Start() error {
...
+	level, err := strconv.Atoi(parts[1])
+	if err != nil {
+		return nil, err
+	}
+
+	return newStressReliefMessage(uint(level), parts[0]), nil
+}
+
+func (s *StressRelief) onStressLevelUpdate(ctx context.Context, msg string) {
+	stressMsg, err := unmarshalStressRelifMessage(msg)
```
kentquirk, line 214:
```suggestion
	stressMsg, err := unmarshalStressReliefMessage(msg)
```

collect/stressRelief.go
```diff
@@ -119,25 +179,62 @@ func (s *StressRelief) Start() error {
...
+
+func newStressReliefMessage(level uint, peerID string) *stressReliefMessage {
+	return &stressReliefMessage{level: level, peerID: peerID}
+}
+
+func (msg *stressReliefMessage) String() string {
+	return msg.peerID + ":" + fmt.Sprint(msg.level)
+}
+
+func unmarshalStressRelifMessage(msg string) (*stressReliefMessage, error) {
```
kentquirk, line 195:
```suggestion
func unmarshalStressReliefMessage(msg string) (*stressReliefMessage, error) {
```

collect/stressRelief.go
```diff
@@ -119,25 +179,62 @@ func (s *StressRelief) Start() error {
...
+	if len(msg) < 2 {
+		return nil, fmt.Errorf("empty message")
+	}
+
+	parts := strings.SplitAfter(msg, ":")
```
kentquirk, line 200: I think you want Split, not SplitAfter -- SplitAfter is going to keep the : which I don't think you want.

collect/stressRelief.go
```diff
@@ -119,25 +179,62 @@ func (s *StressRelief) Start() error {
...
 	case "never", "":
 		s.mode = Never
 	case "monitor":
-		// If we're switching into monitor mode from some other state (which
-		// happens on startup), we will start up in stressed mode for a
-		// configurable time to try to make sure that we can handle the load
-		// before we start processing it in earnest. This is to help address the
-		// problem of trying to bring a new node into an already-overloaded
-		// cluster. If the time is 0 we won't do this at all.
-		if s.mode != Monitor && cfg.MinimumStartupDuration != 0 {
```
kentquirk, line 136: Since we're no longer using MinimumStartupDuration, we should obsolete it in the docs and remove it from the test code.

collect/stressRelief.go
```diff
@@ -305,28 +406,70 @@ func (s *StressRelief) Recalc() {
...
+	}
+
+	s.lock.Lock()
+	defer s.lock.Unlock()
+
+	s.stressLevels[report.key] = report
+	var total float64
+	availablePeers := 0
+	for _, report := range s.stressLevels {
+		// TODO: maybe make the expiration time configurable
```
kentquirk, line 455: Hmmm...maybe this should also be the same value as peer expiration time.

kentquirk (commented):
I haven't had time to finish my review but I have a meeting, so here are some things to get started on; more later.

collect/stressRelief.go
VinozzZ: I like the idea of having a uniformed way within Refinery to identify a n instance. Since we have stress relief metrics that contains hostname already, I think it's fine to use IP address as the identifier

collect/stressRelief.go
VinozzZ: Good catch! I switched to SplitN since we already have the assumption that the message should only contain 2 parts separated by :

collect/collect.go
VinozzZ: I added here for debugging purpose for stress relief mode. I'm ok to remove it once we are done with testing. Plus, I don't think users will turn on tracing all the time. The overhead should only occur if they have tracing turned on

collect/stressRelief.go
```diff
@@ -119,25 +164,62 @@
...
+	if len(parts) != 2 {
+		return nil, fmt.Errorf("invalid message format")
+	}
+
+	level, err := strconv.ParseUint(parts[1], 10, 64)
+	if err != nil {
+		return nil, err
+	}
+
+	return newStressReliefMessage(uint(level), parts[0]), nil
```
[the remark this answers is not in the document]
kentquirk: I hate that this won't go away, even though the issue was fixed AND the alert was closed.

kentquirk (approved):
Looks good now!

PR #1227, opened 2024-07-17
maint: Add jitter to peer traffic, fix startup

kentquirk:

- Add jitter to peer traffic so we don't get thundering herds
- Inject the right things for proper startup


VinozzZ (approved):
💯

PR #1229, opened 2024-07-17
maint: add instrumentation for GoRedisPubSub

VinozzZ:

Add instrumentation for GoRedisPubSub to gain more insight on how the redis pubsub system works

- add tracer to `GoRedisPubSub`
- fix tests
- define a type for the pubsub subscription callback function

### DIFF
--- a/pubsub/pubsub_goredis.go
+++ b/pubsub/pubsub_goredis.go
@@ -161,11 +161,11 @@ func (ps *GoRedisPubSub) Subscribe(ctx context.Context, topic string, callback S
 				})
 				ps.Metrics.Count("redis_pubsub_received", 1)
 
-				go func(cbCtx context.Context, span trace.Span) {
+				go func(cbCtx context.Context, span trace.Span, payload string) {
 					defer span.End()
 
-					sub.cb(cbCtx, msg.Payload)
-				}(receiveCtx, span)
+					sub.cb(cbCtx, payload)
+				}(receiveCtx, span, msg.Payload)
 			}
 		}
 	}()
--- a/pubsub/pubsub_goredis.go
+++ b/pubsub/pubsub_goredis.go
@@ -5,7 +5,10 @@ import (
 	"strings"
 	"sync"
 
+	"go.opentelemetry.io/otel/trace"
+
 	"github.com/honeycombio/refinery/config"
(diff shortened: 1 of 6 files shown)


pubsub/pubsub_goredis.go
```diff
@@ -25,6 +28,7 @@ type GoRedisPubSub struct {
 	Config  config.Config   `inject:""`
 	Logger  logger.Logger   `inject:""`
 	Metrics metrics.Metrics `inject:"metrics"`
+	Tracer  trace.Tracer    `inject:"tracer"`
```
kentquirk, line 31: If we don't expect to inject multiple different tracers, we don't need to name them. You could remove the name here, and in main where it's created.

pubsub/pubsub_goredis.go
```diff
@@ -139,8 +153,18 @@ func (ps *GoRedisPubSub) Subscribe(ctx context.Context, topic string, callback f
 				if msg == nil {
 					continue
 				}
+				ctx, span := otelutil.StartSpanMulti(context.Background(), ps.Tracer, "GoRedisPubSub.Receive", map[string]interface{}{
```
kentquirk, line 156: I'm slightly worried about the performance overhead of creating a context every time. How about creating just one outside the for loop and reusing it for each span?

kentquirk (approved):

pubsub/pubsub_goredis.go
```diff
@@ -139,8 +154,18 @@ func (ps *GoRedisPubSub) Subscribe(ctx context.Context, topic string, callback f
...
+					"message":            msg.Payload,
+				})
 				ps.Metrics.Count("redis_pubsub_received", 1)
-				go sub.cb(msg.Payload)
+
+				go func(cbCtx context.Context, span trace.Span) {
+					defer span.End()
+
+					sub.cb(cbCtx, msg.Payload)
+				}(receiveCtx, span)
```
kentquirk, line 168: Why did you pass the context and span as parameters, but not the payload? I was originally thinking you needed none of them, but now, after thinking it through, it's probably a good idea to pass the message payload as well.

VinozzZ: Good catch, I just changed it

PR #1231, opened 2024-07-17
feat: Health/Ready system imported from R3

kentquirk:

- Refinery needs a proper health monitoring system that can watch multiple subsystems and deliver both "alive" and "ready" responses to health check endpoints, both for grpc and http endpoints. This is that system, imported from R3 and improved a bit.

- Import health and tests from R3 and clean up its documentation and some of the internal names
- Hook it up to the router including the grpc health endpoints
- Tie it in to stress relief so that it's monitoring something (otherwise it will never be ready)
- Inject it into the app

### DIFF
--- a/route/route.go
+++ b/route/route.go
@@ -214,6 +214,7 @@ func (r *Router) LnS(incomingOrPeer string) {
 		IdleTimeout: r.Config.GetHTTPIdleTimeout(),
 	}
 
+	r.donech = make(chan struct{})
 	if r.Config.GetGRPCEnabled() && len(grpcAddr) > 0 {
 		l, err := net.Listen("tcp", grpcAddr)
 		if err != nil {
--- a/route/route.go
+++ b/route/route.go
@@ -23,6 +23,7 @@ import (
 	"github.com/pelletier/go-toml/v2"
 	"github.com/vmihailenco/msgpack/v5"
 	"google.golang.org/grpc"
+	healthserver "google.golang.org/grpc/health"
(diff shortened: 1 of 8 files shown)


route/route.go
```diff
@@ -947,6 +978,46 @@ func (r *Router) Watch(req *grpc_health_v1.HealthCheckRequest, server grpc_healt
...
+		} else {
+			r.hsrv.SetServingStatus(svc, grpc_health_v1.HealthCheckResponse_NOT_SERVING)
+		}
+	}
+
+	r.doneWG.Add(1)
+	go func() {
+		defer r.doneWG.Done()
+		// TODO: Does this time need to be configurable?
+		watchticker := time.NewTicker(3 * time.Second)
```
VinozzZ, line 1002: I wonder if we could make this configurable so that it can be set based on the probe in k8s https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes

VinozzZ (approved):

route/route.go
kentquirk: It could definitely be configurable, but the k8s timing is unrelated. The longest you can go without detecting a failure is the sum of this time plus the k8s time. Is there a real case for ever trying to be significantly faster than this? Also, it's rare enough that there's no real benefit in going much slower. To me, it seems like the practical range is from 1-10 seconds, and 3 is a good compromise.

kentquirk: I'm going to merge this and we can add configuration later if we decide we want it.

PR #1236, opened 2024-07-18
fix: Root spans must have a non-empty parent ID field

### LINKED ISSUE #1234: Refinery does not properly handle empty parentId strings when determining root spans
The Refinery considers a span to not be a root if the parentId is a string, but it doesn't check if the string is empty. This can lead to incorrect identification of root spans.
https://github.com/honeycombio/refinery/blob/main/collect/collect.go#L604
 lso this will cause delays on the span being transmitted, because Refinery will wait for the parent span to show up, for TraceTimeout (60s by default). And in this case it never will, so any of these real rootspans are actually delayed by 60s for sending to Honeycomb.

Steps to Reproduce:
1. Create a span with an empty string as the parentId.
2. Observe that the span is not considered a root, even though it should be.

Expected Behavior:

Spans with an empty parentId string should be considered root spans.

- Refinery version: v2.6.1
- Operating system: Linux

MikeGoldsmith:

Span's with a `parentID` field with an empty value `""` are considered root spans. This PR updates the logic to determine root spans to check that both the parentID field exists and is not empty for it to be considered a root span.

- Fixes #1234 

- Check the parentID field is both present and has a value when checking if the span is a root span
- Update tests to verify updated behaviour


kentquirk (approved):
This is counter to spec for some reason, but I don't see any reason we shouldn't do this. Thanks.

PR #1238, opened 2024-07-19
fix: downgrade otel back to v1.27.0

VinozzZ:

make sure we don't run into `* conflicting Schema URL: https://opentelemetry.io/schemas/1.25.0 and https://opentelemetry.io/schemas/1.26.0`

- downgrade otel back to v1.27.0


VinozzZ:
We will remove otel-config-go instead

PR #1240, opened 2024-07-19
fix: remove otel-config-go as a dependency

VinozzZ:

The maintenance effort required to keep semantic convention in sync between `otel-config-go` and `otel` has led us to remove `otel-config-go` as a dependency

- configure otel using vanilla otel sdk


internal/otelutil/otel_tracing.go
```diff
@@ -106,17 +116,37 @@ func SetupTracing(cfg config.OTelTracingConfig, resourceLibrary string, resource
...
+			otlptracehttp.WithHeaders(headers),
+			otlptracehttp.WithCompression(otlptracehttp.GzipCompression),
+		),
 	)
 	if err != nil {
-		log.Fatalf("failure configuring otel: %v", err)
+		log.Fatalf("failure configuring otel trace exporter: %v", err)
+	}
+
+	otel.SetTextMapPropagator(propagation.NewCompositeTextMapPropagator(
```
kentquirk, line 134: Why do we need this? I don't think we want refinery to do baggage propagation or anything having to do with b3?

VinozzZ: I copied from otel-config-go lol

kentquirk (approved):
Thanks. Please do a `go mod tidy` before merging.

PR #1241, opened 2024-07-23
feat: gossip config reload information

kentquirk:

- When a configuration is reloaded (and it has changed from the previous version) this publishes the news of the change by publishing the configuration hashes. Other peers will receive this message and immediately reload their configs.

- Because config needs to be loaded to start the app, it can't depend on pubsub which hasn't been set up yet
- So I created configwatcher, which lives outside config and can have both config and pubsub injected
- It registers with config for callbacks, AND with pubsub for messages from peers
- It publishes news of config changes with a timestamp
- When it receives a message it checks the hashes; if they're different, it reloads
- It publishes news of config changes only if it hasn't received a message in a little while (controlled by config)
- Config tests had to be moved to config_test package in order to test reloading in this new world


internal/configwatcher/watcher.go
```diff
@@ -0,0 +1,109 @@
...
+func (cw *ConfigWatcher) ReloadCallback(cfgHash, rulesHash string) {
+	ctx := context.Background()
+	ctx, span := otelutil.StartSpanMulti(ctx, cw.Tracer, "ConfigWatcher.ReloadCallback", map[string]any{
+		"new_config_hash": cfgHash,
+		"new_rules_hash":  rulesHash,
+	})
+	defer span.End()
+
+	// don't publish if we have recently received a message (this avoids storms)
+	now := time.Now()
```
VinozzZ, line 47: Should we use clockwork's clock here too?

internal/configwatcher/watcher.go
```diff
@@ -0,0 +1,109 @@
...
+	if now.Sub(cw.msgTime) < time.Duration(cw.Config.GetGeneralConfig().ConfigReloadInterval) {
```
VinozzZ, line 48: I wonder if we need to have a mutex to protect access to `cw.msgTime`. It's assigned in the pub/sub goroutine and read from the `monitor` goroutine

internal/configwatcher/watcher.go
```diff
@@ -0,0 +1,109 @@
...
+
+// SubscriptionListener listens for messages on the config pubsub topic and reloads the config
+// if a new set of hashes is received.
+func (cw *ConfigWatcher) SubscriptionListener(ctx context.Context, msg string) {
+	_, span := otelutil.StartSpanWith(ctx, cw.Tracer, "ConfigWatcher.SubscriptionListener", "message", msg)
+	defer span.End()
+
+	// parse message as a time in RFC3339 format
+	msgTime, err := time.Parse(time.RFC3339, msg)
+	if err == nil {
```
VinozzZ, line 66: Is this a typo? should this be `err != nil`?

internal/configwatcher/watcher.go
```diff
@@ -0,0 +1,109 @@
...
+	if cw.Config.GetGeneralConfig().ConfigReloadInterval != 0 {
+		go cw.monitor()
+	}
+	cw.subscr = cw.PubSub.Subscribe(context.Background(), ConfigPubsubTopic, cw.SubscriptionListener)
+	cw.Config.RegisterReloadCallback(cw.ReloadCallback)
+	return nil
+}
+
+func (cw *ConfigWatcher) Stop() error {
+	cw.subscr.Close()
```
VinozzZ, line 106: Will this race with an ongoing `Reload` ? Maybe we can close `cw.done` first to make sure no code is going to utilize the pub/sub channel anymore

VinozzZ (approved):

internal/configwatcher/watcher.go
kentquirk: Eeek! Thanks.

PR #1244, opened 2024-07-23
maint: Update licenses

TylerHelmuth:

- Updates licenses
- Re-enables license verification

Closes https://github.com/honeycombio/refinery/issues/1242

- ran `update-licenses`
- updated `verify-licenses` to run the logic again


Makefile
```diff
@@ -77,10 +77,6 @@ update-licenses: install-tools
 
 .PHONY: verify-licenses
 verify-licenses:
```
JamieDanielson, line 79:
```suggestion
verify-licenses: install-tools
```

do you need this here too?

TylerHelmuth: Yup

codeboten:
```
Only in LICENSES: crypto
Common subdirectories: temp/github.com and LICENSES/github.com
Common subdirectories: temp/golang.org and LICENSES/golang.org
Common subdirectories: temp/google.golang.org and LICENSES/google.golang.org
Common subdirectories: temp/go.opentelemetry.io and LICENSES/go.opentelemetry.io
Common subdirectories: temp/gopkg.in and LICENSES/gopkg.in
Common subdirectories: temp/go.uber.org and LICENSES/go.uber.org
Only in LICENSES: LICENSE
Only in LICENSES: vendor
LICENSES directory must be updated. Run make update-licenses
```

Makefile
```diff
@@ -68,16 +68,16 @@ clean:
 
 .PHONY: install-tools
 install-tools:
-	go install github.com/google/go-licenses@v1.0.0
+	go install github.com/google/go-licenses/v2@v2.0.0-alpha.1
```
JamieDanielson, line 71: Do we want to use an alpha version here? Also wondering if we should separate this into an env var further up so it's easier to spot and change (similar to the dockerize versions), vs in the middle of the command. No strong opinion here though.

JamieDanielson (approved):
Added a question but nothing blocking

Makefile
kentquirk: This tool has been slow to evolve with many changes in the 1.x versions; we were stuck on 1.0 because of problems with later versions. I'd personally rather see us standardize on 2.x even if it's alpha.

MikeGoldsmith: I don't think there would be a need to dynamically update the liceses version and as Kent said, they release versions *very slowly*. Let's go with this as is for now, and add the env var later if needed.

MikeGoldsmith (approved):

LICENSES/github.com/honeycombio/husky/LICENSE
friedrichg, line 1: This got rid of the main LICENSE fail for the repo @TylerHelmuth @codeboten. was that intended?

PR #1245, opened 2024-07-23
fix: make sure stress relief pub/sub topic is consistent

VinozzZ:

Make sure subscriber and publisher uses the same channel for stress relief

- define a constant for stress relief pub/sub topic
- use the constant in test and stress relief code


kentquirk (approved):
Yes.

PR #1249, opened 2024-07-25
fix: sharder should use peer identity from Peers package

VinozzZ:

`DeterministicSharder` wasn't updated to use the new peer identification system to determine which peer a given trace belongs to. This resulted Refinery couldn't identify itself in the peer list within the sharding system. 

- Consolidate and unify code for retrieving peer identity
- replace `:` with `|` as the delimiter in stress relief message 
- make sure `filepeer` returns `PeerListenAddr` as its public address

### DIFF
--- a/app/app_test.go
+++ b/app/app_test.go
@@ -327,12 +327,7 @@ func TestPeerRouting(t *testing.T) {
 	// Parallel integration tests need different ports!
 	t.Parallel()
 
-	peers := &peer.MockPeers{
-		Peers: []string{
-			"http://localhost:11001",
-			"http://localhost:11003",
-		},
-	}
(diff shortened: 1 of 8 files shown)


VinozzZ:
> I think this looks fine. I am realizing, though, that FilePeers is now almost pointless, because its main reason for existing for more than one node is so that someone can avoid running Redis. But now we depend on Redis for more than just peers, so even if you do FilePeers you _still_ have to do Redis.

The other place that depends on Redis is stress relief and configs. Using `filepeer` means, peers will act independently in both stress relief and config change system. Essentially, if a Refinery cluster is using `filepeer`, they won't get much benefit from this release. We can document it, but at least it's not breaking existing behavior.

PR #1256, opened 2024-07-26
fix: allow a single node to activate stress relief mode during significant load increase

VinozzZ:

Without resolving trace locality issue, a single peer can receive a large trace that significantly raise its stress level than the rest of the cluster. To address this issue, we can allow individual refineries to go into stress relief mode if their own stress is too high, even if the cluster's isn't.

- use the max of the individual stress level and the cluster stress level as the overall stress level when calculating stress relief activation and deactivation
- record the stress level that determined stress relief activation as `stress_level` metric
- add tests

### DIFF
--- a/collect/stressRelief.go
+++ b/collect/stressRelief.go
@@ -35,8 +35,8 @@ var _ StressReliever = &MockStressReliever{}
 type MockStressReliever struct {
 	IsStressed              bool
 	SampleDeterministically bool
-	SampleRate              uint
 	ShouldKeep              bool
+	SampleRate              uint
 }
 
 func (m *MockStressReliever) Start() error                                   { return nil }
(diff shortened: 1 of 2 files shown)


collect/stressRelief.go
```diff
@@ -414,18 +419,18 @@ func (s *StressRelief) Recalc() uint {
 		// If it's off, should we activate it?
 		if !s.stressed && s.overallStressLevel >= s.activateLevel {
 			s.stressed = true
-			s.Logger.Warn().WithField("cluster_stress_level", s.overallStressLevel).WithField("stress_formula", s.formula).WithField("reason", s.reason).Logf("StressRelief has been activated")
+			s.Logger.Warn().WithField("stress_level", s.overallStressLevel).WithField("stress_formula", s.formula).WithField("reason", s.reason).Logf("StressRelief has been activated")
```
kentquirk, line 428: What do you think of a single log message here and on deactivation with all 3 stress levels in it -- individual, cluster, and net?

kentquirk (approved):
This is great, and it also preserves our previous `stress_level` metric name, which is a nice bonus.

I made one suggestion that we add all levels to the single log message.

PR #1259, opened 2024-07-30
refactor: Remove error returns from config functions, fix tests.

kentquirk:

- Since the introduction of validation of config, the config functions can't fail, and the error returns are useless. 

- Remove error returns from all the config getters
- Fix all the places we used (or didn't use them)
- Fix a couple of tests that assumed things that are no longer true

### DIFF
--- a/test/config.yaml
+++ b/test/config.yaml
@@ -13,13 +13,11 @@ Logger:
 LegacyMetrics:
   Enabled: true
   Dataset: refinery_metrics
-  APIKey: $REFINERY_HONEYCOMB_API_KEY
   APIHost: https://api-dogfood.honeycomb.io
 
 OTelMetrics:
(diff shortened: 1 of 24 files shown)


test/config.yaml
```diff
@@ -13,13 +13,11 @@ Logger:
 LegacyMetrics:
   Enabled: true
   Dataset: refinery_metrics
-  APIKey: $REFINERY_HONEYCOMB_API_KEY
```
VinozzZ, line 16: Why are these deleted?

VinozzZ (approved):
Yahoo!

test/config.yaml
kentquirk: Because the $ syntax doesn't work, but there are environment variables that are automatic if they're set.

PR #1260, opened 2024-07-31
feat: unregister peer asap on shutdown

VinozzZ:

In order to recalculate new destination for traces during shutdown as soon as possible, we need to announce peer unregistration first thing.

- use the `done` channel from catching the termination signal in main.go in peer management


kentquirk (approved):
👍

PR #1263, opened 2024-08-01
feat: Allow more complex key behavior

### LINKED ISSUE #634: Allow setting API Key on Refinery
Right now, refinery requires api key be attached to the incoming events.  There could be a scenario where many OTel collectors are in place behind refinery. Each having to maintain an API Key to Honeycomb.

Instead, refinery is the endpoint anyways, if it could have a setting to use an API Key in the refinery config and use that when sending events to Honeycomb, there is now just 1 place to have to maintain an API Key.

Putting a proxy after refinery, hijacking the requests, and attaching the api key there.

**Additional context**

kentquirk:

- People want to use Refinery to help manage their Honeycomb keys. 
- They've wanted it for a while. See #634.

- [x] Add SendKeys and SendKeysMode to config
- [x] Implement a new route handler for http
- [x] Document it in the metadata
- [x] Document it in the readme
- [x] Implement it for OTLP traces
- [x] Implement it for OTLP logs

Closes #634.


VinozzZ (approved):
It looks good to me. I think we should add some tests for the `CheckAndMaybeReplaceKey` function

kentquirk:
It was a good idea to write some more tests. I found a couple of bugs that way. Thanks for the nudge.

PR #1265, opened 2024-08-06
feat: make InMemoryCache concurrent safe

VinozzZ:

To be able to process incoming traces while redistributing traces when a peer membership change happens, the trace cache needs to be able to support concurrent operations.

- Add mutex to each cache access operations
- Add benchmark for cache operations

### DIFF
--- a/collect/cache/cache.go
+++ b/collect/cache/cache.go
@@ -10,11 +10,13 @@ import (
 	"github.com/honeycombio/refinery/types"
 )
 
-// Cache is a non-threadsafe cache. It must not be used for concurrent access.
+// Cache is a multi-threadsafe cache. It is used to store traces that have
+// been received but not yet sent.
 type Cache interface {
 	// Set adds the trace to the cache. If it is kicking out a trace from the cache
 	// that has not yet been sent, it will return that trace. Otherwise returns nil.
 	Set(trace *types.Trace) *types.Trace
+	//Get returns the trace with the given traceID, or nil if it is not in the cache.
 	Get(traceID string) *types.Trace
 	// GetAll is used during shutdown to get all in-flight traces to flush them
 	GetAll() []*types.Trace
@@ -39,6 +41,7 @@ type DefaultInMemCache struct {
 
 	// traceBuffer is a circular buffer of currently stored traces
 	traceBuffer []*types.Trace
+
 	// currentIndex is the current location in the circle.
 	currentIndex int
 }
--- a/collect/cache/cache.go
+++ b/collect/cache/cache.go
@@ -1,6 +1,7 @@
 package cache
 
 import (
+	"sync"
 	"time"
 
 	"github.com/honeycombio/refinery/generics"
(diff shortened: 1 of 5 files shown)


VinozzZ:
**Single threaded Cache**
```
BenchmarkCache_Set-12                              	100000000	      149.0 ns/op	      0 B/op	      0 allocs/op
BenchmarkCache_Get-12                              	150195169	       96.85 ns/op	      0 B/op	      0 allocs/op
BenchmarkCache_TakeExpiredTraces-12                	 166215	   177033 ns/op	     60 B/op	      2 allocs/op
BenchmarkCache_RemoveTraces-12                     	  66396	  1124842 ns/op	     53 B/op	      2 allocs/op
```
**Concurrent Cache**
```
BenchmarkCache_Set-12                              	100000000	      165.5 ns/op	      0 B/op	      0 allocs/op
BenchmarkCache_Get-12                              	122417331	       94.92 ns/op	      0 B/op	      0 allocs/op
BenchmarkCache_SetConcurrent-12                    	25489023	      496.8 ns/op	      0 B/op	      0 allocs/op
BenchmarkCache_TakeExpiredTraces-12                	 143635	   189503 ns/op	     67 B/op	      2 allocs/op
BenchmarkCache_RemoveTraces-12                     	  65197	  1108219 ns/op	     54 B/op	      2 allocs/op
```

collect/cache/cache.go
```diff
@@ -23,6 +24,8 @@ type Cache interface {
 	TakeExpiredTraces(now time.Time) []*types.Trace
 }
 
+var _ Cache = (*DefaultInMemCache)(nil)
+
 // DefaultInMemCache keeps a bounded number of entries to avoid growing memory
 // forever. Traces are expunged from the cache in insertion order (not access
```
kentquirk, line 30: There are now a lot of misleading comments in this file -- some are from this PR, some are older than that (like this one, which is no longer quite true and hasn't been for a while). 

Could you try to clean things up a bit? There are a couple of comments about thread safety, a couple about the circular buffer.

collect/collect.go
```diff
@@ -333,13 +330,16 @@ func (i *InMemCollector) collect() {
...
-				i.sendTracesInCache(time.Now())
-				i.checkAlloc()
+				select {
+				case <-i.done:
+				default:
+					i.sendTracesInCache(time.Now())
+					i.checkAlloc()
+
+					// Briefly unlock the cache, to allow test access.
+					runtime.Gosched()
```
kentquirk, line 340: I don't think we need these lines anymore.

collect/collect.go
```diff
@@ -752,21 +752,23 @@ func (i *InMemCollector) Stop() error {
...
+			i.Transmission.Flush()
+		}
 
+	}
 	i.sampleTraceCache.Stop()
 
 	return nil
 }
 
+func (i *InMemCollector) redistributeTraces() {
```
kentquirk, line 765: I don't think this is supposed to be here yet.

kentquirk (approved):
A couple of notes but otherwise OK.

VinozzZ:
Closing since traces are stored as pointers in the cache

PR #1266, opened 2024-08-08
feat: Add SpanLimit (includes some config changes)

### LINKED ISSUE #1257: Add span count limit as an alternative way to kick off trace evaluation

Sometimes, Refinery will get a giant trace with tens of thousands of spans.

This is a good way to crash Refinery because its caches are based on trace count, not span count, and because of trace locality only one Refinery will have an issue. This causes stress issues and can even crash Refinery.

Allow a configuration option next to TraceTimeout -- let's call it `SpanLimit` -- which, when we add a span to a trace and increase that trace's span count above the limit, immediately marks the trace as ready for decision. 

This will force a trace to be decided when it gets "too big", and prevent a trace from becoming unmanageably big.

We want a configuration value to avoid breaking existing customers.

Crashing.

**Additional context**

Honeycomb's prod refinery was crashing today, possibly because of this.

kentquirk:

#1257 -- helps to keep Refinery from exploding when a giant trace arrives.

- Adds a SpanLimit that kicks off trace evaluation as soon as the number of spans in a trace exceeds that value
- Moves some config stuff around (reducing the size of Config) and fixes the tests that touch it
- Fixes #1257

### DIFF
--- a/app/app_test.go
+++ b/app/app_test.go
@@ -96,11 +96,13 @@ func newStartedApp(
 	enableHostMetadata bool,
 ) (*App, inject.Graph) {
 	c := &config.MockConfig{
-		GetSendDelayVal:          1 * time.Millisecond,
-		GetTraceTimeoutVal:       10 * time.Millisecond,
-		GetMaxBatchSizeVal:       500,
+		GetTracesConfigVal: config.TracesConfig{
+			SendTicker:   config.Duration(2 * time.Millisecond),
+			SendDelay:    config.Duration(1 * time.Millisecond),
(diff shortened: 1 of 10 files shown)


VinozzZ (approved):
Looks good to me! Would nice to also have a unit test in `collect_test.go` to confirm a trace is sent when its span count surpass the configured SpanLimit. 

I'm also wondering if the `SpanLimit` can be used in `CacheImpact` calculation so that we can eject giant traces first when there's memory pressure

kentquirk:
I added that test. 

We already use a cache impact calculation, which is a combination of memory usage and duration in the cache to eject the largest traces first.

Issue #1258, opened 2024-07-29
Redistribute spans when the number of peers changes


kentquirk:
**Is your feature request related to a problem? Please describe.**

When the number of peers changes, the decision node for traces in flight may move to a new node. We should also evaluate spans that are currently stored and re-evaluate if they need to move to a new peer. This will enable cluster size changes to not break telemetry in flight (different decisions causing missing spans in Honeycomb).

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

PR #1268, opened 2024-08-12
feat: redistribute traces on peer membership changes

VinozzZ:

Draining incoming and peer queue on peer membership changes
#1258 

A `RedistributeNotifier` is introduced in the collector in this PR. 
- It's initialized on refinery start up
- When a peer membership change is broadcasted, the notifier immediately signals the `collect()` goroutine to start the redistribution process
- When multiple peer membership changes happen in a row, it ignores the signal if there's already a redistribution process running

### DIFF
--- a/collect/cache/cache.go
+++ b/collect/cache/cache.go
@@ -18,10 +18,17 @@ type Cache interface {
 	// GetAll is used during shutdown to get all in-flight traces to flush them
 	GetAll() []*types.Trace
 
+	GetCacheSize() int
+
 	// Retrieve and remove all traces which are past their SendBy date.
 	// Does not check whether they've been sent.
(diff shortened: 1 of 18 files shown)


collect/cache/cache.go
```diff
@@ -18,11 +18,18 @@ type Cache interface {
 	// GetAll is used during shutdown to get all in-flight traces to flush them
 	GetAll() []*types.Trace
 
+	GetCacheSize() int
```
kentquirk, line 21: I think this shoulid be renamed to `GetCacheCapacity` and include a comment on what units it's in (traces, bytes, spans, etc).

kentquirk (approved):
Looks good. One suggestion on a rename.

PR #1275, opened 2024-08-14
feat: allow root. in field list for dynamic sampler

### LINKED ISSUE #1130: Feature Request: Allow trace root context in Dynamic Sampler FieldLIsts
Not a problem, just an enhancement of the current use of root context that can be added to Rule conditions.

Currently, in rule conditions, you can prefix a field with `root.` to indicate that you only want the condition met if the attribute value is on the root span of a trace.

It would be really nice to add this to the capability of the FieldList in the dynamic sampler configurations, so that you can use a trace's root span data only, to reduce cardinality in the dynamic sampler keyspace.

Currently, users do this in instrumentation or with the collector with the transform processor.  But it would support automatic instrumentation better and be less work for customers to just be able to specify the desire to only use the root span in dynamic sampler field lists.

**Additional context**

VinozzZ:

To reduce the cardinality in sampler keys for dynamic sampler in Refinery, it's helpful to be able to only select fields on root spans.

fixes #1130 

- parse field list on dynamic sampler startup to see if there's `root.` in field list
- store fields for root span only separately from fields applies to all spans in `traceKey`
- loop through both field list to make sure keys are correctly populated

### DIFF
--- a/collect/collect.go
+++ b/collect/collect.go
@@ -495,7 +495,7 @@ func (i *InMemCollector) processSpan(sp *types.Span) {
 			timeout = 60 * time.Second
 		}
 
-		now := time.Now()
+		now := i.Clock.Now()
 		trace = &types.Trace{
 			APIHost:     sp.APIHost,
 			APIKey:      sp.APIKey,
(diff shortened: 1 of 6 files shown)


kentquirk (approved):
Looks good!

PR #1278, opened 2024-08-15
fix: allow draining traces even if only 1 peer left

VinozzZ:

When in peer mode, there could be a situation that the load is low enough for just one refinery instance. We should allow a scale down event to happen even if there's only one refinery left

- change the condition to run redistribution logic to as long as there's 1 peer in the peer list
- add `trace_redistribution_count` metric to track how often a redistribution event has been triggered.


kentquirk (approved):
👍

PR #1281, opened 2024-08-15
fix: record previous value of sampler counter metrics so they report correctly

### LINKED ISSUE #1279: Counter-type metrics for samplers are broken
**Versions**

- Refinery: all recent

**Steps to reproduce**

1. Use an EMADynamic sampler (or anything that has an AdjustmentInterval)
2. Send its metrics to Honeycomb
3. Note that the metric count seems to increase exponentially

The problem is that Refinery is reading a counter and wanting to increment the metrics counter to get the same result -- but never records the previous value, so is just always incrementing by the value of the counter.

This is a bug in several sampler implementations. Can we centralize the code to a helper function?

**Additional context**
[Pollinators thread](https://honeycombpollinators.slack.com/archives/C01BZ3J7RMJ/p1723738971140929)

kentquirk:

- Fixes #1279 

- Update the previous values 
- Fix a test that was sensitive to existing standard envvar

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -614,13 +614,15 @@ func TestHoneycombLoggerConfig(t *testing.T) {
 	config, rules := createTempConfigs(t, cm, rm)
 	defer os.Remove(rules)
 	defer os.Remove(config)
+	// Set the environment variable to test that it overrides the config
+	os.Setenv("REFINERY_HONEYCOMB_API_KEY", "321cba")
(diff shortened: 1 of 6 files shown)


VinozzZ (approved):
Looks good to me. Is there a way to add test for the sampler to make sure we are not going to break it in the future?

kentquirk:
Well, I tried, but it ended up being an extremely picky, implementation-dependent test -- not only on the implementation of the sampler, but also the implementations of metrics AND MockMetrics. I didn't think it was worth it, so I left it off.

PR #1284, opened 2024-08-16
fix: set up tls for redis when it's enabled

VinozzZ:

Currently, we are not setting the tls config for go-redis even through we do have `UseTLS` configuration option.

- config tls with go-redis when it's enabled

### DIFF
--- a/pubsub/pubsub_goredis.go
+++ b/pubsub/pubsub_goredis.go
@@ -2,6 +2,7 @@ package pubsub
 
 import (
 	"context"
+	"crypto/tls"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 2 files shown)


pubsub/pubsub_goredis.go
```diff
@@ -69,6 +70,20 @@ func (ps *GoRedisPubSub) Start() error {
...
+				MinVersion: tls.VersionTLS12,
+			}
+
+			if tlsInsecure {
+				tlsConfig.InsecureSkipVerify = true
+			}
+
+			options.TLSConfig = tlsConfig
+		}
+
```
robbkidd, line 86: I notice that ps.Config.Get* functions return two things: a value and an error.

What do you think of:

```suggestion
		useTLS, err := ps.Config.GetUseTLS()
		if err != nil {
			return err
		}
		if useTLS {
			tlsInsecure, err := ps.Config.GetUseTLSInsecure()
			if err != nil {
				return err
			}
			options.TLSConfig = &tls.Config{
				MinVersion: tls.VersionTLS12,
				InsecureSkipVerify: tlsInsecure,
			}
		}
```

VinozzZ: I thought we have fixed the config interface for all getters to not return an error?

robbkidd: Ah. Yes, on `main`. I was looking at this code path in `v2.7.0` which still have the error returns.

Then how about? … 

```go
		if ps.Config.GetUseTLS() {
			options.TLSConfig = &tls.Config{
				MinVersion: tls.VersionTLS12,
				InsecureSkipVerify: ps.Config.GetUseTLSInsecure(),
			}
		}
```

VinozzZ:
We should not squash merge this PR since we need to fix commit from 2.7.2-dev.1

robbkidd (approved):
👍🏻 

I mixed this branch with the "[run refinery locally with a TLS'd redis](https://github.com/honeycombio/refinery/pull/1291)" branch. The refinery in there (based on `main`) went from being unhappy to being happy.

Yes.

PR #1285, opened 2024-08-16
fix: support TLS connections to Redis

VinozzZ:

Currently, we are not setting the tls config for go-redis even through we do have `UseTLS` configuration option.

- log the pubsub publish error
- configure redis client tls


kentquirk (approved):
This has been tested and shown to work.

robbkidd (approved):
Tested in an isolated docker-compose setup with `honeycombio/refinery:2.7.2-dev.1` Refinery image and a Redis running with TLS enabled. 

encrypted

PR #1287, opened 2024-08-18
maint: use non-forked cuckoofilter again

kentquirk:

The cuckoofilter library has been updated with similar changes to the ones we made in [our fork of it](https://github.com/honeycombio/cuckoofilter) (faster hash, instanced RNG for better parallel performance).

Benchmarks now show it is slightly faster than our implementation, so we're switching back.

- Remove `replace` directive in our go.mod, use latest version of cuckoofilter.


VinozzZ (approved):
💯

PR #1288, opened 2024-08-19
maint: change default receive message size from 5MB to 15MB

### LINKED ISSUE #1131: Discussion: Raising the max receive message size default for GRPC connections?
Had a couple of customers in the last couple weeks getting error messages that their GRPC batched events are larger than the default 5MB uncompressed size.  

```shell
2024-05-07T21:03:47.962Z	error	exporterhelper/queue_sender.go:97	Exporting failed. Dropping data.	{
    "kind": "exporter",
    "data_type": "traces",
    "name": "otlp/traces_refined",
    "error": "not retryable error: Permanent error: rpc error: code = ResourceExhausted desc = grpc: received message after decompression larger than max (5747657 vs. 5000000)",
    "dropped_items": 5323
}
```

This is a tunable setting with the [MaxRecvMsgSize](https://honeycombpollinators.slack.com/archives/C06529PHA0H/p1715124869780659?thread_ts=1715116081.640759&cid=C06529PHA0H) config option, so we've allowed customers to work around the error, but I'm wondering if this default shouldn't be higher.

Shepherd currently has a limit at 15MB according to some issues I've seen raised internally from customers sending very large batches to us directly from the collector. I don't know if we need to match Honeycomb's default here, but maybe we could default to 10MB?

Thought I'd bring this forward for discussion.

kentquirk:

- Fixes #1131 
- The default receive message size is smaller than Honeycomb accepts, which can be confusing.

- Change the default from 5MB to 15MB
- This is *technically* a breaking change but only by the most pedantic definition -- there's no realistic situation in which this actually matters, so we're not going to stress over it.

### DIFF
--- a/config/metadata/configMeta.yaml
+++ b/config/metadata/configMeta.yaml
@@ -1552,7 +1552,7 @@ groups:
       - name: MaxRecvMsgSize
         type: memorysize
         valuetype: memorysize
-        default: 5MB
+        default: 15MB
         reload: false
         firstversion: v2.2
         validations:
--- a/go.mod
+++ b/go.mod
@@ -22,7 +22,7 @@ require (
 	github.com/jonboulle/clockwork v0.4.0
 	github.com/json-iterator/go v1.1.12
 	github.com/klauspost/compress v1.17.9
-	github.com/panmari/cuckoofilter v1.0.3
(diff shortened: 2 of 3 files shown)


kentquirk:
I didn't make this branch from the right starting branch.

PR #1291, opened 2024-08-19
maint: add a docker'd Redis TLS local setup

robbkidd:

- make it easy to run a TLS'd Redis locally in development
- maybe a foundation for smoke-tests later

- add a make target for building the Docker image locally like CI does for releases
- add a docker-compose config in a new `smoke-tests/` directory
    - redis is run with TLS enabled using self-signed certs generated as needed by the gen-certs service container


kentquirk:
I know this is still in draft, but please add a README to the folder explaining how to use it?

smoke-test/config.yaml
```diff
@@ -0,0 +1,33 @@
+General:
+  ConfigurationVersion: 2
+  MinRefineryVersion: v2.0
+  ConfigReloadInterval: 50s
```
VinozzZ, line 4: Just curious on why you wanted to set this to be 50s?

smoke-test/config.yaml
```diff
@@ -0,0 +1,33 @@
...
+  HoneycombAPI: https://api-dogfood.honeycomb.io
+
+Logger:
+  Type: stdout
+  Level: info
+
+LegacyMetrics:
+  Enabled: true
+  Dataset: refinery_metrics
+  APIHost: https://api-dogfood.honeycomb.io
```
VinozzZ, line 16: I'm not sure if we want to send this local refinery's metrics into this dataset. It could be confusing since that's where the production refinery sends its metrics

smoke-test/config.yaml
robbkidd: These configs were copied from the ones currently in the `test/` directory and then I added the TLS settings to them. I have no informed opinion about their contents other than enabling TLS.

VinozzZ (approved):
I tested this on my machine and it works great

smoke-test/config.yaml
robbkidd: These *are* the same EXCEPT that which team and environment receive the telemetry depends on the API key provided. There is no default key in this config, so a fresh clone of the project running this setup won't send anywhere. 

I am happy with any config we put in here.

smoke-test/config.yaml
robbkidd: I removed a few things from this config to have them default to … the defaults.

PR #1292, opened 2024-08-20
maint: Delete unused Dockerfile

TylerHelmuth:

- the repo contains a dockerfile that is not used in any of our CI or `make` commands. It has been the source of confusion around our released image's source.  This Dockerfile was only intended to be used in local `docker build .` commands, but has not outlived its usefulness.

- Deletes the Dockerfile.

If alternatively, we want to keep the dockerfile, I could clean it up to bring it in line with best security practices AND try to make it clearer that it isn't used to build our images.


codeboten (approved):

robbkidd (approved):
I agree with this removal. The shipped release image is built with `ko`. When using an image locally, we should be using the same build process.

Thank you. NEXT!

VinozzZ (approved):

PR #1293, opened 2024-08-20
maint: Add warning about cli flags

TylerHelmuth:

- Refinery has a couple CLI flags that, if used, could expose credetials. This PR adds warnings against such behavior.

- Update the `--help` output to warn against using some flags
- Update config docs to warn against using the command line flag option for some configuration fields.


config/metadata/configMeta.yaml
```diff
@@ -757,7 +757,7 @@ groups:
         envvar: REFINERY_OTEL_METRICS_API_KEY, HONEYCOMB_API_KEY
         commandline: otel-metrics-api-key
         firstversion: v2.0
-        summary: is the API key used to send Honeycomb metrics via OpenTelemetry.
+        summary: is the API key used to send Honeycomb metrics via OpenTelemetry. Setting this value via a command line flag may expose credentials - it is recommended to use the env var or configuration.
```
VinozzZ, line 760: Maybe we can specify through configuration files instead of just configuration?

VinozzZ (approved):

PR #1294, opened 2024-08-21
feat: support configure refinery to use redis in cluster mode

### LINKED ISSUE #1286: Support Redis cluster mode

AWS defaults to cluster mode, so we just need to enable people to use it.

Config option for it. Our redis library supports it with the Universal Client we use, so we need to enable it. It's slightly tricky to get the config right, but we'll figure it out.

VinozzZ:

To support Redis instances in cluster mode, we need to provide a way for users to configure Refinery accordingly.
fixes #1286 

- Add a new config option, `ClusterHosts`, under `RedisPeerManagement` 
- Consolidate `RedisPeerManagementConfig` getters into one
- Add support for `env-delim:","` in `CmdEnv` 
- Add tests


config/metadata/configMeta.yaml
```diff
@@ -987,6 +987,20 @@ groups:
...
+        valuetype: stringarray
+        example: "- localhost:6379"
+        firstversion: v2.8
+        validations:
+          - type: elementType
+            arg: hostport
+        reload: false
+        summary: is a list of host and port pairs for the instances in a Redis Cluster, used for managing peer cluster membership.
+        description: >
+          This configuration enables Refinery to connect to a Redis cluster. Each entry in the list should follow the format `host:port`.
```
kentquirk, line 1001:
```suggestion
          This configuration enables Refinery to connect to a Redis deployment set up in Cluster Mode. 
          Each entry in the list should follow the format `host:port`.
```

kentquirk (commented):
I don't think this is quite enough to work with AWS. We need to use the Cluster client if ClusterHosts is specified, because with AWS there will only be one host.

config/mock.go
```diff
@@ -216,6 +217,12 @@ func (m *MockConfig) GetRedisHost() string {
 	return m.GetRedisHostVal
 }
 
+func (m *MockConfig) GetRedisClusterHosts() []string {
```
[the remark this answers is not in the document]
VinozzZ: I agree! It's not too much changes so I just did it in this PR

config/file_config.go
```diff
@@ -285,6 +285,7 @@ type PeerManagementConfig struct {
 
 type RedisPeerManagementConfig struct {
 	Host           string   `yaml:"Host" cmdenv:"RedisHost"`
+	ClusterHosts   []string `yaml:"ClusterHosts"`
```
robbkidd, line 288: What do you think of this also being a `cmdenv` option so that cluster hosts can be provided in an environment variable?

VinozzZ: added the environment variable support

kentquirk (approved):
I like this better! Thanks for making the changes.

PR #1297, opened 2024-08-22
fix: use trace.DescendantCount for span limit

VinozzZ:

Refinery converts SpanEvent and SpanLink to individual events just like regular spans. We should use DescendantCount to check against the SpanLimit instead of only spans.

- use DescendantCount to check against SpanLimit
- modify test to explicitly check for span limit trace send reason


kentquirk (approved):
Good catch. I thought of this at one point and then forgot to check it.

PR #1298, opened 2024-08-22
fix: Use sha256 digest in image reference

TylerHelmuth:

- Improve security by referencing digest instead of tag

- Updates all used images in circleci config to use sha256 digest.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -65,8 +65,10 @@ commands:
 jobs:
   test:
     docker:
-      - image: cimg/go:1.21
-      - image: redis:6.2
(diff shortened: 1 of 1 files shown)


codeboten (approved):

kentquirk:
I'm being a little cranky about this, but doesn't this essentially prevent us from benefiting from upstream security fixes by making it so we can't automatically get them?

TylerHelmuth:
> I'm being a little cranky about this, but doesn't this essentially prevent us from benefiting from upstream security fixes by making it so we can't automatically get them?

Yup

PR #1299, opened 2024-08-22
fix: escape use input in debug route

TylerHelmuth:

- The debug endpoint doesn't sanitize the user input before using it in the response.

- use `html.EscapeString` to escape the user's input string for use in refinery and the response.


kentquirk (changes requested):
Sadly, you can't put it here, because it might calculate the wrong shard destination for the trace. You need to move the EscapeString call to be around the traceID in the sprintf and not affect the shard calculation.

TylerHelmuth:
@kentquirk updated

kentquirk (approved):
LGTM

Issue #956, opened 2023-12-19
Allow a throughput limit


kentquirk:
**Is your feature request related to a problem? Please describe.**

- Throughput samplers aren't good enough for variable traffic. "What would be great would be a Dynamic mode that's the default and then Throughput mode kicks in only if we are above some threshold for overall volume."

**Describe the solution you'd like**

- Refinery monitors throughput (events per second) as a constant feature
- There's a configuration value that allows setting throughput either at the cluster level or at the individual refinery level
- It maintains a sample rate multiplier that is 1 if the total throughput is below the limit
- If the throughput value goes above the specified limit, then the sample rate multiplier starts increasing
- All sample rates generated by the selected samplers are multiplied by this throughput multiplier to bring sample rates back down to within the limit
- The current value of the limit is added as a `meta.` value.

**Describe alternatives you've considered**

- Creating special sampler variants that do this, which would require creating multiple new sampler types.

**Additional context**

- Partially discussed in #525 
- Also discussed in [Pollinators](https://honeycombpollinators.slack.com/archives/C01BZ3J7RMJ/p1703010013565929)

VinozzZ:
I would like to work on this. To ensure I understand both the targeted problem and the proposed solution, I’m going to reiterate them here.

Problem:
Customers want to maintain a consistent sample rate for both bursty traffic and low-volume traffic, ensuring that each type receives equal representation in the sampled data. However, when an unexpected surge of bursty traffic occurs, customers still need to keep the overall event volume under control. This means the sample rate needs to dynamically adjust in response to unexpected traffic bursts, thereby enforcing a total throughput limit.

Proposed Solution:
We propose introducing a new configuration option, ThroughputLimit, in the DynamicSampler. If this option is set, a sample rate multiplier will be applied based on the specified limit and the current throughput. When the current throughput exceeds the limit, the sample rate will be adjusted using this multiplier to bring the throughput back within the defined limit.

Please let me know if this makes sense or if I have misunderstood any aspects.

kentquirk:
I agree with what you said, except for where it is and what it affects -- the configuration option I'm talking about would apply to Refinery as a whole, and would multiply all calculated sample rates from any sampler configuration.

This means that each Refinery in a cluster needs to calculate its output events per second using a moving average based on the number of spans being sent and gossip that periodically. Like stress level, all the refineries in the cluster should share the same calculated value and use it together.

[...]

PR #1300, opened 2024-08-23
feat: support ThroughputLimit in samplers

VinozzZ:

relates to #956 

- Create a `EMAThroughputCalculator` to calculate cluster throughputs  and publish individual throughput on an interval
- Change `Sampler` interface to separate the sample rate calculation logic and sampling decision logic
- Add a new configuration group `EMAThroughputLimit` in rules config


collect/throughput_calculator.go
```diff
@@ -0,0 +1,188 @@
...
+		return err
+	}
+	c.hostID = peerID
+	c.throughputs = make(map[string]throughputReport)
+	// Subscribe to the throughput topic so we can react to throughput
+	// changes in the cluster.
+	c.Pubsub.Subscribe(context.Background(), stressReliefTopic, c.onThroughputUpdate)
+
+	go func() {
+		ticker := c.Clock.NewTicker(c.intervalLength)
```
VinozzZ, line 65: We should only publish if the throughput is different from the previous calculation

sample/rules.go
```diff
@@ -117,36 +117,47 @@ func (s *RulesBasedSampler) GetSampleRate(trace *types.Trace) (rate uint, keep b
...
-			}
-
-			if keep {
-				s.Metrics.Increment(s.prefix + "num_kept")
-			} else {
-				s.Metrics.Increment(s.prefix + "num_dropped")
 				if rule.Drop {
-					// If we dropped because of an explicit drop rule, then increment that too.
-					s.Metrics.Increment(s.prefix + "num_dropped_by_drop_rule")
+					rate = 0
```
VinozzZ, line 129: We don't have the rule's config in the `MakeSamplingDecision` function. I changed the logic to return `rate` as 0 to signal that it's a drop decision due to the `Drop` config

sample/dynamic.go
```diff
@@ -94,5 +90,22 @@ func (d *DynamicSampler) GetSampleRate(trace *types.Trace) (rate uint, keep bool
 			d.Metrics.Gauge(name, val)
 		}
 	}
-	return rate, shouldKeep, "dynamic", key
+	return rate, "dynamic", key
+}
+
+func (d *DynamicSampler) MakeSamplingDecision(rate uint, trace *types.Trace) bool {
```
kentquirk, line 96: I am pretty happy with the removal of the keep parameter, but I thought I understood that we were not going to add a new function to every sampler. I think MakeSamplingDecision is the same for all samplers, except for the metrics. Why is this better than centralizing that into the call from collect()?

VinozzZ: Unfortunately, the `DeterministicSampler`'s decision making logic is quite different from the rest of the samplers

kentquirk: Ugh. You're right, but deterministic sampling breaks with the throughput throttle anyway, so maybe we should consider treating it differently or exempting it from the throttle.

VinozzZ:
I'm closing this for now as we don't have plans to pursue this feature in the near future

PR #1301, opened 2024-08-27
feat: support layered (multiple) configuration files

### LINKED ISSUE #871: Can we support layered configs?
There are times when it would be valuable to have a base configuration that is then modified by applying one or more detail configurations on top of it. 

If we support this, it still has to integrate with command line options and support hot reloading even if only one of the configurations has changed. 

(Potential implementation note -- have cmdenv generate a config that is applied on top of the existing config, rather than using the current toolset.)

Tagging this for a 2.2 investigation, but it may not make it into 2.2 if at all.

Use cases for this feature:
* Splitting configurations into sensitive and non-sensitive sections
* Splitting configurations into shared (perhaps remote) config and local (machine or region-specific) config
* Having the main configuration and a temporary one used for overriding sample rates, perhaps while debugging during an incident.

kentquirk:

- Support the ability to specify multiple configuration files; this is in support of a future feature to allow remote editing of some of a configuration, while leaving other parts untouched.
- Closes #871 

- Make config/env values into slices
- Make it possible to read/combine configurations
- Add a little documentation


config/cmdenv.go
```diff
@@ -26,8 +26,8 @@ import (
 // that this system uses reflection to establish the relationship between the
 // config struct and the command line options.
 type CmdEnv struct {
-	ConfigLocation        string     `short:"c" long:"config" env:"REFINERY_CONFIG" default:"/etc/refinery/refinery.yaml" description:"config file or URL to load"`
-	RulesLocation         string     `short:"r" long:"rules_config" env:"REFINERY_RULES_CONFIG" default:"/etc/refinery/rules.yaml" description:"config file or URL to load"`
+	ConfigLocations       []string   `short:"c" long:"config" env:"REFINERY_CONFIG" env-delim:"," default:"/etc/refinery/refinery.yaml" description:"config file or URL to load; can be specified more than once"`
+	RulesLocations        []string   `short:"r" long:"rules_config" env:"REFINERY_RULES_CONFIG" default:"/etc/refinery/rules.yaml" description:"config file or URL to load; can be specified more than once"`
```
VinozzZ, line 30: This is missing the delimiter

config/configLoadHelpers.go
```diff
@@ -108,17 +109,83 @@ func load(r io.Reader, format Format, into any) error {
...
+		location := strings.TrimSpace(location)
+		r, format, err := getReaderFor(location)
+		if err != nil {
+			return "", err
+		}
+		defer r.Close()
+		// write the data to the hash as we read it
+		rdr := io.TeeReader(r, h)
+
+		// when working on a struct, load doesn't overwrite destination values
```
VinozzZ, line 130: nit - do you mind to change this comment from a double negative to something like `load only overwrite destination values that are specified in a file`

config/configLoadHelpers.go
```diff
@@ -108,17 +109,83 @@ func load(r io.Reader, format Format, into any) error {
...
+		r, format, err := getReaderFor(location)
+		if err != nil {
+			return err
+		}
+		defer r.Close()
 
-	var userData map[string]any
-	if err := load(r, format, &userData); err != nil {
-		return nil, fmt.Errorf("validateConfig unable to load config %s: %w", location, err)
+		// when working on a map, load *DOES* overwrite destination values that are not mentioned in a file, so we
```
VinozzZ, line 151: Do you mean if the first config file has `{"d": "15s", "t": "test"}` and the second config file has `{"d": "25s"}`.
The `load()` will overwrite the `"t"` value with empty string after loading these two files?

kentquirk: I just rewrote it; what do you think?

config/configLoadHelpers_test.go
```diff
@@ -147,3 +151,133 @@ func Test_ConfigHashMetrics(t *testing.T) {
 		})
 	}
 }
+
+// creates temporary yaml files from the strings passed in and returns a slice of their filenames
+func createTempConfigs(t *testing.T, cfgs ...string) []string {
+	tmpDir, err := os.MkdirTemp("", "")
```
VinozzZ, line 157: we can use `t.TempDir()` here so that we can take advantage of the automatic cleanup after all tests have finished running

kentquirk: Nice, I didn't know about that one!

config/configLoadHelpers.go
VinozzZ: Thank you! I understand it now

config/configLoadHelpers_test.go
kentquirk: I also made this change in the other places we didn't use it. Thanks!

config/configLoadHelpers_test.go
```diff
@@ -147,3 +151,127 @@ func Test_ConfigHashMetrics(t *testing.T) {
...
+
+func Test_loadConfigsIntoMap(t *testing.T) {
+	cm1 := makeYAML("General.ConfigurationVersion", 2, "General.ConfigReloadInterval", Duration(1*time.Second), "Network.ListenAddr", "0.0.0.0:8080")
+	cm2 := makeYAML("General.ConfigReloadInterval", Duration(2*time.Second), "General.DatasetPrefix", "hello")
+	cfgfiles := createTempConfigs(t, cm1, cm2)
+
+	cfg := map[string]any{}
+	err := loadConfigsIntoMap(cfg, cfgfiles)
+	require.NoError(t, err)
+	fmt.Println(cfg)
```
VinozzZ, line 221: nit - extra print line

PR #1302, opened 2024-08-29
feat: add IN operator

### LINKED ISSUE #1074: IN operator
Currently the operator supports these options:
`=`, `!=`, `>`, `<`, `>=`, `<=`, `starts-with`, `contains`, `does-not-contain`, `exists`, `not-exists`, `has-root-span`, `matches`

For a high cardinality field, e.g., service name, there's no way we can specify a rule not to sample traces of a list of services. This requires an `in` operator.

kentquirk:

- Fixes #1074 
- Adds `In` and `NotIn` operators to Refinery rules, and allows the Value field to become a list of values.

- Change validation so that we can specify multiple Values
- Implement In and NotIn operators
- Write tests
- Document

### DIFF
--- a/config/validate.go
+++ b/config/validate.go
@@ -99,6 +99,24 @@ func validateDatatype(k string, v any, typ string) string {
 		default:
 			return fmt.Sprintf("field %s must be a string, int, float, or bool", k)
 		}
+	case "sliceorscalar":
+		switch vt := v.(type) {
+		case string, int, int64, float64, bool:
+			// we're good
+		case []any:
+			// we need to check that the slice is all the same type
+			// if it's empty or 1 element, it's fine
+			if len(v.([]any)) > 1 {
+				firstType := fmt.Sprintf("%T", vt[0])
+				for i, a := range vt {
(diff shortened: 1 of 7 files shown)


config/validate.go
```diff
@@ -99,6 +99,24 @@ func validateDatatype(k string, v any, typ string) string {
...
 			return fmt.Sprintf("field %s must be a string, int, float, or bool", k)
 		}
+	case "sliceorscalar":
+		switch vt := v.(type) {
+		case string, int, int64, float64, bool:
+			// we're good
+		case []any:
+			// we need to check that the slice is all the same type
+			// if it's empty or 1 element, it's fine
+			if len(v.([]any)) > 1 {
```
VinozzZ, line 109: Should this be `len(v.([]any)) > 0` if  it's ok to only have 1 element?

kentquirk: No, if it's only 1 element that means all elements are the same type. We only need to check once there's more than 1.

VinozzZ (approved):

PR #1308, opened 2024-09-03
maint: Add some extra logging to pubsub systems

### LINKED ISSUE #1277: Add a lot more diagnostics during startup before Refinery is running.

If Refinery is waking up and can't reach Redis or its peers, it doesn't log much at all (or at least not anything useful). 

We should be a lot more proactive during startup about logging accurate guesses for the reasons for issues encountered, and not simply pass errors up the stack that get logged at the top level and often have very little to do with the actual problem.

We can probably also improve some documentation around configuration if you're not using Kubernetes.

kentquirk:

- Add a bit more logging to redis stuff to make it easier to diagnose
- Closes #1277 

- Add some logging in pubsub
- Delete an unused function

### DIFF
--- a/internal/peer/mock.go
+++ b/internal/peer/mock.go
@@ -25,5 +25,3 @@ func (p *MockPeers) Start() error {
 	}
 	return nil
 }
-
-func (p *MockPeers) stop() {}
--- a/internal/peer/pubsub_redis.go
+++ b/internal/peer/pubsub_redis.go
@@ -134,6 +134,7 @@ func (p *RedisPubsubPeers) Start() error {
 
 	p.peers = generics.NewSetWithTTL[string](PeerEntryTimeout)
 	p.callbacks = make([]func(), 0)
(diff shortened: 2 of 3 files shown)


cartermp (commented):
Is there a way to also dump (maybe with a flag?) the config and rules config to stdout on startup? This would help field teams a bunch.

kentquirk:
We have the ability to query the config through an endpoint if you specify the debug flag. 

What you ask for is possible, but how/why is it different than reading the config files directly? Those are also specified on the command line.

cartermp:
In a chat with @irvingpop he mentioned that it's often a struggle to get the actual config/rules used in a given refinery run, and there's more hopping back and forth with "wait, this doesn't match what you're saying, are you sure this is right?" than they'd like. @irvingpop is that a fair assessment?

VinozzZ (commented):
LGTM. The test failed in CI is due to missing `logger` initialization in `pubsub` test data

VinozzZ (approved):

PR #1320, opened 2024-09-11
feat: Allow proxying of OTLP/gRPC metrics requests

MikeGoldsmith:

Enables proxying of OTLP metric requests over gRPC. This is useful when users want to control egress.

*Note*: OTLP metrics requests over HTTP are already proxied using a generic handler.

This is an updated version of the following PR:
- #609

- add simple OTLP metrics server implementation
- register server during app start up and create proxy connection
- dispose of connection during app shutdown

### DIFF
--- a/route/otlp_metrics.go
+++ b/route/otlp_metrics.go
@@ -1,0 +1,34 @@
+package route
+
+import (
+	"context"
+
+	collectormetric "go.opentelemetry.io/proto/otlp/collector/metrics/v1"
+	"google.golang.org/grpc/metadata"
+)
+
+type MetricServer struct {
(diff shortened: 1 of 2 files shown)


VinozzZ (commented):
The change looks good to me. Do you know how we can test this locally to make sure the metrics are forwarded as expected?

MikeGoldsmith:
> The change looks good to me. Do you know how we can test this locally to make sure the metrics are forwarded as expected?

Yeah, I will work on adding some tests 👍🏻

MikeGoldsmith:
We're gonna park this until we get more feedback we need it.

PR #1322, opened 2024-09-12
feat: Improve log messages to be more informative

### LINKED ISSUE #1321: Make log messages more self-explanatory
Quote from internal slack with a request to improve wording and documentation of some things:

A couple of requests where users see things and get concerned or don't get concerned when they should.

* Channel buffer full, not adding span => Dropping it on the floor and it's never going anywhere
* Evicting large trace ... => Evicting means sending to honeycomb

Not adding and evicting could mean anything in these contexts. Are there retries. Do evicted spans go in the trash or get decided or go to honeycomb?

Basic option, add these log lines to a list of log lines that mean stuff in Docs

Superior option, clarify the log with "Data loss" when there is data loss or "Delivering early" when there's no data loss.  These are 2 examples but many other logs mean very harmless or very bad things and aren't necessarily clear to operators who need to run it but didn't write it.

MikeGoldsmith:

Improve some log messages to be more accurate about what's actually happening.

- closes #1321

- Improve log messages
- Prefer using WithField / WithString to include parameters instead of inline formatting
- Update memory overrun messages from error -> warn - they do not constitute data lost so error feels heavy handed

### DIFF
--- a/collect/collect.go
+++ b/collect/collect.go
@@ -30,7 +30,7 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-var ErrWouldBlock = errors.New("not adding span, channel buffer is full")
+var ErrWouldBlock = errors.New("not adding span, channel buffer is full. Span will not be processed and will be lost.")
 var CollectorHealthKey = "collector"
 
 type Collector interface {
@@ -270,7 +270,7 @@ func (i *InMemCollector) checkAlloc() {
 		WithField("num_traces_sent", len(tracesSent)).
 		WithField("datasize_sent", totalDataSizeSent).
 		WithField("new_trace_count", i.cache.GetCacheCapacity()).
-		Logf("evicting large traces early due to memory overage")
+		Logf("evicting large traces early due to memory overage. Evicted traces will be processed and if selected by sampling, sent to Honeycomb.")
 
 	// Manually GC here - without this we can easily end up evicting more than we
 	// need to, since total alloc won't be updated until after a GC pass.
@@ -796,7 +796,7 @@ func (i *InMemCollector) send(trace *types.Trace, sendReason string) {
 	// if we're supposed to drop this trace, and dry run mode is not enabled, then we're done.
 	if !shouldSend && !i.Config.GetIsDryRun() {
 		i.Metrics.Increment("trace_send_dropped")
-		i.Logger.Info().WithFields(logFields).Logf("Dropping trace because of sampling")
+		i.Logger.Info().WithFields(logFields).Logf("Dropping trace because of sampling decision")
 		return
 	}
 	i.Metrics.Increment("trace_send_kept")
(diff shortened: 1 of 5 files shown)


collect/collect.go
```diff
@@ -30,7 +30,7 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-var ErrWouldBlock = errors.New("not adding span, channel buffer is full")
+var ErrWouldBlock = errors.New("not adding span, channel buffer is full. Span will not be processed and will be lost.")
```
VinozzZ, line 33: I wonder if it would be easier to read this error message if the impact is stated first? Something like `Span lost. Not adding span, channel buffer is full`

collect/collect.go
```diff
@@ -270,7 +270,7 @@ func (i *InMemCollector) checkAlloc() {
 		WithField("num_traces_sent", len(tracesSent)).
 		WithField("datasize_sent", totalDataSizeSent).
 		WithField("new_trace_count", i.cache.GetCacheCapacity()).
-		Logf("evicting large traces early due to memory overage")
+		Logf("evicting large traces early due to memory overage. Evicted traces will be processed and if selected by sampling, sent to Honeycomb.")
```
VinozzZ, line 273: Same as this one. Something like `Making trace decision for evicted large traces early due to memory overage`

VinozzZ (approved):
Looks good to me. Just have two suggestions

collect/collect.go
kentquirk:
```suggestion
		Logf("Making some trace decisions early due to memory overrun.")
```

metrics/otel_metrics.go
```diff
@@ -189,7 +189,7 @@ func (o *OTelMetrics) Register(name string, metricType string) {
 	case "counter":
 		ctr, err := o.meter.Int64Counter(name)
 		if err != nil {
-			o.Logger.Error().WithString("msg", "failed to create counter").WithString("name", name)
```
kentquirk, line 192: Ooohhh, I'm not sure about this change. Having a field consistently called "msg" is a pattern that I at least have used repeatedly. Our log queries all use "msg" as a quick way to see what's going on in the log.

I don't entirely hate giving it a different name, but I do think we shouldn't eliminate the idea of a common field that all the logs have.

MikeGoldsmith: I updated these because I don't believe the log output at all because it doesn't end with a `Logf("..")`. The 
`WithField` and `WithString` func return an `Entry` struct so it can build chained with multiple things, but does nothing until you finish with a `LogF`.

kentquirk: Oh, wow, yeah, I've been burned by that before! OK, I still think that it's useful to have a low(ish)-cardinality "msg" field in our logs so you can group by msg and see spikes.

MikeGoldsmith: After further discussion 1:1 - the body of the message will be low cardinality so it's easy to see the volume of a specific log message. Log attributes set using `WithField`, `WithFields` and `WithString` can contain higher cardinality values.

collect/collect.go
```diff
@@ -270,7 +270,7 @@ func (i *InMemCollector) checkAlloc() {
 		WithField("num_traces_sent", len(tracesSent)).
 		WithField("datasize_sent", totalDataSizeSent).
 		WithField("new_trace_count", i.cache.GetCacheCapacity()).
-		Logf("evicting large traces early due to memory overage")
+		Logf("Making some trace decisions early due to memory overrun.")
```
mterhar, line 273: A few lines up this is deemed an `error` severity message. Since it's not dropping data, does it make sense to downgrade it to `warning`?

mterhar (commented):
Are we mostly consistent about error logs come out every time data is dropped? And then a retry or early decision with some delivery time or sampling impact is a warning?

collect/collect.go
kentquirk: Good catch! I agree.

MikeGoldsmith: Fixed in [c50a4ca](https://github.com/honeycombio/refinery/pull/1322/commits/c50a4ca093fa8937ec59ae810d34a29c151b1fc0)

Issue #1222, opened 2024-07-09
Ability to drop stress recalculation logs


cartermp:
From a customer:

"We're storing them only in our test environment so we can debug issues while we onboard our services onto honeycomb+refinery. It saves us from having to re-deploy refinery every time we want to onboard a new service"

They would like functionality to remove the stress recalc logs as these are causing the most bloat in their instance and making them run above their limits.

kentquirk:
This is already debug-level logging; most likely the best strategy is to only use info except for a few hours when actively debugging a problem.

PR #1328, opened 2024-09-13
feat: Update Honeycomb logger to use EMAThroughput sampler

MikeGoldsmith:

Updates the Honeycomb logger to use the [EMAThroughput](https://github.com/honeycombio/dynsampler-go/blob/main/emathroughput.go#L77) sampler. This sampler has built-in support for burst protection that the [PerKeyThroughput](https://github.com/honeycombio/dynsampler-go/blob/main/perkeythroughput.go#L17) sampler does not.

Burst protection would be useful when a cluster node does down unexpectedly because the other nodes in the cluster will fail to make peer requests until it is removed from the cluster. This can result in a very high number of "failed to send" log messages in a very small window, faster than what the PerKeyThroughput sampler can adjust the sample rate which results in all log messages being sent. The burst protection from the EMAThroughput sampler would help here as it will schedule an update to sample rates if a high number of events are received very quickly, allowing it to react quicker.

This is a like for like replacement, and such I haven't added any additional configuration options the new sampler supports. We can add more later if desired.

- Replace the Honeycomb logger sampler with the EMAThroughput sampler.

### DIFF
--- a/logger/honeycomb.go
+++ b/logger/honeycomb.go
@@ -58,7 +58,7 @@ func (h *HoneycombLogger) Start() error {
 
 	if loggerConfig.GetSamplerEnabled() {
 		h.sampler = &dynsampler.EMAThroughput{
-			AdjustmentInterval:   10 * time.Second,
+			AdjustmentInterval:   30 * time.Second,
 			GoalThroughputPerSec: loggerConfig.SamplerThroughput,
 			MaxKeys:              1000,
 		}
--- a/logger/honeycomb.go
+++ b/logger/honeycomb.go
@@ -57,10 +57,10 @@ func (h *HoneycombLogger) Start() error {
 	}
 
 	if loggerConfig.GetSamplerEnabled() {
-		h.sampler = &dynsampler.PerKeyThroughput{
-			ClearFrequencyDuration: 10 * time.Second,
-			PerKeyThroughputPerSec: loggerConfig.SamplerThroughput,
-			MaxKeys:                1000,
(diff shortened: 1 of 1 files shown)


cartermp:
Will this also aid https://github.com/honeycombio/refinery/issues/1222?

MikeGoldsmith:
> Will this also aid #1222?

Yes, it probably 👍🏻

logger/honeycomb.go
```diff
@@ -57,10 +57,10 @@ func (h *HoneycombLogger) Start() error {
 	}
 
 	if loggerConfig.GetSamplerEnabled() {
-		h.sampler = &dynsampler.PerKeyThroughput{
-			ClearFrequencyDuration: 10 * time.Second,
-			PerKeyThroughputPerSec: loggerConfig.SamplerThroughput,
-			MaxKeys:                1000,
+		h.sampler = &dynsampler.EMAThroughput{
+			AdjustmentInterval:   10 * time.Second,
```
kentquirk, line 61: Let's lengthen this interval -- maybe to 30s, please?

kentquirk (approved):
Sounds good, but given what we've been learning about EMA, we should extend the interval.

VinozzZ (commented):
Given all the quirks that we recently discovered in EMA samplers, I'm worried that it's going to be harder to reason why a log line doesn't get sampled

MikeGoldsmith:
We already sample refinery log messages; this change is just give burst protection when a sudden increase increase happens. The throughput samplers ensure any unique log message is sent at least once per window.

PR #1329, opened 2024-09-13
fix: assertion for ClusterSizer

VinozzZ:
## Which problem is this PR solving?

- ClusterSize was not properly set for samplers that supports `UseClusterSize`

## Short description of the changes

- implement `ClusterSizer` for rules sampler
- convert a `Sampler` to `ClusterSizer` by going through `any`

PR #1333, opened 2024-09-16
fix: remove unnecessary assertion to any

VinozzZ:

- in #1329 , we introduced the assertion to `any` as the intermediate step to check whether a Sampler is a ClusterSizer. 
- It turns out that this intermediate step is unnecessary.

- remove assertion to `any` in `SetClusterSize`


kentquirk (approved):
I still don't know how we convinced ourselves that this was necessary.

MikeGoldsmith (approved):

Issue #1335, opened 2024-09-17
is_alive and is_ready metrics are always zero


kentquirk:
The Refinery metrics system uses a function called `ConvertNumeric` to convert an `any` value to a number. But it doesn't do booleans, and we call it with a boolean for `is_ready` and `is_alive`. Both of them are always zero because of it.

Add a bool caes to ConvertNumeric to convert true to 1.0 and false to 0.0.

PR #1336, opened 2024-09-17
fix: ConvertNumeric now handles bools

kentquirk:

- #1335 

- Add bool case to ConvertNumeric
- Add test for ConvertNumeric function

### DIFF
--- a/metrics/metrics.go
+++ b/metrics/metrics.go
@@ -75,6 +75,11 @@ func ConvertNumeric(val interface{}) float64 {
 		return n
 	case float32:
 		return float64(n)
+	case bool:
+		if n {
+			return 1
+		}
+		return 0
 	default:
 		return 0
 	}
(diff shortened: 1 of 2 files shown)


metrics/metrics.go
```diff
@@ -75,6 +75,11 @@ func ConvertNumeric(val interface{}) float64 {
...
 	case float32:
 		return float64(n)
+	case bool:
+		if n {
+			return 1
+		}
+		return 0
 	default:
 		return 0
 	}
```
MikeGoldsmith, line 85: nit: we could clean this up a little to not have multiple `return 0`'s by removing the default clause.

```suggestion
	case bool:
		if n {
			return 1
		}
	}
	return 0
```

MikeGoldsmith (approved):
Looks good - one minor suggestion for the switch statement.

metrics/metrics.go
kentquirk: Doh, I could have just accepted your suggestion, but somehow didn't.

Issue #1318, opened 2024-09-10
Forward only key fields to peers


VinozzZ:
We will extend the existing peer span forwarding system to include forwarding a “proxy span” to its decider node. This proxy span will include:

- Key fields necessary for trace decisions (this can be empty if no key fields are present).
- A marker that differentiates proxy spans from normal spans.

Conditions for Forwarding a Proxy Span:
- ForceTraceLocality is set to false.
- A new span arrives, even if it does not contain any key fields.

Ignoring Proxy Spans:
When sending spans to Honeycomb, Refinery should ignore proxy spans by checking the marker to ensure they are not forwarded.

MikeGoldsmith:
Requires the following issue to be resolved first:
- #1325

PR #1338, opened 2024-09-17
feat: extract decision span from full span

VinozzZ:

Extract only necessary information that's needed for trace decision from a full span so that we can forward only the key fields to peers later
part of #1318 

- add a method on `Span` to extract only necessary information into a new types.Event
- add a method to differentiate a full span from a decision span
- store `IsRoot` information on `Span`
- simplify naming for span annotation types


types/event.go
```diff
@@ -187,13 +195,60 @@ type Span struct {
...
+	IsRoot      bool
+}
+
+// IsDecicionSpan returns true if the span is a decision span based on
+// a flag set in the span's metadata.
+func (sp *Span) IsDecisionSpan() bool {
+	if sp.Data == nil {
+		return false
+	}
+	v, ok := sp.Data["meta.refinery.min_span"]
```
MikeGoldsmith, line 207: It looks like we only ever set the value of `meta.refinery.min_span` as a bool, do we need to check and convert from a string?

MikeGoldsmith (changes requested):
I think this mostly makes sense, but would like to see the rename of AnnotationType to SpanType moved into a separate PR to help make this one a little more concise.

MikeGoldsmith (approved):
Looks good, thanks @VinozzZ

Issue #1326, opened 2024-09-12
Add a config option for disabling the new trace distribution strategy


VinozzZ:
We need to allow users to keep trace locality if the new trace distribution feature does not work for certain use cases

MikeGoldsmith:
Not ready yet, need to wait for the following to be complete first:
- #1325 
- #1319 
- #1318

PR #1342, opened 2024-09-18
feat: forward decision span through peer endpoint

VinozzZ:

implements: #1318 #1326 

- Add a new config option `ForceTraceLocality` to turn off trace distribution feature
- forward decision spans when peer membership changes


types/event.go
```diff
@@ -241,8 +241,14 @@ func (sp *Span) GetDataSize() int {
 
 	if sp.IsDecisionSpan() {
 		if v, ok := sp.Data["meta.refinery.span_data_size"]; ok {
-			return v.(int)
+			switch value := v.(type) {
```
MikeGoldsmith, line 244: I'm not sure this is related to the new trace locality option. Is this a bug that's got caught up in this PR?

config/metadata/configMeta.yaml
```diff
@@ -1286,6 +1286,16 @@ groups:
...
 
+      - name: ForceTraceLocality
+        type: bool
+        valuetype: nondefault
+        firstversion: v2.9
+        default: false
+        reload: true
+        summary: controls whether to forward spans that belongs to the same trace to a single Refinery.
+        description: >
+          If `true`, Refinery will forward full spans that belongs to the same trace to a single peer.
```
MikeGoldsmith, line 1297:
```suggestion
        summary: controls whether all spans that belongs to the same trace are sent to a single Refinery for processing.
        description: >
          If `true`, Refinery's will route all spans that belongs to the same trace to a single peer.
```

config/file_config.go
```diff
@@ -306,6 +306,7 @@ type CollectionConfig struct {
 	MaxAlloc              MemorySize `yaml:"MaxAlloc"`
 	DisableRedistribution bool       `yaml:"DisableRedistribution"`
 	ShutdownDelay         Duration   `yaml:"ShutdownDelay" default:"15s"`
+	ForceTraceLocality    bool       `yaml:"ForceTraceLocality"`
```
MikeGoldsmith, line 309: I'm not sure `Force` feels right here, maybe we could name it `EnableTraceLocality` instead? Also, should we provide the default of `false` here too for clarity?

Also, all of our references to this setting have to invert the value (eg !i.Config.GetCollectionConfig().ForceTraceLocality) - we could invert the config option and default true.

eg `DisableTraceLocality` with default true

MikeGoldsmith (commented):
Looks good - left some feedback regarding new config field name.

config/file_config.go
VinozzZ: I like the name `EnableTraceLocality`. However, defaulting to true adds complexity when we access this field due to nil pointer issue. An example use case is the `AddSpanCountToRoot`. We had to do this for backwards compatibility. I would prefer not have to continue the pattern
```
AddSpanCountToRoot     *DefaultTrue `yaml:"AddSpanCountToRoot" default:"true"` // Avoid pointer woe on access, use GetAddSpanCountToRoot() instead.
```

MikeGoldsmith (approved):
Looks good 👍🏻

PR #1346, opened 2024-09-19
maint: rename sent_reason_cache to kept_reason_cache

VinozzZ:

In the `collector`, we use `sendReason` to describe why a trace is moved into the decision-making process. Additionally, we have `sentReason`, which indicates the reason behind a trace receiving a “kept” decision.

To avoid confusion between these terms when sharing this information from the decider node to the rest of the cluster, I propose renaming `sentReasonCache` to `keptReasonCache` to better reflect its purpose.

- rename all `sentReasonCache` to `keptReasonCache`


collect/collect.go
```diff
@@ -480,12 +480,12 @@ func (i *InMemCollector) processSpan(sp *types.Span) {
 	trace := i.cache.Get(sp.TraceID)
 	if trace == nil {
 		// if the trace has already been sent, just pass along the span
-		if sr, sentReason, found := i.sampleTraceCache.CheckSpan(sp); found {
+		if sr, keptReason, found := i.sampleTraceCache.CheckSpan(sp); found {
 			i.Metrics.Increment("trace_sent_cache_hit")
```
kentquirk, line 484: maybe change the name of this metric? And are there other ones?

collect/collect.go
```diff
@@ -620,18 +620,18 @@ func (i *InMemCollector) ProcessSpanImmediately(sp *types.Span) (processed bool,
...
 
 	if i.Config.GetAddRuleReasonToTrace() {
 		var metaReason string
-		if len(sentReason) > 0 {
-			metaReason = fmt.Sprintf("%s - late arriving span", sentReason)
+		if len(keptReason) > 0 {
+			metaReason = fmt.Sprintf("%s - late arriving span", keptReason)
 		} else {
 			metaReason = "late arriving span"
 		}
```
kentquirk, line 637: maybe "meta.refinery.send_reason" also needs a rename? But it will break people's metrics dashboards.

kentquirk (approved):
Generally OK, but maybe there's more? Did you search for "send" as well as "sent"?

collect/collect.go
VinozzZ: The `meta.refinery.send_reason` is recording the reason why a trace is moved into the decision making process. We use `meta.refinery.reason` to record the `keptReason`. We should be ok to keep it as it is

collect/collect.go
VinozzZ: I think this one is ok. This metric is for the `TraceSentCache`, not the `keptReasonCache`. Hopefully with this rename, things will be a bit more clear

VinozzZ:
> Generally OK, but maybe there's more? Did you search for "send" as well as "sent"?

My apology that the PR description wasn't specific enough. I just updated it. This PR is only renaming things that are related to the `SentReasonCache` to `KeptReasonCache`

Issue #1348, opened 2024-09-23
The InMemoryCollector can crash somehow and then refinery's just not working right


kentquirk:
I've seen a customer say that this happened to them.

In my case, I've got some code changes that are crashing InMemoryCollector; the following log message came out:
```
ERRO[0005] IsAlive: subsystem dead due to timeout        subsystem=collector
```

If this happens, the system is dead and won't come back to life. It also will hang forever because it crashed while holding a lock.

I've addressed the shutdown problem with #1347 but we should figure out whether there's a lock issue in collector.

VinozzZ:
Would it be possible that the timeout is actually due to a large amount of span redistribution during peer membership change?
The liveness check is currently set to 3s in the InMemoryCollector. If there's a giant trace or lots of traces need to be redistributed to other peers, the `redistributeTraces` function may stuck doing the work before the `collect()` loop can report to the healthcheck again. Maybe we need to give the health check a longer timeout?

kentquirk:
This is related to #1375 and probably fixed by #1393.

codeboten:
Closing this as probably fixed

PR #1349, opened 2024-09-23
fix: remove InMemoryCollector from liveness check on shutdown

VinozzZ:

related to #1348 

Currently, the collector's timeout for healthcheck is set to 3 seconds. During a shutdown, there could be a large number of traces or spans that need to be forwarded to peers. If this is taking longer than 3 seconds, it will definitely timeout the liveness check. 

Since we are shutting down anyway, just let the ShutdownDelay to do the work and not worry about the liveness.

- Unregister `collector` from Refinery's liveness check

### DIFF
--- a/collect/collect.go
+++ b/collect/collect.go
@@ -848,9 +848,10 @@ func (i *InMemCollector) send(trace *types.Trace, sendReason string) {
 func (i *InMemCollector) Stop() error {
 	i.redistributeTimer.Stop()
 	close(i.done)
-	// signal the health system to not be ready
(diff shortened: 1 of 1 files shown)


kentquirk (approved):
I think this is a good idea, but I will also note that I've seen the collector log message when we weren't shutting down -- so there's a problem somewhere that it can exit while holding the lock, without causing a stack trace.

VinozzZ:
I added my theory in the issue itself: https://github.com/honeycombio/refinery/issues/1348#issuecomment-2368712531

Issue #1152, opened 2024-05-16
Add metrics annotations and generate docs


kentquirk:
**Is your feature request related to a problem? Please describe.**

Refinery has lots of metrics. It's not always obvious what they do, what type they are, or how they differ from other, similar metrics. 

**Describe the solution you'd like**

Refinery's original (legacy) metrics were just simple events, but now we support OTel, and OTel supports extra things like unit, description, datatype. We can and should add those elements to the Register call for metrics so that OTel metrics can include them in the metric attributes.

Also, doing this would allow us to identify these calls and extract them automatically and use that to generate a `metrics.md` document that would describe the metrics refinery emits. Customers would find this helpful.

It could be either:
* A command line flag that causes Refinery to emit a (YAML?) file of all the registration data
* A tool that greps the Refinery source code and extracts the data from the calls

In either case, this data could then be fed to a template that writes out the metrics.md file.

**Describe alternatives you've considered**

Doing this all once, by hand, and then trying to remember to keep it up to date. Given the positive experience we've had with documentation generation for configuration, I think it would be worth extending the concept to metrics.

PR #1350, opened 2024-09-25
maint: Refactor metrics registration to streamline declaration and enable easier documentation generation

VinozzZ:

- part 1 for: #1152 
This PR prepares us to have a consistent metrics registration pattern so that in a later PR we can use https://pkg.go.dev/golang.org/x/tools/go/packages to automatically generate metrics documentation.

- part 2 is implemented in #1351

- Introduce a new struct type `metrics.Metadata` to represent all information needed for a Refinery metric
- Change `metrics.Register` signature to accept `metrics.Metadata` as its argument
- Refactor metrics registration calls in each packages so that we declare metrics we want to register first in a package level variable first
- Set zero value for each OTel metrics during registration
- set Unit and Description for each OTel metrics


kentquirk:
If we're going to use OTel units, they are supposed to conform to [this standard](https://unitsofmeasure.org/ucum).

I'm not sure I care much about things like "trace" and "hash" but anything that is an actual measurable quantity should conform to that.  

I didn't go through the whole thing to identify any problems yet. Happy to help if you want.

VinozzZ:
@kentquirk Oh, i didn’t know there’s a standard for OTel units. If you want a easier way to look through all the units, here’s the generated markdown file with all the metrics https://github.com/honeycombio/refinery/blob/yingrong.metric_doc/metrics.md

kentquirk (approved):
I think this is a very nice cleanup. Thank you.

PR #1351, opened 2024-09-25
feat: generate metrics documentation

VinozzZ:

This PR implements the logic to parse out all metrics defined in the Refinery code base and generates a markdown table for them.

- part 2 of #1152 

- add two new commands to the `convert` tool to generate the metricsMeta.yaml file and a metrics.md file


kentquirk:
I have a style suggestion here -- instead of writing out a .md file, save the data in a []struct, where the struct has fields for Name, Type, Unit, Description. Then write out that entire slice to a .yaml file that we store in the same place as other metadata. Then we can write a quick template and use convert to make a really pretty markdown file, maybe a table instead, and it can have headers and intro text, etc.

kentquirk (approved):
I like it!

VinozzZ:
> I like it!

Awesome! Do you mind to give the part 1 PR another look as well? https://github.com/honeycombio/refinery/pull/1350

PR #1354, opened 2024-09-27
feat(doc): separate table for metrics contains prefix

VinozzZ:

The current metrics metadata generator does not resolve dynamically assigned prefixes during metrics registration. As a result, metrics with dynamic prefixes are not displayed with their full names. To improve clarity, these metrics should be shown in a separate table.

- Identified that certain packages in Refinery—specifically those used for peer and incoming nodes, as well as the sampling package—assign dynamic prefixes to their metrics.
- Added logic to the generator to detect metrics from these packages and categorize them separately.
- Updated the metrics.tmpl template to display metrics with dynamic prefixes in a distinct table, ensuring clear differentiation between metrics with and without full names.


kentquirk (approved):
Another solution would have been to write an "IsComplete" boolean into the data structure -- then the template could have done two loops over the data to generate the two tables.

But I don't think that's important, I just wanted to point it out.

VinozzZ:
> Another solution would have been to write an "IsComplete" boolean into the data structure -- then the template could have done two loops over the data to generate the two tables.
> 
> But I don't think that's important, I just wanted to point it out.

That's a very good point! Thank you for pointing it out.

PR #1358, opened 2024-10-02
feat: Record original user agent for spans and logs

### LINKED ISSUE #1356: Record original user agent in a meta field
Refinery drops the incoming user-agent fields on requests. It's useful to know what was sending the original telemetry, so we should add a metadata field to telemetry that is `meta.refinery.incoming_user_agent`, containing the text of the user agent for http requests, and the equivalent instrumentation library information from OTLP requests.

MikeGoldsmith:

Refinery loses the original user agent of the sending application because it aggregates spans until a trace decision is made. This PR records the original user agent for both HTTP and gRPC requests to the event, batch and OTLP trace/log endpoints in a new metadata field `meta.refinery.incoming_user_agent`.

- Closes #1356

- Update event, batch, OTLP trace/logs endpoints to record original user agent in new field
- Update tests to verify events have the new field set
- Add defer for bad API key test to restore allowed keys so subsequent tests can pass


cartermp:
We'll also want to make an update in shepherd telemetry to capture this like we do the user_agents strings today.

route/route.go
```diff
@@ -491,6 +492,7 @@ func (r *Router) batch(w http.ResponseWriter, req *http.Request) {
 			Data:        bev.Data,
 		}
 
+		addIncomingUserAgentFromRequest(ev, req)
```
VinozzZ, line 495: nit - Sadly, this is going to retrieve the same user-agent from the request header for all events. However, this does make the code cleaner

VinozzZ (approved):
LGTM

route/route.go
MikeGoldsmith: Yeah, you're right. I might refactor to retrieve once and write once per event.

route/otlp_trace_test.go
```diff
@@ -498,6 +505,57 @@ func TestOTLPHandler(t *testing.T) {
...
+					Spans: helperOTLPRequestSpansWithStatus(),
+				}},
+			}},
+		}
+		traceServer := NewTraceServer(router)
+		_, err := traceServer.Export(ctx, req)
+		if err != nil {
+			t.Errorf(`Unexpected error: %s`, err)
+		}
+		assert.Equal(t, 2, len(mockTransmission.Events))
```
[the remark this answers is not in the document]
MikeGoldsmith: The span created in `helperOTLPRequestSpansWithStatus` has a span event, which husky separates into two events.

route/route.go
MikeGoldsmith: Updated in [b901286](https://github.com/honeycombio/refinery/pull/1358/commits/b90128645022242e50cac6f97889fe62f504d5c7)

PR #1359, opened 2024-10-03
feat: Update in-memory trace cache to use LRU instead of ring buffer

### LINKED ISSUE #1339: Revisiting trace cache strategy
Currently, Refinery uses a circular buffer to store inflight traces in-memory, with an ejection policy based on memory consumption and cache capacity. While the circular buffer provides constant-time performance and fixed memory usage, it becomes less effective as traces are distributed more evenly across peers in a Refinery cluster. In this new model, strict memory consumption control is less critical, and instead, memory usage should inform Kubernetes to auto-scale the Refinery cluster.

Two potential strategies for improvements:
- Don't eject traces any more; keep track of a cache fullness metric (number of items in the circular cache) and include that in the stress calculation. When cache gets too close to full, we'll start shedding load.
This also probably means that we need to search more than just 5 items to find an empty slot. (We'll know if an empty slot exists because we have a fullness metric)
- Lose the circular buffer entirely -- it's increasingly irrelevant. One model would be to have two pools of slots -- InUse and Available -- and we move items back and forth between them. Then we always know how much capacity we have, or we can expand the cache size on demand.

MikeGoldsmith:

Updates the InMemCache cache implementation to use uses a [hasicorp/golang-lru](https://github.com/hashicorp/golang-lru) cache to order traces. The cache checks if the capacity is set to the default cache size (10000), and replaces the capacity with `math.MaxInt32` instead.

This means the cache will only contain active traces and consume only the required bytes as expired because old traces are removed instead of lingering in the buffer. The collector regularly checks the Refinery process's memory usage, and will shed old traces if over the configured allocation then manually called GC to clean up used resources.

Because Refinery applies a consistent trace timeout, the LRU enables us to efficiently retrieve the oldest trace when evicting expired traces unlike the ring buffer than required us to loop over the entire buffer when searching for expired traces or removing sent traces.

Below is a comparison of the cache before and after the update

- Closes #1339

- Add new trace cache, backed by an LRU cache
- Update benchmarks to compare current and new cache implementations for various actions (Set, Get, Remove, etc)

---

Before

```
BenchmarkCache_Set/InMemCache-10           	   	1000000000	         0.0000008 ns/op	       0 B/op	       0 allocs/op
BenchmarkCache_Get/InMemCache-10          	    	1000000000	         0.0000004 ns/op	       0 B/op	       0 allocs/op
BenchmarkCache_RemoveTraces/InMemCache-10              	1000000000	         0.0000335 ns/op	       0 B/op	       0 allocs/op
BenchmarkCache_TakeExpiredTraces/InMemCache-10         	   18817	     63078 ns/op	      50 B/op	       1 allocs/op
```

After

```
BenchmarkCache_Set/TraceCache-10           	   	1000000000	         0.0000004 ns/op	       0 B/op	       0 allocs/op
BenchmarkCache_Get/TraceCache-10          	    	1000000000	         0.0000003 ns/op	       0 B/op	       0 allocs/op
BenchmarkCache_RemoveTraces/TraceCache-10              	1000000000	         0.0000015 ns/op	       0 B/op	       0 allocs/op
BenchmarkCache_TakeExpiredTraces/TraceCache-10         	27758473	        45.23 ns/op	      41 B/op	       0 allocs/op
```

**Note** the updated cache is considerable faster removing traces (x10 faster) and removing expired traces (> x1000).


kentquirk:
I haven't even looked at the implementation, but why do we even want an LRU any more? There's no reason to limit the number of traces in the cache; the size of the trace cache is basically a rounding error in the total memory usage.

You weren't in the conversation, but when I talked to Yingrong about this, we talked about simply letting the cache size adjust as necessary. It's the total memory usage that actually matters.

collect/cache/cache.go
```diff
@@ -191,3 +193,84 @@ func (d *DefaultInMemCache) RemoveTraces(toDelete generics.Set[string]) {
...
+func NewTraceCache(metrics metrics.Metrics, logger logger.Logger) *TraceCache {
+	logger.Debug().Logf("Starting TraceCache")
+	defer func() { logger.Debug().Logf("Finished starting TraceCache") }()
+
+	for _, metadata := range traceCacheMetadata {
+		metrics.Register(metadata)
+	}
+
+	// allow the cache to grow really large by using math.MaxInt32 (2147483647) as cache size
+	cache, err := lru.New[string, *types.Trace](math.MaxInt32)
```
VinozzZ, line 222: Ohhhh, this is a very cool idea!

collect/cache/cache.go
```diff
@@ -191,3 +193,84 @@ func (d *DefaultInMemCache) RemoveTraces(toDelete generics.Set[string]) {
...
+	return trace
+}
+
+func (c *TraceCache) GetAll() []*types.Trace {
+	return c.cache.Values()
+}
+
+func (c *TraceCache) Set(trace *types.Trace) *types.Trace {
+	c.cache.Add(trace.TraceID, trace)
+	return nil
```
VinozzZ, line 250: It's probably not likely to happen but maybe we can add a log line or a metric here as a signal for us just in case the LRU cache is full

collect/cache/cache.go
```diff
@@ -191,3 +193,84 @@ func (d *DefaultInMemCache) RemoveTraces(toDelete generics.Set[string]) {
...
+
+	return &TraceCache{
+		Metrics: metrics,
+		Logger:  logger,
+		cache:   cache,
+	}
+}
+
+func (c *TraceCache) GetCacheCapacity() int {
+	return math.MaxInt32
```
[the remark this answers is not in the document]
MikeGoldsmith: Yes, the cache capacity value was used to determine how full the cache was when using a ring buffer - ie how many free slots were available. The cache capacity was then used to apply other actions.

This cache implementation doesn't have a maximum capacity, instead it will allow the cache to grow and relies on the collect's memory usage check to shed load if neccesary.

collect/cache/cache.go
```diff
@@ -67,127 +68,94 @@ func NewInMemCache(
 		met.Register(metadata)
 	}
 
-	if capacity == 0 {
-		capacity = DefaultInMemCacheCapacity
+	// if using the default capacity, allow the cache to grow really large by using math.MaxInt32 (2147483647)
+	if capacity == DefaultInMemCacheCapacity {
```
VinozzZ, line 72: I'm not sure if we still want to allow customers to configure CacheCapacity anymore. If we do, we need to also support the trace ejection case from the cache

collect/collect_test.go
```diff
@@ -194,7 +194,7 @@ func TestOriginalSampleRateIsNotedInMetaField(t *testing.T) {
 		GetTracesConfigVal: config.TracesConfig{
 			SendTicker:   config.Duration(2 * time.Millisecond),
 			SendDelay:    config.Duration(1 * time.Millisecond),
-			TraceTimeout: config.Duration(60 * time.Second),
+			TraceTimeout: config.Duration(1 * time.Second),
```
VinozzZ, line 197: Why do we change this to 1 second?

VinozzZ (changes requested):

collect/cache/cache.go
MikeGoldsmith: Yeah, true. I was trying to think what additional effects it would have as incoming and peer queues are based on this config value. I think you're right though, the cache doesn't be user configurable.

collect/collect_test.go
MikeGoldsmith: The non-sampled event is added and we only waits 5 seconds for it to appear in the transmission queue. The cache will only expire and send the events once the trace timeout is reached.

I'm not sure how / why it currently works in main.

collect/cache/cache.go
MikeGoldsmith: Updated in https://github.com/honeycombio/refinery/pull/1359/commits/06023d02593395c1fa96f92e6865c0cb2a2802aa.

collect/collect_test.go
VinozzZ: Oh, it's because the cache size is set to `3` in the `cache.NewInMemCache()`. Once the cache is full, traces will be ejected from the cache when a new span comes in and put into the transmission queue.

VinozzZ (approved):

PR #1362, opened 2024-10-03
feat: publish trace decisions through redis pubsub

VinozzZ:

When a sampling decision is made, the decider node needs to publish the decision alongside all metadata related to that decision to all peers.

- introduce a new `TraceDecision` type to capture all information needed
- separate `redistributeNotifier` into its own file to reduce the length of the `collect.go` file
- break up `send()` into two functions `makeDecision` and `send`
- create a new `trace_decision` pubsub topic for transferring trace decisions 
- trace decision messages are multiplexed on the same pubsub topic through prefix `drop` and `kept`
- refactor `MockSharder` so that we can test trace ownership easier


collect/collect.go
```diff
@@ -671,7 +718,49 @@ func (i *InMemCollector) dealWithSentTrace(ctx context.Context, tr cache.TraceSe
 	// if we receive a proxy span after a trace decision has been made,
 	// we should just broadcast the decision again
 	if sp.IsDecisionSpan() {
-		// TODO: broadcast the decision again
+		var (
+			msg string
+			err error
+		)
+		if tr.Kept() {
+			// TODO: late span in this case won't get HasRoot
```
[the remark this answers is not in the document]
VinozzZ: The decision is still honored. The late span just won't be decorated with metadata like span count and the it won't increase `trace_send_got_root` metric

collect/collect.go
```diff
@@ -1119,106 +1191,146 @@ func (i *InMemCollector) createDecisionSpan(sp *types.Span, trace *types.Trace,
...
-	return r
 }
+func (i *InMemCollector) processTraceDecision(msg string) {
+	tds, err := unmarshalTraceDecisionMessage(msg)
+	if err != nil {
+		i.Logger.Error().Logf("Failed to unmarshal trace decision message. %s", err)
+	}
+	toDelete := generics.NewSet[string]()
+	for _, td := range tds {
+		trace := i.cache.Get(td.TraceID)
```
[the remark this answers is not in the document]
VinozzZ: I agree. My intention is to retrieve the trace object from the cache before processing it in the `send()` function. If it doesn't exist in the cache, we don't need to make sampling decision for it

collect/collect.go
```diff
@@ -1119,106 +1191,146 @@ func (i *InMemCollector) createDecisionSpan(sp *types.Span, trace *types.Trace,
...
-	maxAttempts  int
-	maxDelay     time.Duration
-	metrics      metrics.Metrics
-
-	reset     chan struct{}
-	done      chan struct{}
-	triggered chan struct{}
-	once      sync.Once
-}
+		i.sampleTraceCache.Record(trace, td.Kept, td.KeptReason)
```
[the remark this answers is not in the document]
VinozzZ: Yes, I think this cache can be renamed to something like `TraceDecisionCache` to make it more clear

collect/collect.go
```diff
@@ -354,6 +371,12 @@ func (i *InMemCollector) collect() {
 			return
 		case <-i.redistributeTimer.Notify():
 			i.redistributeTraces()
+		case msg, ok := <-i.traceDecisions:
+			if !ok {
+				return
```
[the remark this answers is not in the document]
VinozzZ: I added a comment since this means that refinery is shutting down

collect/collect.go
```diff
@@ -215,7 +226,11 @@ func (i *InMemCollector) reloadConfigs() {
 		// pull the old cache contents into the new cache
 		for j, trace := range i.cache.GetAll() {
 			if j >= imcConfig.CacheCapacity {
```
VinozzZ, line 228: needs to be updated with the new cache implementation

collect/collect.go
```diff
@@ -533,27 +566,47 @@ func (i *InMemCollector) sendExpiredTracesInCache(ctx context.Context, now time.
...
-			duration, reason = i.send(ctx, t, TraceSendGotRoot)
+			td, err := i.makeDecision(t, TraceSendGotRoot)
+			if err != nil {
+				i.Logger.Error().WithFields(map[string]interface{}{
+					"trace_id": t.TraceID,
+				}).Logf("error making decision for trace: %s", err.Error())
+				continue
+			}
+			i.send(ctx, t, td)
+			toDelete.Add(t.TraceID)
```
VinozzZ, line 583: this might be updated if we decide to pass in a filter callback into the `TakeExpiredTraces`

collect/collect.go
```diff
@@ -533,27 +566,47 @@ func (i *InMemCollector) sendExpiredTracesInCache(ctx context.Context, now time.
...
 		totalSpansSent += int64(t.DescendantCount())
 		_, span2 := otelutil.StartSpanWith(ctx, i.Tracer, "sendReadyTrace", "num_spans", int64(t.DescendantCount()))
-		var duration time.Duration
-		var reason string
 		if t.RootSpan != nil {
 			span2.SetAttributes(attribute.String("send_reason", TraceSendGotRoot))
-			duration, reason = i.send(ctx, t, TraceSendGotRoot)
+			td, err := i.makeDecision(t, TraceSendGotRoot)
+			if err != nil {
+				i.Logger.Error().WithFields(map[string]interface{}{
```
VinozzZ, line 577: move this log into `makeDecision`. change to level `warn`

collect/collect.go
```diff
@@ -533,27 +566,47 @@ func (i *InMemCollector) sendExpiredTracesInCache(ctx context.Context, now time.
...
+					i.Logger.Error().WithFields(map[string]interface{}{
+						"trace_id": t.TraceID,
+					}).Logf("error making decision for trace: %s", err.Error())
+					continue
+				}
+				i.send(ctx, t, td)
+				toDelete.Add(t.TraceID)
 			}
 		}
-		span2.SetAttributes(attribute.Int64("get_sample_rate_duration_ms", duration.Milliseconds()), attribute.String("sample_reason", reason))
```
VinozzZ, line 553: `get_sample_rate_duration_ms` as a histogram

collect/collect.go
```diff
@@ -604,7 +657,10 @@ func (i *InMemCollector) processSpan(ctx context.Context, sp *types.Span) {
 		ejectedTrace := i.cache.Set(trace)
 		if ejectedTrace != nil {
 			span.SetAttributes(attribute.String("disposition", "ejected_trace"))
-			i.send(ctx, ejectedTrace, TraceSendEjectedFull)
+			td, err := i.makeDecision(ejectedTrace, TraceSendEjectedFull)
```
VinozzZ, line 660: this will be impacted by the cache implementation

collect/collect.go
```diff
@@ -671,7 +718,49 @@ func (i *InMemCollector) dealWithSentTrace(ctx context.Context, tr cache.TraceSe
...
+		)
+		if tr.Kept() {
+			// TODO: late span in this case won't get HasRoot
+			msg, err = newKeptDecisionMessage(TraceDecision{
+				TraceID:    sp.TraceID,
+				Kept:       tr.Kept(),
+				KeptReason: keptReason,
+				SampleRate: tr.Rate(),
+				Count:      uint32(tr.SpanCount()),
+				EventCount: uint32(tr.SpanEventCount()),
```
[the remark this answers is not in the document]
VinozzZ: We found that the keptTraceCache does store the counts as `uint32`. Why do we return the count as `uint` from the cache??

collect/collect.go
```diff
@@ -833,104 +933,81 @@ func mergeTraceAndSpanSampleRates(sp *types.Span, traceSampleRate uint, dryRunMo
...
-	}
-
-	i.Metrics.Increment(sendReason)
-
-	var sampler sample.Sampler
-	var found bool
-
-	// get sampler key (dataset for legacy keys, environment for new keys)
-	samplerKey, isLegacyKey := trace.GetSamplerKey()
+	// "trace_span_count"
```
VinozzZ, line 950: we can record these metrics in `makeDecision`

collect/collect.go
```diff
@@ -187,6 +193,11 @@ func (i *InMemCollector) Start() error {
 		i.Peers.RegisterUpdatedPeersCallback(i.redistributeTimer.Reset)
 	}
 
+	if !i.Config.GetCollectionConfig().EnableTraceLocality {
+		i.PubSub.Subscribe(context.Background(), traceDecisionTopic, i.signalTraceDecisions)
+		i.traceDecisions = make(chan string, traceDecisionsBufferSize)
```
VinozzZ, line 198: have a metric to monitor the buffer and increase the buffer size to 1_000_000

collect/collect.go
```diff
@@ -1221,103 +1302,141 @@ func (i *InMemCollector) sendTraces() {
...
-		maxAttempts:  5,
-		done:         make(chan struct{}),
-		clock:        clock,
-		logger:       logger,
-		metrics:      met,
-		triggered:    make(chan struct{}),
-		reset:        make(chan struct{}),
+func (i *InMemCollector) signalTraceDecisions(ctx context.Context, msg string) {
+	select {
+	case <-ctx.Done():
```
VinozzZ, line 1307: we should return from ctx.Done

collect/collect.go
```diff
@@ -1221,103 +1302,141 @@ func (i *InMemCollector) sendTraces() {
...
+	case i.traceDecisions <- msg:
+	default:
+		i.Logger.Warn().Logf("trace decision channel is full. Dropping message")
 	}
-
-	return r
 }
+func (i *InMemCollector) processTraceDecision(msg string) {
+	tds, err := unmarshalTraceDecisionMessage(msg)
```
VinozzZ, line 1314: separate pubsub topics for kept decisions and drop decisions

collect/collect.go
```diff
@@ -1221,103 +1302,141 @@ func (i *InMemCollector) sendTraces() {
...
+	i.cache.RemoveTraces(toDelete)
 }
 
-func (r *redistributeNotifier) Reset() {
-	var started bool
-	r.once.Do(func() {
-		go r.run()
-		started = true
+func (i *InMemCollector) makeDecision(trace *types.Trace, sendReason string) (*TraceDecision, error) {
+	if !i.IsMyTrace(trace.ID()) {
```
VinozzZ, line 1339: this check should only happen if `EnableTraceLocality` is set to false

collect/collect.go
```diff
@@ -1221,103 +1302,141 @@ func (i *InMemCollector) sendTraces() {
...
+				"error":    err.Error(),
+			}).Logf("Failed to publish trace decision")
+			return nil, err
 		}
 	}
+
+	return &td, nil
+}
+
+func (i *InMemCollector) IsMyTrace(traceID string) bool {
```
VinozzZ, line 1440: check traceLocality here as part of the `IsMyTrace` logic

collect/trace_decision.go
```diff
@@ -0,0 +1,79 @@
...
+		}
+		td = append(td, keptDecision)
+
+		return td, nil
+	default:
+		return nil, fmt.Errorf("unexpected message prefix for trace decision")
+	}
+}
+
+type TraceDecision struct {
```
VinozzZ, line 60: we can also have separate data struct for kept and dropped decision

collect/redistributer.go
```diff
@@ -0,0 +1,113 @@
...
+	"math/rand/v2"
+	"sync"
+	"time"
+
+	"github.com/honeycombio/refinery/logger"
+	"github.com/honeycombio/refinery/metrics"
+	"github.com/jonboulle/clockwork"
+)
+
+type redistributeNotifier struct {
```
VinozzZ, line 14: remove this change from this PR

app/app_test.go
```diff
@@ -89,25 +89,43 @@ func (w *countingWriterSender) waitForCount(t testing.TB, target int) {
 	}
 }
 
-func defaultConfig(basePort int) *config.MockConfig {
+// defaultConfig returns a config with the given basePort and redisDB.
+// tests in this file are running in parallel, so we need to ensure that
+// each test gets a unique port and redisDB.
+//
+// by default, every Redis instance supports 16 databases.
+func defaultConfig(basePort int, redisDB int) *config.MockConfig {
```
MikeGoldsmith, line 97:
```suggestion
// defaultConfig returns a config with the given basePort and redisDB.
// tests in this file are running in parallel, so we need to ensure that
// each test gets a unique port and redisDB.
//
// by default, every Redis instance supports 16 databases, we use redisDB as a way to separate test data
func defaultConfig(basePort int, redisDB int) *config.MockConfig {
```

MikeGoldsmith (approved):
I think this looks great. I'd love to see it being used in a test deployment.

PR #1364, opened 2024-10-04
fix: Put a limit on the size of sampler keys

### LINKED ISSUE #1363: Protect against giant sample keys
Imagine a large trace with thousands of spans, one for each user, and someone accidentally sets up the key fields for the sampler to include a high-cardinality field like user ID. The value of the sampler key would be an enormous string containing every user ID. 

We think we saw this happen today. We should protect against it by limiting the size of the key, most likely by limiting the number of items that can be placed into the key.

kentquirk:

- If someone sends a big trace where one of the sampler keys is a high-cardinality field, Refinery could generate a huge sampler key value. This causes problems both for refinery but also the downstream telemetry. Let's not do that.

- Put a limit of 100 unique values to make up any one sampler key. Even that is big and probably useless but it should avoid any real use cases. 
- Add a test to show it works.

Fixes #1363

### DIFF
--- a/sample/trace_key.go
+++ b/sample/trace_key.go
@@ -55,16 +55,20 @@ outer:
 	for _, field := range d.fields {
 		for _, span := range spans {
 			if val, ok := span.Data[field]; ok {
-				v := fmt.Sprintf("%v", val)
-				uniques.Add(v)
+				u := fmt.Sprintf("%s/%v", field, val)
+				// don't bother to add it if we've already seen it
+				if uniques.Contains(u) {
+					continue
+				}
+				uniques.Add(u)
 				if len(uniques) >= maxKeyLength {
 					break outer
 				}
-				fieldCollector[field] = append(fieldCollector[field], v)
+				fieldCollector[field] = append(fieldCollector[field], fmt.Sprintf("%v", val))
 			}
 		}
 	}
-	// ok, now we have a map of fields to a list of all values for that field.
+	// ok, now we have a map of fields to a list of all unique values for that field.
 	// (unless it was huge, in which case we have a bunch of them)
 
 	var key string
--- a/sample/trace_key.go
+++ b/sample/trace_key.go
@@ -6,6 +6,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/honeycombio/refinery/generics"
 	"github.com/honeycombio/refinery/types"
 )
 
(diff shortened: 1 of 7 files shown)


sample/trace_key.go
```diff
@@ -42,16 +43,29 @@ func (d *traceKey) build(trace *types.Trace) string {
...
 	spans := trace.GetSpans()
+	uniques := generics.NewSetWithCapacity[string](maxKeyLength)
+outer:
 	for _, field := range d.fields {
 		for _, span := range spans {
 			if val, ok := span.Data[field]; ok {
-				fieldCollector[field] = append(fieldCollector[field], fmt.Sprintf("%v", val))
+				v := fmt.Sprintf("%v", val)
+				uniques.Add(v)
+				if len(uniques) >= maxKeyLength {
```
MikeGoldsmith, line 60: We don't use the uniques set for anything other than counting, could we just use simple integer counter instead?

sample/trace_key.go
```diff
@@ -42,16 +43,29 @@ func (d *traceKey) build(trace *types.Trace) string {
...
+					break outer
```
MikeGoldsmith, line 61: We should log that the key exceeded the configured max so we know it's stopped looking for additional keys.

sample/trace_key.go
```diff
@@ -42,16 +43,29 @@ func (d *traceKey) build(trace *types.Trace) string {
 	// if they happen multiple times.
 	fieldCollector := make(map[string][]string)
 
+	// once a key gets this many unique values, it's off the charts in terms of uniqueness
+	// so we just stop looking for more.
+	// This is a safety valve to prevent us from someone sending a high-cardinality field
+	// in a giant trace.
+	const maxKeyLength = 100
```
MikeGoldsmith, line 50: Probably for the future if the need arises, but this might be a nice configuration option in the future.

MikeGoldsmith (changes requested):
Makes sense in general, but I think a generic set to hold the keys is a bit too much. Can we simplify to an int counter?

sample/trace_key.go
kentquirk: I had done it that way originally, but the problem is that it's OK to have 1000 instances of a low-cardinality field since we only keep one of them. I just noticed, though, that fieldCollector can still get really large in that situation, so I've just added some code that actually checks that we have a new value before we add it, which will prevent fieldCollector from getting big.

sample/trace_key.go
kentquirk: OK; the function here doesn't have access to a logger, so I had it return the key size it's using and log if it is the max value.

It occurs to me that we could now make a metric out of this, but I don't have time to do that. Please feel free if you think that's a good idea.

sample/trace_key.go
MikeGoldsmith: Created a follow-up issue to track adding in the future if we decide to do it.

sample/trace_key.go
MikeGoldsmith: Created a follow-up issue to track adding in the future if we decide to do it.

MikeGoldsmith (approved):

PR #1370, opened 2024-10-09
maint: DO NOT MERGE Add additional telemetry for testing collect timing

TylerHelmuth:

- need additional metrics to investigate a timing issue in collect

- add duration metric to time collect loop
- add caller specification to health check
- Fix is_alive/is_ready metrics


VinozzZ (changes requested):
do not merge

PR #1371, opened 2024-10-09
feat: make collector health check timeout configurable

VinozzZ:

When there's a big volume of traffic, collector can take longer to respond to the health check

- add a new config option `HealthCheckTimeout` in `Collection`

### DIFF
--- a/collect/collect.go
+++ b/collect/collect.go
@@ -387,16 +387,12 @@ func (i *InMemCollector) collect() {
 					return
 				}
 				i.processSpan(sp)
-				i.Metrics.Gauge("collector_collect_loop_duration_ms", float64(time.Now().Sub(startTime).Milliseconds()))
-				continue
 			case sp, ok := <-i.fromPeer:
 				if !ok {
 					// channel's been closed; we should shut down.
 					return
 				}
 				i.processSpan(sp)
-				i.Metrics.Gauge("collector_collect_loop_duration_ms", float64(time.Now().Sub(startTime).Milliseconds()))
-				continue
 			case <-i.reload:
 				i.reloadConfigs()
 			}
--- a/collect/collect.go
+++ b/collect/collect.go
@@ -141,7 +141,7 @@ func (i *InMemCollector) Start() error {
 	// listen for config reloads
 	i.Config.RegisterReloadCallback(i.sendReloadSignal)
 
-	i.Health.Register(CollectorHealthKey, 3*time.Second)
+	i.Health.Register(CollectorHealthKey, time.Duration(imcConfig.HealthCheckTimeout))
 
 	for _, metric := range inMemCollectorMetrics {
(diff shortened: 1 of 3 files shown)


config/metadata/configMeta.yaml
```diff
@@ -1285,6 +1285,16 @@ groups:
           This value should be set to a bit less than the normal timeout period
           for shutting down without forcibly terminating the process.
 
+      - name: HealthCheckTimeout
+        type: duration
+        valuetype: nondefault
+        firstversion: v2.8
+        default: 3s
+        reload: false
+        summary: Controls the maximum duration allowed for health checks to complete.
```
TylerHelmuth, line 1294:
```suggestion
        summary: Controls the maximum duration allowed for collection health checks to complete.
```

config/metadata/configMeta.yaml
```diff
@@ -1285,6 +1285,16 @@ groups:
...
+        description: >
+          The `HealthCheckTimeout` setting specifies the maximum duration allowed for the health checks of registered subsystems to complete. If a subsystem does not respond within this timeout period, it will be marked as unhealthy. This timeout value should be set carefully to ensure that transient delays do not lead to unnecessary failure detection while still allowing for timely identification of actual health issues.
```
TylerHelmuth, line 1296:
```suggestion
          The `HealthCheckTimeout` setting specifies the maximum duration allowed for the health checks of the collection subsystems to complete. If a subsystem does not respond within this timeout period, it will be marked as unhealthy. This timeout value should be set carefully to ensure that transient delays do not lead to unnecessary failure detection while still allowing for timely identification of actual health issues.
```

TylerHelmuth (commented):
Left suggestions to update wording to specify this only applies to the collection health check subsystems.

TylerHelmuth (approved):

collect/collect.go
```diff
@@ -376,18 +378,22 @@ func (i *InMemCollector) collect() {
 					return
 				}
 				i.processSpan(sp)
+				i.Metrics.Gauge("collector_collect_loop_duration_ms", float64(time.Now().Sub(startTime).Milliseconds()))
```
MikeGoldsmith, line 381: I'm not sure I understand why the "collector_collect_loop_duration_ms" metric is being recorded when handling incoming spans, peer spans and at the end of the for outer for loop.

I expected it only happen at the end of the outer for loop.

MikeGoldsmith (commented):
Making the health check timeout configurable is good. I've left a query regarding how the new metric is recorded.

MikeGoldsmith:
I'd also suggest updating the target branch to main, then we can cherry-pick it back into the release branch.

collect/collect.go
VinozzZ: It's because of the `continue` statements. I have removed them since they are unnecessary to have right now

PR #1387, opened 2024-10-15
maint: add comments about an edge case in SendKey

VinozzZ:

When using Beeline with a classic API key, events are sent with dataset configured. However, when Refinery replaces the API key with an E&S key, events are stored using service.name instead of dataset. Refinery cannot control Beeline’s dataset propagation, as this is handled by beeline configuration. This issue does not occur with OpenTelemetry (OTel) SDKs, which have a different propagation mechanism.

- adding a note in README to call out this edge case


README.md
```diff
@@ -155,6 +155,9 @@ A quick start for specific scenarios is below:
 * Set `SendKey` to a valid Honeycomb Key
 * Set `SendKeyMode` to `listedonly`
 
++**NOTE**
++
++When using Beelines with a classic API key to send data to Refinery, ensure that the `SendKey` is also a classic Key, not an environment and service(E&S) key.
```
TylerHelmuth, line 160: Would that work tho or would that header still get dropped? I think if you're sending with a Classic Key/Header combo that you should not use SendKey at all.

VinozzZ: If the `SendKey` is a classic key, it should work since the `dataset` header is present

MikeGoldsmith (approved):

Issue #1369, opened 2024-10-09
Nodes sending peer traffic to localhost (and failing to send peer traffic to localhost)


tdarwin:
One of our prospects using hosted Refineries for POCs increased the volume of telemetry they were sending to us and we found in the investigation of the failover that there was a disproportionate increase of peer traffic, and looking through the error logs, a lot of nodes were failing to send peer traffic to the localhost.  Though it's concerning that the nodes were failing to send to peers, the bigger concern is that we're trying to send peer traffic to localhost at all?

![Refinery Send Event Error Logs 2024-10-09 at 10 53 31 AM](https://github.com/user-attachments/assets/645c14c8-ea45-4b84-9cba-fe2d441f4424)

**Versions**

- Go: Whatever version is used in the 2.8.2 release
- Refinery: 2.8.2

**Steps to reproduce**

Not sure how to reproduce this without a tremendous amount of volume to a larger cluster, but 

**Additional context**

Reach out to me in slack for relevant queries.

PR #1388, opened 2024-10-15
fix: use current node address as default peer list

VinozzZ:

- #1369 

Since `GetPeers()` in redis peer implementation returns `127.0.0.1:8080` when no peer is present, it causes trace forwarding to fail if all peers have unregistered itself from the system.
- use current node address as the default peer list


kentquirk:
OK, but the reason that FilePeers returns one peer is that the old trace distribution code expected "peers" to include self, and everything expected at least one peer. And I think that this assumption is still valid for the partitioning code. So why do you ever expect 0 peers?

If we've changed that assumption, we maybe should instead make FilePeers not do that. I'm worried that there's an inconsistency here.

MikeGoldsmith (approved):

Issue #1375, opened 2024-10-10
Collector loop taking longer than expected


TylerHelmuth:
**Versions**

- Refinery: 2.8.x

**Steps to reproduce**

1. Send a lot of data through refinery
2. Observe that sometimes the msg `IsAlive: subsystem dead due to timeout` is recorded for the `collector` subsystem.

**Additional context**

Ever since we added the collect loop to the health check system in 2.8.0 we now see it sometimes take longer than 3 seconds to complete its work. We added support for configuring this timeout value in 2.8.3, but it would be good to investigate why it is taking a long time and if there are any performance improvements that can be made.

VinozzZ:
One proposed idea is to start a goroutine to asynchronously send all spans in a trace, preventing the collect loop from being blocked.

TylerHelmuth:
Through a lot of tracing of the collect loop we found `GetSampleRate`, specifically for a Rules-Based sampler using the EMA dynamic sampler, to be the culprit.  There is a lock in that sampler to allow it do make the calculations it needs.  If we are trying to make a sampling decision during the lock we end up having to wait.  Process enough traces in a single time through collect and this adds up.

Changing the sampler would be quite difficult.  Instead we'll opt to process less traces each time through the collect loop.

PR #1393, opened 2024-10-21
fix: Allow the number of traces processed during each SendTicker tick to be configurable

TylerHelmuth:

- closes https://github.com/honeycombio/refinery/issues/1375

We found, by tracing the collect loop, that Refinery was sometimes doing A TON of work each `SendTicker` tick depending on if there were a lot of expired traces for which to make a sampling decision and send. Refinery experienced _significant_ performance improvements by setting a max value for how many expired traces would be processed each tick and by moving the actual send into a separate goroutine.

- `send` no longer actually sends all the spans in a trace. Instead, it adds the read-to-send trace to a new channel.  A new goroutine ranges over this channel and does the actual work of preparing the spans and adding them to the outgoing queue. This allows the collect loop to make the sampling decision and get back to work faster, while the actually send off can be handled independently.
- Adds a new configuration option, `Traces.MaxExpiredTraces` that limits how many expired traces will be processed during a `SendTicker` tick. The default value is 5000, which was the value we found kept the collect loop consistently under 3 seconds.
- Fixes existing tests
- Adds a lot of tracing to the collect loop and its functions
- Adds new `collector_outgoing_queue` histogram to track the new queue's size.

### DIFF
--- a/collect/collect.go
+++ b/collect/collect.go
@@ -12,6 +12,7 @@ import (
 	"sync"
 	"time"
 
+	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
 
 	"github.com/honeycombio/refinery/collect/cache"
@@ -340,6 +341,7 @@ func (i *InMemCollector) collect() {
 	defer i.mutex.Unlock()
 
 	for {
+		ctx, span := otelutil.StartSpan(context.Background(), i.Tracer, "collect")
 		startTime := time.Now()
 
 		i.Health.Ready(CollectorHealthKey, true)
@@ -355,56 +357,64 @@ func (i *InMemCollector) collect() {
 		// other.
 		select {
 		case <-i.done:
+			span.End()
 			return
 		case <-i.redistributeTimer.Notify():
-			i.redistributeTraces()
+			i.redistributeTraces(ctx)
 		case sp, ok := <-i.fromPeer:
 			if !ok {
 				// channel's been closed; we should shut down.
+				span.End()
 				return
 			}
-			i.processSpan(sp)
+			i.processSpan(ctx, sp)
 		default:
 			select {
 			case <-i.done:
+				span.End()
 				return
 			case <-ticker.C:
 				select {
 				case <-i.done:
 				default:
-					i.sendExpiredTracesInCache(i.Clock.Now())
+					i.sendExpiredTracesInCache(ctx, i.Clock.Now())
+					_, span2 := i.Tracer.Start(ctx, "checkAlloc")
 					i.checkAlloc()
 
 					// Briefly unlock the cache, to allow test access.
 					i.mutex.Unlock()
 					runtime.Gosched()
 					i.mutex.Lock()
+					span2.End()
 				}
 			case <-i.redistributeTimer.Notify():
-				i.redistributeTraces()
+				i.redistributeTraces(ctx)
 			case sp, ok := <-i.incoming:
 				if !ok {
 					// channel's been closed; we should shut down.
+					span.End()
 					return
 				}
-				i.processSpan(sp)
+				i.processSpan(ctx, sp)
 			case sp, ok := <-i.fromPeer:
 				if !ok {
 					// channel's been closed; we should shut down.
+					span.End()
 					return
 				}
-				i.processSpan(sp)
+				i.processSpan(ctx, sp)
 			case <-i.reload:
 				i.reloadConfigs()
 			}
 		}
 
 		i.Metrics.Histogram("collector_collect_loop_duration_ms", float64(time.Now().Sub(startTime).Milliseconds()))
+		span.End()
 	}
 }
 
-func (i *InMemCollector) redistributeTraces() {
-	_, span := otelutil.StartSpan(context.Background(), i.Tracer, "redistributeTraces")
+func (i *InMemCollector) redistributeTraces(ctx context.Context) {
+	_, span := otelutil.StartSpan(ctx, i.Tracer, "redistributeTraces")
 	redistrubutionStartTime := i.Clock.Now()
 
 	defer func() {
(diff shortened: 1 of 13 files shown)


collect/collect.go
```diff
@@ -810,39 +871,14 @@ func (i *InMemCollector) send(trace *types.Trace, sendReason string) {
...
-		if isDryRun {
-			sp.Data[config.DryRunFieldName] = shouldSend
-		}
-		if i.hostname != "" {
-			sp.Data["meta.refinery.local_hostname"] = i.hostname
-		}
-		mergeTraceAndSpanSampleRates(sp, trace.SampleRate(), isDryRun)
-		i.addAdditionalAttributes(sp)
-		i.Transmission.EnqueueSpan(sp)
+	i.outgoingTraces <- sendableTrace{
```
VinozzZ, line 874: Can we add a metric for the number of items in the `outgoingTraces` queue?

config/file_config.go
```diff
@@ -185,12 +185,13 @@ type RefineryTelemetryConfig struct {
...
-	MaxBatchSize uint     `yaml:"MaxBatchSize" default:"500"`
-	SendTicker   Duration `yaml:"SendTicker" default:"100ms"`
-	SpanLimit    uint     `yaml:"SpanLimit"`
+	SendDelay        Duration `yaml:"SendDelay" default:"2s"`
+	BatchTimeout     Duration `yaml:"BatchTimeout" default:"100ms"`
+	TraceTimeout     Duration `yaml:"TraceTimeout" default:"60s"`
+	MaxBatchSize     uint     `yaml:"MaxBatchSize" default:"500"`
+	SendTicker       Duration `yaml:"SendTicker" default:"100ms"`
+	SpanLimit        uint     `yaml:"SpanLimit"`
+	MaxExpiredTraces uint     `yaml:"MaxExpiredTraces" default:"5000"`
```
VinozzZ, line 194: I wonder if the default should be 0 which is no limit?

TylerHelmuth: I thought about that, but since we know this is causing problems and that 5000 with 100ms SendTicker caps the loop around 3s and that our default Healthcheck ticker is 3s, I figured this will be useful for most users instead of them needing to configure it themselves.

collect/collect.go
TylerHelmuth: Will do

TylerHelmuth: Added `collector_peer_queue` histogram.

VinozzZ: How about naming it `collector_send_queue`? I don't think this is the queue to send to peers

TylerHelmuth: I had a typo, the new metric is named `collector_outgoing_queue`

VinozzZ (approved):

config/metadata/configMeta.yaml
```diff
@@ -399,6 +399,27 @@ groups:
           Decreasing this will check the trace cache for timeouts more
           frequently.
 
+      - name: MaxExpiredTraces
+        type: int
+        valuetype: nondefault
+        default: 5000
+        reload: true
+        validations:
```
TylerHelmuth, line 407:
```suggestion
        firstversion: v2.8.5
        validations:
```

TylerHelmuth, line 407: Does this need a `firstversion` entry for `2.8.5`?

kentquirk (commented):
I might have more but wanted to get this in so far.

VinozzZ:
I'm going to update the PR to be against `main`

collect/collect.go
```diff
@@ -235,6 +248,7 @@ func (i *InMemCollector) checkAlloc() {
 	// To do this, we sort the traces by their CacheImpact value and then remove traces
 	// until the total size is less than the amount to which we want to shrink.
 	allTraces := i.cache.GetAll()
+	cacheSize := len(allTraces)
```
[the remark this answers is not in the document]
VinozzZ: I think here it makes more sense to just take the length of all returned traces since the `Len()` function from the new cache requires grabbing a lock first

config/metadata/configMeta.yaml
VinozzZ: Since we are now including this into 2.9, I updated the `firstVersion` to be `2.9`

VinozzZ:
This PR is ready for review now

collect/collect.go
```diff
@@ -130,6 +140,7 @@ var inMemCollectorMetrics = []metrics.Metadata{
 	{Name: "trace_aggregate_sample_rate", Type: metrics.Histogram, Unit: metrics.Dimensionless, Description: "aggregate sample rate of both kept and dropped traces"},
 	{Name: "collector_redistribute_traces_duration_ms", Type: metrics.Histogram, Unit: metrics.Milliseconds, Description: "duration of redistributing traces to peers"},
 	{Name: "collector_collect_loop_duration_ms", Type: metrics.Histogram, Unit: metrics.Milliseconds, Description: "duration of the collect loop, the primary event processing goroutine"},
+	{Name: "collector_outgoing_queue", Type: metrics.Histogram, Unit: metrics.Dimensionless, Description: "number of traces waiting to be send to upstream"},
```
MikeGoldsmith, line 143:
```suggestion
	{Name: "collector_outgoing_queue", Type: metrics.Histogram, Unit: metrics.Dimensionless, Description: "number of traces waiting to send  upstream"},
```

collect/collect.go
```diff
@@ -285,6 +303,7 @@ func (i *InMemCollector) checkAlloc() {
 	// Manually GC here - without this we can easily end up evicting more than we
 	// need to, since total alloc won't be updated until after a GC pass.
 	runtime.GC()
+	return
```
MikeGoldsmith, line 306: Don't need this extra return
```suggestion
```

collect/collect.go
```diff
@@ -355,56 +375,64 @@ func (i *InMemCollector) collect() {
 		// other.
 		select {
 		case <-i.done:
+			span.End()
```
MikeGoldsmith, line 378: Can we defer `span.End()` instead of manually closing in a number of the select clauses?

collect/collect.go
```diff
@@ -420,27 +448,36 @@ func (i *InMemCollector) redistributeTraces() {
...
 	}
 
 	traces := i.cache.GetAll()
+	span.SetAttributes(attribute.Int("num_traces_to_redistribute", len(traces)))
 	forwardedTraces := generics.NewSetWithCapacity[string](len(traces) / numOfPeers)
 	for _, trace := range traces {
 		if trace == nil {
 			continue
 		}
+		_, span2 := otelutil.StartSpanWith(ctx, i.Tracer, "distributeTrace", "num_spans", trace.DescendantCount())
```
MikeGoldsmith, line 463: Can we rename the span, something like `distributeTraceSpan`?

collect/collect.go
```diff
@@ -481,29 +519,51 @@ func (i *InMemCollector) redistributeTraces() {
...
+
+	span.SetAttributes(attribute.Int("num_traces_to_expire", len(traces)), attribute.Int64("take_expired_traces_duration_ms", dur.Milliseconds()))
+
 	spanLimit := uint32(i.Config.GetTracesConfig().SpanLimit)
+
+	var totalSpansSent int64
+
 	for _, t := range traces {
+		totalSpansSent += int64(t.DescendantCount())
+		_, span2 := otelutil.StartSpanWith(ctx, i.Tracer, "sendReadyTrace", "num_spans", int64(t.DescendantCount()))
```
MikeGoldsmith, line 538: Let's rename this span, something like `sendReadySpan`.

MikeGoldsmith (approved):

collect/collect.go
VinozzZ: This loop is working on each trace returned from the cache. That's why we named it with the `Trace` suffix instead of span

PR #1397, opened 2024-10-23
feat: Try out FIFO queue of trace IDs for Trace Cache

MikeGoldsmith:

After further testing with higher load, we found the LRU cache maintained an unexpected number of items in the cache. This is because while the SentBy time is static and set only once when the first span is received, the LRU would update it's last used timestamp for each interaction such as receiving new spans.

The cache needs to be ordered by the SentBy time otherwise traces are not considered to have expired when intended and so stay in memory for longer than expected.

This PR updates the cache to use a FIFO queue of trace IDs that is ordered on first insertion into the map. The order of the queue aids in ordering traces so we can effectively traverse older traces first.

Note: by design, it's expected that items can be removed from the map of traces (eg redistribute or max alloc event) but the trace ID may still exist in the queue. When iterating the queue, we must check the trace still exists in the cache and ignore it if it's not present in the map.

- replace LRU with map and `[]sting` array that's used as a FIFO of trace IDs
- update cache implementation to maintain both map and queue
- add unit test to verify that if the same trace is set multiple times, it does not get re-added to the queue

Benchmarks after changes:
```
BenchmarkCache_Set-10                  	 6006460	       229.7 ns/op	       2 B/op	       0 allocs/op
BenchmarkCache_Get-10                  	11458786	       130.9 ns/op	       0 B/op	       0 allocs/op
BenchmarkCache_TakeExpiredTraces-10    	 3227960	       429.2 ns/op	     108 B/op	       2 allocs/op
BenchmarkCache_RemoveTraces-10         	17387629	        97.48 ns/op	       8 B/op	       0 allocs/op
```

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,1 +1,1 @@
-golang 1.21.8
+golang 1.23.1
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,1 +1,1 @@
-golang 1.23.1
+golang 1.21.8
--- a/collect/cache/cache.go
+++ b/collect/cache/cache.go
@@ -171,10 +171,12 @@ func (d *DefaultInMemCache) TakeExpiredTraces(now time.Time, max int) []*types.T
 			// trace has expired
 			expired = append(expired, trace)
 
-			// remove the trace from the cache and queue
+			// remove the trace from the cache
 			delete(d.cache, traceID)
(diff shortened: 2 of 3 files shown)


kentquirk:
I don't believe your benchmarks. :)
In 0.0000006 nanoseconds, a photon travels about the width of an atom.

PR #1401, opened 2024-10-25
maint: use a queue for events buffer in MockTransmission

VinozzZ:

Refinery's test suit has been flaky in the CI due to the fact that the CI has way less CPU than our dev machines.

In the past, we have tried to use `Eventually` from the test framework to stabilize our collector tests. Due to CPU constraint in the CI environment and the amount of lock and unlock operations, tests have been failing depends on when a goroutine is picked up by go scheduler.

This PR changes the event buffer in MockTransmission from a mutex protected slice to a channel. This separates the `Enqueue` operation from the `Get` operation so that a writer will not be blocked by a reader

- implement event buffer using a go channel
- update all existing tests to use the new semantic
- check whether we are in shutdown before enqueue a event from the collector to libhoney

### DIFF
--- a/collect/collect_test.go
+++ b/collect/collect_test.go
@@ -418,239 +418,222 @@ func TestAddSpan(t *testing.T) {
 
 }
 
-//// TestDryRunMode tests that all traces are sent, regardless of sampling decision, and that the
-//// sampling decision is marked on each span in the trace
-//func TestDryRunMode(t *testing.T) {
-//	conf := &config.MockConfig{
-//		GetTracesConfigVal: config.TracesConfig{
-//			SendTicker:   config.Duration(20 * time.Millisecond),
-//			SendDelay:    config.Duration(1 * time.Millisecond),
-//			TraceTimeout: config.Duration(60 * time.Second),
-//			MaxBatchSize: 500,
-//		},
-//		GetSamplerTypeVal: &config.DeterministicSamplerConfig{
-//			SampleRate: 10,
-//		},
-//		DryRun:             true,
-//		ParentIdFieldNames: []string{"trace.parent_id", "parentId"},
-//		GetCollectionConfigVal: config.CollectionConfig{
-//			ShutdownDelay: config.Duration(1 * time.Millisecond),
-//		},
-//	}
-//	transmission := &transmit.MockTransmission{}
-//	transmission.Start()
-//	peerTransmission := &transmit.MockTransmission{}
-//	peerTransmission.Start()
-//	coll := newTestCollector(conf, transmission, peerTransmission)
-//
-//	samplerFactory := &sample.SamplerFactory{
-//		Config: conf,
-//		Logger: &logger.NullLogger{},
-//	}
-//	sampler := samplerFactory.GetSamplerImplementationForKey("test", true)
-//	coll.SamplerFactory = samplerFactory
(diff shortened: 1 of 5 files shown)


collect/collect_test.go
```diff
@@ -929,9 +909,9 @@ func TestDependencyInjection(t *testing.T) {
 	}
 }
 
-// TestAddCountsToRoot tests that adding a root span winds up with a trace object in
-// the cache and that that trace gets span count, span event count, span link count, and event count added to it
-// This test also makes sure that AddCountsToRoot overrides the AddSpanCountToRoot config.
+// // TestAddCountsToRoot tests that adding a root span winds up with a trace object in
```
kentquirk, line 912: I think you didn't want these extra slashes

collect/collect_test.go
```diff
@@ -1011,26 +990,23 @@ func TestAddCountsToRoot(t *testing.T) {
...
+	assert.Equal(t, int64(2), events[4].Data["meta.span_event_count"], "root span metadata should be populated with span event count")
+	assert.Equal(t, int64(1), events[4].Data["meta.span_link_count"], "root span metadata should be populated with span link count")
+	assert.Equal(t, int64(5), events[4].Data["meta.event_count"], "root span metadata should be populated with event count")
+	assert.Nil(t, coll.getFromCache(traceID), "after adding a leaf and root span, it should be removed from the cache")
 
 }
 
-// TestLateRootGetsCounts tests that the root span gets decorated with the right counts
-// even if the trace had already been sent
+// // TestLateRootGetsCounts tests that the root span gets decorated with the right counts
```
kentquirk, line 1008: or these

collect/collect_test.go
```diff
@@ -1111,27 +1091,19 @@ func TestLateRootGetsCounts(t *testing.T) {
...
+	assert.Equal(t, int64(1), event[0].Data["meta.span_link_count"], "root span metadata should be populated with span link count")
+	assert.Equal(t, int64(5), event[0].Data["meta.event_count"], "root span metadata should be populated with event count")
+	assert.Equal(t, "deterministic/always - late arriving span", event[0].Data["meta.refinery.reason"], "late spans should have meta.refinery.reason set to rules + late arriving span.")
+	assert.Nil(t, coll.getFromCache(traceID), "after adding a leaf and root span, it should be removed from the cache")
 
 }
 
-// TestAddSpanCount tests that adding a root span winds up with a trace object in
-// the cache and that that trace gets span count added to it
+// // TestAddSpanCount tests that adding a root span winds up with a trace object in
```
kentquirk, line 1105: `s'// //'//'g`

transmit/mock.go
```diff
@@ -1,35 +1,71 @@
...
+	if expectedCount == 0 {
+		maxDelayCount = 3
+	}
+
+	for {
+		select {
+		case ev := <-m.Events:
+			events = append(events, ev)
+		default:
+			if maxDelayCount > 0 {
```
kentquirk, line 48: this is going to happen super fast if there are no events in the channel. Do you want to maybe use a fast ticker instead of a default?

kentquirk (approved):
I think this is great. I asked one question in Slack  - why did you need to add the done channel rather than just closing outgoing traces?

VinozzZ:
> I think this is great. I asked one question in Slack - why did you need to add the done channel rather than just closing outgoing traces?

I have removed the done channel check since we do want the outgoingTraces channel to drain on shutdown

PR #1402, opened 2024-10-29
feat: send drop decisions in batch

VinozzZ:

To reduce network traffic from nodes communicating about trace decisions, drop decisions will be sent in batches. The batching logic is based on either a maximum batch size or a timer. Refinery will send a batch when the batch size limit is reached or when the timer expires, whichever comes first.

- add `sendDropDecisions` goroutine
- add two new config options for the batch size and the ticker interval. This is mostly for easier testing. We may not include these two configs in the final release.

### DIFF
--- a/collect/collect.go
+++ b/collect/collect.go
@@ -155,6 +155,7 @@ var inMemCollectorMetrics = []metrics.Metadata{
 	{Name: "collector_redistribute_traces_duration_ms", Type: metrics.Histogram, Unit: metrics.Milliseconds, Description: "duration of redistributing traces to peers"},
 	{Name: "collector_collect_loop_duration_ms", Type: metrics.Histogram, Unit: metrics.Milliseconds, Description: "duration of the collect loop, the primary event processing goroutine"},
 	{Name: "collector_outgoing_queue", Type: metrics.Histogram, Unit: metrics.Dimensionless, Description: "number of traces waiting to be send to upstream"},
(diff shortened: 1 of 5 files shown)


collect/collect.go
```diff
@@ -1620,3 +1612,72 @@ func (i *InMemCollector) publishTraceDecision(ctx context.Context, td TraceDecis
...
+				eg.Wait()
+				return
+			}
+			// if we get a trace ID, add it to the list
+			traceIDs = append(traceIDs, id)
+			// if we exceeded the max count, we need to send
+			if len(traceIDs) >= i.Config.GetCollectionConfig().MaxDropDecisionBatchSize {
+				send = true
+			}
+		case <-ticker.Chan():
```
kentquirk, line 1647: This gives us a possible situation like this:
* Max size is reached at time T
* We send the batch
* We get one decision
* At T+1ms the ticker fires
* We send that batch

If the number of decisions per second is close to the max batch size, then the above is very likely to happen.

I've thought of two possible ways to deal with this:
* Keep a timestamp of the oldest item in the batch; if it's less than our ticker time, then don't send yet.
* Don't use a ticker, use a timer, and restart the timer after every send, no matter why we sent it.

What do you think?

kentquirk (approved):
I'm happy with this, but left one comment to think about.

collect/collect.go
VinozzZ: I like the timer idea. I will update the code to do that

Issue #1376, opened 2024-10-10
Peer communication failures when using SendKey feature


VinozzZ:
**Versions**

- Refinery: 2.8.2

**Steps to reproduce**

1. configure a `SendKey`
2. configure a list of `ReceiveKeys` that doesn't contain the `SendKey`
3. set `AcceptOnlyListedKeys` to true

Run Refinery and error messages like below will show up
```
time="2024-10-09T08:55:55Z" level=error msg="error when sending event" api_host="http://peer-address:8081/" dataset=teams-api environment="test" error="got unexpected HTTP status 400: Bad Request" 
```

**Additional context**
Adding the send key to the list of accepted receive keys has worked around it

PR #1404, opened 2024-10-30
fix: replace api key with SendKey before transmission

VinozzZ:

Replacing API key with the configured SendKey in the router causes peer communication to fail due to api key validation on both peer and incoming traffic.

To resolve this issue, this PR adjusts the replacement logic to occur immediately before transmission.
- #1376 

- only check if the api key is acceptable in the router
- do replacement in `sendTraces`


config/file_config.go
```diff
@@ -92,20 +92,21 @@ type AccessKeyConfig struct {
...
 
-// CheckAndMaybeReplaceKey checks the given API key against the configuration
-// and possibly replaces it with the configured SendKey, if the settings so indicate.
-// It returns the key to use, or an error if the key is invalid given the settings.
-func (a *AccessKeyConfig) CheckAndMaybeReplaceKey(apiKey string) (string, error) {
-	// Apply AcceptOnlyListedKeys logic BEFORE we consider replacement
-	if a.AcceptOnlyListedKeys && !slices.Contains(a.ReceiveKeys, apiKey) {
-		err := fmt.Errorf("api key %s not found in list of authorized keys", a.sanitize(apiKey))
-		return "", err
+func (a *AccessKeyConfig) IsAccepted(key string) bool {
```
MikeGoldsmith, line 99: I think this could be a little clearer, maybe something like `IsValid` instead? We should add a function description too as it's public function.

config/file_config_test.go
```diff
@@ -79,3 +70,36 @@ func TestAccessKeyConfig_CheckAndMaybeReplaceKey(t *testing.T) {
...
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			a := &AccessKeyConfig{
+				ReceiveKeys:          tt.fields.ReceiveKeys,
+				SendKey:              tt.fields.SendKey,
+				SendKeyMode:          tt.fields.SendKeyMode,
+				AcceptOnlyListedKeys: tt.fields.AcceptOnlyListedKeys,
+			}
+			if got := a.IsAccepted(tt.key); got != tt.want {
```
MikeGoldsmith, line 100: This isn't very clear, let's separate it out a little:

```suggestion
			got := a.IsAccepted(tt.key)
			if got != tt.want {
```

route/middleware.go
```diff
@@ -13,7 +13,7 @@ import (
 
 // for generating request IDs
 func init() {
-	rand.Seed(time.Now().UnixNano())
+	rand.New(rand.NewSource(time.Now().UnixNano()))
```
MikeGoldsmith, line 16: What's the change for?

config/file_config.go
MikeGoldsmith: Also, we create similar errors when this returns false so we could update this func to return the error and check if it's nil when calling it.

MikeGoldsmith (changes requested):
Looks good. Left some suggestions to improve clarity and add return error to IsAccepted

route/middleware.go
kentquirk: Go's stdlib changed the way random numbers worked; the original version of this is no longer functional.

config/file_config.go
VinozzZ: I changed the function signature to return an error which eliminated a lot of duplicated code. Hooray! 

I'm a bit hesitant to name this function `IsValid` since the key could be a very valid api key but it's just not in the accepted key list. I would like to keep the name matching with what the config is doing

config/file_config_test.go
MikeGoldsmith: Yep, we definetly have both styles. It would be nice to consolidate onto one style, but not important to do now.

MikeGoldsmith (approved):

PR #1405, opened 2024-10-31
fix: deal with orphan traces in trace cache

VinozzZ:

If for some reason, a trace that doesn't belong to a peer in the trace cache missed its trace decision from its decider, it will stay in the cache forever.

This PR introduce a clean up mechanism for the scenario above. 

- When a trace has expired more than 2 times of the TraceTimeout, we will forward a proxy span to its decider so that the decider will publish the trace decision again.
- When a trace has expired more than 4 times of the TraceTimeout, we will just make a decision for that partial trace from the current host.

- modify the `filter` function in `sendExpiredTraces` to also return traces that have expired for too long
- remove return value from trace cache interface since its no longer used
- fix trace_cache histogram metrics to pass in the duration as float64
- add test for the new `sendExpiredTraces` behavior

### DIFF
--- a/collect/cache/cache.go
+++ b/collect/cache/cache.go
@@ -16,7 +16,7 @@ import (
 type Cache interface {
 	// Set adds the trace to the cache. If it is kicking out a trace from the cache
 	// that has not yet been sent, it will return that trace. Otherwise returns nil.
-	Set(trace *types.Trace) *types.Trace
+	Set(trace *types.Trace)
 	Get(traceID string) *types.Trace
 
 	// GetAll is used during shutdown to get all in-flight traces to flush them
@@ -90,18 +90,18 @@ func (d *DefaultInMemCache) GetCacheEntryCount() int {
 	return len(d.cache)
 }
 
-func (d *DefaultInMemCache) Set(trace *types.Trace) *types.Trace {
+func (d *DefaultInMemCache) Set(trace *types.Trace) {
 	// we need to dereference the trace ID so skip bad inserts to avoid panic
 	if trace == nil {
-		return nil
+		return
 	}
 	start := time.Now()
 
-	defer d.Metrics.Histogram("trace_cache_set_dur_ms", time.Since(start).Milliseconds())
+	defer d.Metrics.Histogram("trace_cache_set_dur_ms", float64(time.Since(start).Milliseconds()))
 	// update the cache and priority queue
 	d.cache[trace.TraceID] = trace
 	d.pq.Set(trace.TraceID, trace.SendBy)
-	return nil
+	return
 }
 
 func (d *DefaultInMemCache) Get(traceID string) *types.Trace {
@@ -113,7 +113,7 @@ func (d *DefaultInMemCache) Get(traceID string) *types.Trace {
 func (d *DefaultInMemCache) GetAll() []*types.Trace {
 	start := time.Now()
 
-	defer d.Metrics.Histogram("trace_cache_get_all_dur_ms", time.Since(start).Milliseconds())
+	defer d.Metrics.Histogram("trace_cache_get_all_dur_ms", float64(time.Since(start).Milliseconds()))
 	return maps.Values(d.cache)
 }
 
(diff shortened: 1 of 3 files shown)


collect/collect.go
```diff
@@ -291,6 +293,15 @@ func (i *InMemCollector) checkAlloc(ctx context.Context) {
 	tracesSent := generics.NewSet[string]()
 	// Send the traces we can't keep.
 	for _, trace := range allTraces {
+		if !i.IsMyTrace(trace.ID()) {
+			err := errors.New("cannot make a decision for partial traces")
+
+			i.Logger.Warn().WithFields(map[string]interface{}{
```
kentquirk, line 299: Isn't this really common? Do we actually want to be sending thousands of copies of this warning? 

If we keep it, I think the error message could be clearer (what kind of decision are we trying to make?).

And finally, it's a constant string -- why construct an error out of it, and then convert it back to a string?

collect/collect.go
```diff
@@ -560,11 +571,36 @@ func (i *InMemCollector) sendExpiredTracesInCache(ctx context.Context, now time.
...
+		// if a trace has expired more than 4 times the trace timeout, we should just make a decision for it
+		// instead of waiting for the decider node
+		if timeoutDuration > traceTimeout*4 {
+			orphanTraceCount++
+			return true
+		}
+
+		// if a trace has expired more than 2 times the trace timeout, we should forward it to its decider
+		// and wait for the decider to publish the trace decision again
+		if now.Sub(t.SendBy) > traceTimeout*2 {
```
TylerHelmuth, line 592: This condition can also use `timeoutDuration`

collect/collect.go
```diff
@@ -574,32 +610,40 @@ func (i *InMemCollector) sendExpiredTracesInCache(ctx context.Context, now time.
...
+
+	for _, trace := range expiredTraces {
+		// if a trace has expired and it doesn't belong to this peer, we should ask its decider to
+		// publish the trace decision again
+		i.PeerTransmission.EnqueueEvent(i.createDecisionSpan(&types.Span{
+			TraceID: trace.ID(),
+			Event: types.Event{
+				Context: trace.GetSpans()[0].Context,
+			},
+		}, trace, i.Sharder.WhichShard(trace.ID())))
```
TylerHelmuth, line 646: I'm curious what impact this will have on network traffic.

collect/cache/cache.go
```diff
@@ -90,18 +90,18 @@ func (d *DefaultInMemCache) GetCacheEntryCount() int {
...
+func (d *DefaultInMemCache) Set(trace *types.Trace) {
 	// we need to dereference the trace ID so skip bad inserts to avoid panic
 	if trace == nil {
-		return nil
+		return
 	}
 	start := time.Now()
 
-	defer d.Metrics.Histogram("trace_cache_set_dur_ms", time.Since(start).Milliseconds())
+	defer d.Metrics.Histogram("trace_cache_set_dur_ms", float64(time.Since(start).Milliseconds()))
```
kentquirk, line 100: I'm curious why you needed to do this. Milliseconds() returns an integer, and the metrics code converts everything to floats anyway.

TylerHelmuth (approved):

kentquirk (approved):
I'm worried about that warning, but otherwise this looks good.

collect/cache/cache.go
VinozzZ: Hmm, I thought that's the reason why it's reporting 0 value right now. Is it just that fast???
https://ui.honeycomb.io/kibble/environments/kibble/datasets/refinery-metrics/result/gwPNGa96CNB

collect/collect.go
VinozzZ: This should be an edge case and we are essentially just sending a trace id. Hopefully it won't have too big of an impact

collect/collect.go
VinozzZ: I agree with you concern. I changed the log level to debug and the message to be more specific

PR #1411, opened 2024-11-06
feat: only redistribute traces when its ownership has changed

VinozzZ:

To reduce the amount of traffic produced during a redistribution event, this PR changes the logic to only redistribute traces that has its ownership changed

- update `redistributeTraces` to only forward data when a trace's ownership has changed
- make redistribution delay configurable
- add tests

### DIFF
--- a/config/metadata/configMeta.yaml
+++ b/config/metadata/configMeta.yaml
@@ -1306,6 +1306,19 @@ groups:
           By disabling this behavior, it can help to prevent disruptive bursts of network traffic when large traces with long `TraceTimeout`
           are redistributed.
 
+      - name: RedistributionDelay
+        type: duration
+        valuetype: nondefault
+        firstversion: v2.9
+        default: 30s
+        reload: false
+        summary: controls the amount of time Refinery waits after each cluster scaling event before redistributing in-memory traces.
+        description: >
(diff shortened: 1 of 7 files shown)


config/metadata/configMeta.yaml
```diff
@@ -1306,6 +1306,19 @@ groups:
...
+        valuetype: nondefault
+        firstversion: v2.9
+        default: 30s
+        reload: false
+        summary: controls the amount of time Refinery waits after each cluster scaling event before redistributing in-memory traces.
+        description: >
+          This value should be longer than the amount of time between each scaling operations, such as adding or removing a pod in a
+          kubernetes deployment.
+          Each redistrution generates additional traffic between peers. If this value is too short, multiple consecutive redistribution
+          event will occur and may overwhelm the cluster.
```
kentquirk, line 1320:
```suggestion
        description: >
          This value should be longer than the amount of time between individual pod changes in a bulk scaling operation 
          (changing the cluster size by more than one pod).
          Each redistribution generates additional traffic between peers.
          If this value is too short, multiple consecutive redistributions will occur and the resulting traffic
          may overwhelm the cluster.
```

kentquirk (approved):
Looks good. I tweaked the docs a bit.

PR #1416, opened 2024-11-08
fix: count the number of IDs in drop decision messages

MikeGoldsmith:

The counts for drop and kept decisions are based on the number of messages received. However, because drop decisions are actually a set of Trace IDs vs kept which is per trace, the counts do correlate with each other.

For example, the below query shows around 17 drop messages per minute, while the kept decision count is around 2200.

image

This PR updates the drop count to count the number of IDs in the drop messages, which will give a real representation of the number of drop decisions.

- Update the `drop_decisions_received` counter to count the number of IDs in the message


kentquirk:
I feel like we actually want two metrics - one for the number of drop messages, and one for the number of items dropped in response to that. Both are going to be useful to us.

VinozzZ (approved):

PR #1420, opened 2024-11-08
feat: publish instanceID during peer comms

kentquirk:

This is an alternative implementation of #1417, with 2 key differences:
- Added a new generic type, `MapWithTTL` to store the ID and address and used it instead of SetWithTTL. It's about 15% slower than SetWithTTL for equivalent operations, but it solves the problem we have here.
- I constrained the unique ID to 8 characters, hashed from a UUID. 
- I also cleaned up an unneeded typename in main.go that my editor was complaining about.

Closes #1417.

### DIFF
--- a/cmd/refinery/main.go
+++ b/cmd/refinery/main.go
@@ -15,8 +15,10 @@ import (
 	_ "go.uber.org/automaxprocs"
 	"golang.org/x/exp/slices"
 
+	"github.com/dgryski/go-wyhash"
 	"github.com/facebookgo/inject"
 	"github.com/facebookgo/startstop"
+	"github.com/google/uuid"
 	libhoney "github.com/honeycombio/libhoney-go"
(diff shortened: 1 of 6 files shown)


cmd/refinery/main.go
```diff
@@ -85,6 +87,11 @@ func main() {
 		os.Exit(0)
 	}
 
+	// instanceID is a unique identifier for this instance of refinery.
+	// We use a hash of a UUID to get a smaller string.
+	h := wyhash.Hash([]byte(uuid.NewString()), 356783547862)
```
MikeGoldsmith, line 92: Is `356783547862` a seed for the hash? Could we add a comment / more somewhere else to make that clearer?

MikeGoldsmith (approved):
Looks good, just one suggestion to make the hash seed a little clearer.

PR #1430, opened 2024-11-14
feat: compress kept trace decision message

VinozzZ:

Due to the multiplication effect of pubsub messages through Redis, we would like to bring down the amount of network traffic sent by Redis.

One of the biggest contributors is `TraceDecison` message used to communicate kept trace decisions. 

This PR uses `snappy` to compress the data and then encoded using `gob`

Here's the benchmark results comparing to the original JSON encoding.
In the `BenchmarkCompressionSizes` test, I generated a batch of 1000 trace decisions as the input. The amount of decrease in data size is 90%
```
goos: darwin
goarch: arm64
pkg: github.com/honeycombio/refinery/collect
cpu: Apple M2 Max
BenchmarkDynamicJSONEncoding-12          	   3139	   353164 ns/op	 443032 B/op	      3 allocs/op
BenchmarkDynamicCompressedEncoding-12    	   3050	   383788 ns/op	 475333 B/op	     62 allocs/op
BenchmarkDynamicJSONDecoding-12          	    561	  2114997 ns/op	 590167 B/op	   5018 allocs/op
BenchmarkDynamicCompressedDecoding-12    	   3285	   360892 ns/op	 445667 B/op	   5244 allocs/op
BenchmarkCompressionSizes/JSON_Encoding-12         	   5000	   238046 ns/op	 164091 B/op	      2 allocs/op
--- BENCH: BenchmarkCompressionSizes/JSON_Encoding-12
    trace_decision_test.go:305: JSON Encoding: Total Batch: 1, Total Size: 160671 bytes, Average Size: 160671.00 bytes
    trace_decision_test.go:305: JSON Encoding: Total Batch: 100, Total Size: 16227771 bytes, Average Size: 162277.71 bytes
    trace_decision_test.go:305: JSON Encoding: Total Batch: 5000, Total Size: 819582771 bytes, Average Size: 163916.55 bytes
BenchmarkCompressionSizes/Snappy_Compression-12    	   5052	   236998 ns/op	 299547 B/op	     59 allocs/op
--- BENCH: BenchmarkCompressionSizes/Snappy_Compression-12
    trace_decision_test.go:319: Snappy Compression: Total Batch: 1, Total Size: 12115 bytes, Average Size: 12115.00 bytes
    trace_decision_test.go:319: Snappy Compression: Total Batch: 100, Total Size: 1223615 bytes, Average Size: 12236.15 bytes
    trace_decision_test.go:319: Snappy Compression: Total Batch: 5052, Total Size: 62428595 bytes, Average Size: 12357.20 bytes
```

I also used `sync.Pool` for both compression and decompression buffers as well as `snappy.Writer` to reduce allocation


collect/trace_decision.go
```diff
@@ -135,3 +134,56 @@ func (td *TraceDecision) KeptReason() uint {
...
+
+	// Flush snappy writer
+	if err := compr.Close(); err != nil {
+		return nil, err
+	}
+
+	// Copy the buffer’s bytes to avoid reuse issues when returning
+	out := make([]byte, buf.Len())
+	copy(out, buf.Bytes())
+	return out, nil
```
kentquirk, line 170: I think this is the most efficient way to do that:

```suggestion
	return bytes.Clone(buf.Bytes()), nil
```

collect/trace_decision.go
```diff
@@ -46,19 +49,6 @@ func newDroppedDecisionMessage(tds []TraceDecision) (string, error) {
...
-
-	data, err := json.Marshal(tds)
-	if err != nil {
-		return "", err
-	}
-
-	return string(data), nil
-}
-
 func newDroppedTraceDecision(msg string) ([]TraceDecision, error) {
```
kentquirk, line 52: Why not also compress the dropped messages? Even though they're unlikely to do much better than 50% compression (they're essentially hex-coded noise most of the time), that's not nothing given how many we send.

kentquirk (approved):
This looks good! I had a couple of comments, but they're optional.

MikeGoldsmith (approved):
Looks good. I'd also like to see compression for drop traces as it'll help log term.

kentquirk (approved):
Let's try it!

PR #1431, opened 2024-11-15
feat: Add basic telemetry to event, batch and OTLP endpoints

### LINKED ISSUE #1054: Refinery returns GRPC errors, but doesn't log them

Customer was sending batches of data larger than Refinery was configured to accept. While it did return errors to the sender, it didn't log anything. This is because Refinery uses Husky to accept OTLP traffic, and Husky doesn't have a model for callbacks or telemetry today.

* 

- [x] Enable Husky to have an error callback or some other mechanism for returning a log message
- [ ] * Hook it into Refinery so that Refinery can log it (and maybe add a metric for GRPC incoming errors as well)

The _status quo_

**Additional context**

Inspired by a support ticket.

MikeGoldsmith:

We don't currently have any logging or telemetry for our endpoints that receive data (spans & logs). This PR adds tracing  to the following endpoints using OTel instrumentation libraries. It also configures husky to set appropriate validation error attributes.

Endpoints:
- /event
- /batch
- OTel trace (http & OTLP)
- OTel logs (http & OTLP)

Closes the following issue:
- Closes #1054 

- Configure husky's telemetry callback to set validation attributes on existing spans
- Inject the already configured OTel Tracer into the Router
- Add otelhttp & otelgrpc instrumentation libraries and wire up
- Update tests to set a `NoopTracer` in the router

Example HTTP span
image

Example gRPC span
image


cmd/refinery/main.go
```diff
@@ -237,8 +240,13 @@ func main() {
 	if c.GetOTelTracingConfig().Enabled {
 		// let's set up some OTel tracing
 		tracer, shutdown = otelutil.SetupTracing(c.GetOTelTracingConfig(), resourceLib, resourceVer)
-	}
 
+		// add telemtry callback so husky can enrich spans with attributes
```
kentquirk, line 244:
```suggestion
		// add telemetry callback so husky can enrich spans with attributes
```

kentquirk (approved):
Seems useful. Are there any more attributes you want to add to the spans so we know more about them?

PR #1437, opened 2024-11-18
feat: ignore trace decision messages produced by the publishers

VinozzZ:

collect loop iteration is precious to make sure Refinery is able to process incoming data as soon as possible. Trace decision pubsub messages are broadcasted to all peers including the publisher itself. There's no need for the publisher to process those messages since it already knows about those trace decisions.

- prefix each decision messages with the peer ID. I didn't include it into the compression logic. That way we can skip the whole decompression process for a message if we know it's produced by its publisher
- ignore decision messages in the pubsub callback before we ask the collect loop to process them

### DIFF
--- a/collect/trace_decision.go
+++ b/collect/trace_decision.go
@@ -29,34 +29,46 @@ var (
 	dropDecision decisionType = 2
 )
 
-type newDecisionMessage func([]TraceDecision) (string, error)
+type newDecisionMessage func(tds []TraceDecision, senderID string) (string, error)
 
-func newDroppedDecisionMessage(tds []TraceDecision) (string, error) {
+func newDroppedDecisionMessage(tds []TraceDecision, senderID string) (string, error) {
 	if len(tds) == 0 {
 		return "", fmt.Errorf("no dropped trace decisions provided")
 	}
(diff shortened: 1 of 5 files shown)


collect/trace_decision.go
```diff
@@ -66,25 +78,47 @@ func newDroppedTraceDecision(msg string) ([]TraceDecision, error) {
...
+
+	compressed, err := decompressKeptDecisions([]byte(data[1]))
 	if err != nil {
 		return nil, err
 	}
 	return compressed, nil
 }
 
+func isMyDecision(msg string, senderID string) bool {
+	data := strings.SplitN(msg, "|", 2)
```
kentquirk, line 115: It's 4x faster with no allocations if you replace this function with `strings.HasPrefix()`.

Also, the other calls to strings.SplitN(..,..,2) could also be sped up with `strings.IndexRune()` instead. (I just noticed this, we might have other places that would also benefit.)

kentquirk (approved):
It could be faster if you want.

Also, the choice of `|` is a good one, but it would help to include a note that it can't occur in a URL or an IP address.

PR #1440, opened 2024-11-19
feat: Try to drain incoming and peer queues for an amount of time

MikeGoldsmith:

We're seeing that incoming and peer queues can sometimes build up and it takes a long time to recover. A theory is that this is because the collect loop has so many options, we should allow multiple messages to be read from the incoming and peer queues whenever that it's selected to be worked on.

- Add a new utility func that can read messages from a given span channel for up to 100 milliseconds, exiting early if the channel is empty
- Update incoming and peer channels to use new func

### DIFF
--- a/collect/collect.go
+++ b/collect/collect.go
@@ -164,6 +164,8 @@ var inMemCollectorMetrics = []metrics.Metadata{
 	{Name: "kept_decisions_received", Type: metrics.Counter, Unit: metrics.Dimensionless, Description: "total number of kept decisions received"},
 	{Name: "collector_kept_decisions_queue_full", Type: metrics.Counter, Unit: metrics.Dimensionless, Description: "number of times kept trace decision queue is full"},
 	{Name: "collector_drop_decisions_queue_full", Type: metrics.Counter, Unit: metrics.Dimensionless, Description: "number of times drop trace decision queue is full"},
+	{Name: "num_span_drained_from_incoming", Type: metrics.Gauge, Unit: metrics.Dimensionless, Description: "number of spans drained from incoming queue"},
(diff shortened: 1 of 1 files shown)


collect/collect.go
```diff
@@ -417,58 +417,72 @@ func (i *InMemCollector) collect() {
...
 	}
 }
 
+func drainSpanQueue(ctx context.Context, span *types.Span, ch <-chan *types.Span, queueName string, processSpanFunc func(context.Context, *types.Span, string)) {
+	// process the original span
+	processSpanFunc(ctx, span, queueName)
+
+	// let't try to process as many spans as we can in the next 100ms
+	// TODO: make timer configurable?
+	timer := time.NewTimer(time.Millisecond * 100)
```
VinozzZ, line 468: Right now each collect loop iteration takes between 10ms to 20ms. Setting the timer to 100ms means we probably are going to process 5-10 items off of the queue each time. Is that enough?
I wonder if we should drain the queue based on the number of items process instead of a timer to make sure we are always draining X amount of items for each iteration.

I don't know which strategy is better. We can try the timer based first to see how it goes

VinozzZ (approved):
I would love to give this a try!

collect/collect.go
MikeGoldsmith: I think the timer approach is worth trying. If we want it to be more flexible, we could add a temporary config option while we're testing it out.

VinozzZ: Yes, maybe we can add a metric here to see how many items we drain from the queue each time?

kentquirk (approved):
Looks interesting. Let's try it!

PR #1441, opened 2024-11-20
feat: add a limit to queue draining logic

VinozzZ:

The timer may never fire due to Go scheduler starving the timer goroutine.
To prevent the timer being starved, this PR adds a limit to the number of spans being processed during one iteration of the collect loop

- add a limit to the draining logic


kentquirk (approved):
Just like we talked about.

PR #1442, opened 2024-11-20
feat: Rename EnableTraceLocality to DisableTraceLocality

MikeGoldsmith:

To allow users to opt-in safely for disabling trace locality, we're renaming the option to disable trace locality.

- Rename `EnableTraceLocality` to `DisableTraceLocality`


VinozzZ:
Should we just change the config option to `DisableTraceLocality` and then we will be able to leave the default as `false`?

kentquirk:
I just made a bunch of text edits around this change. Please check them out.

VinozzZ (commented):
I made some changes in this PR as well. @MikeGoldsmith do u mind to give a verbal approval and then i can hit the approve button?

MikeGoldsmith:
Thanks @VinozzZ - updates look good

VinozzZ (approved):

PR #1445, opened 2024-11-21
fix: missing read lock in MapWithTTL

VinozzZ:

```
fatal error: concurrent map read and map write

goroutine 193 [running]:
github.com/honeycombio/refinery/generics.(*MapWithTTL[...]).SortedValues(0xdd8a60)
	github.com/honeycombio/refinery/generics/mapttl.go:118 +0xcc
github.com/honeycombio/refinery/internal/peer.(*RedisPubsubPeers).Ready.func1()
	github.com/honeycombio/refinery/internal/peer/pubsub_redis.go:212 +0x454
created by github.com/honeycombio/refinery/internal/peer.(*RedisPubsubPeers).Ready in goroutine 1
	github.com/honeycombio/refinery/internal/peer/pubsub_redis.go:180 +0x84

```

- add `RLock` for `SortedValues`

### DIFF
--- a/generics/mapttl.go
+++ b/generics/mapttl.go
@@ -113,6 +113,9 @@ func (m *MapWithTTL[K, V]) Values() []V {
 // SortedValues returns the values of the map, sorted by key
 func (m *MapWithTTL[K, V]) SortedValues() []V {
 	keys := m.SortedKeys()
+
+	m.mut.RLock()
+	defer m.mut.RUnlock()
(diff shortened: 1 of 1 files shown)


kentquirk (approved):
I forgot that indexing Items needed a lock.

This solution uses two lock cycles (one to get the sorted keys, and a second one you add here). 

I *think* that's preferable to a single lock because the sort takes place without holding the lock. So yeah, let's go with this.

PR #1446, opened 2024-11-22
feat: Add 'unpublished' flag to configs 

kentquirk:

- We have some config options that we'd like to avoid publishing, because we don't know for sure we'll need them, but we want to give ourselves the ability to mess with them while tuning a configuration. By adding the ability to flag a config item as unpublished, it won't go in the main documentation and we can modify it later without breaking changes.

- Add unpublished flag to metadata struct
- Add it to a couple of fields
- Regenerate docs

### DIFF
--- a/tools/convert/minimal_config.yaml
+++ b/tools/convert/minimal_config.yaml
@@ -1,5 +1,5 @@
 # sample uncommented config file containing all possible fields
-# automatically generated on 2024-11-22 at 17:59:52 UTC
+# automatically generated on 2024-11-22 at 18:35:03 UTC
 General:
   ConfigurationVersion: 2
   MinRefineryVersion: "v2.0"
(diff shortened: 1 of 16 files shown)


tools/convert/minimal_config.yaml
```diff
@@ -102,9 +104,14 @@ Collection:
   MaxMemoryPercentage: 75
   MaxAlloc: 0
   DisableRedistribution: false
+  RedistributionDelay: 30s
   ShutdownDelay: 15s
-  EnableTraceLocality: false
+  DisableTraceLocality: false
   HealthCheckTimeout: 3s
+  MaxDropDecisionBatchSize: 1_000
```
VinozzZ, line 111: Should we have these `unpublished` config in the minimal_config.go file?

kentquirk: I couldn't decide, so I left them in. If you think they shouldn't be, I can take them out. I don't mind either way.

VinozzZ: I see. Yes, I would prefer to take them out since we don't really want users to use them yet

kentquirk: That was a good catch. Turns out that the minimal config was also including all the obsolete items, so I fixed that too.

VinozzZ (approved):

PR #1454, opened 2024-12-03
maint: call out cache changes and add missing changelog entry

VinozzZ:

The cache metrics were an important signal for users to monitor their refinery health. We removed it in 2.9 with the new cache implementation.
We should call out this change in the release note.

- update release note
- add missing changelog entry


cartermp (approved):

kentquirk:
I made some additional edits to improve readability and fix some smart quotes that snuck in.

PR #1456, opened 2024-12-09
fix: register OTEL metrics if it's missing

### LINKED ISSUE #1455: Refinery 2.9.0 OTel Metrics not sending `incoming_router_*` events
A customer noticed today that they weren't seeing any `incoming_router_span` events in their Refinery metrics.  Investigating in my sandbox environment, I found that all the incoming metrics were showing up in my dataset that receives the `LegacyMetrics` data, but none of them are showing up in my `OTelMetrics` data.

**Versions**

- Refinery: 2.9.0

**Steps to reproduce**

1. Send OTel Metrics to Honeycomb
  a. Make sure it's going to a separate dataset from any Legacy Metrics
3. Look at events in explore data tab for the OTel Metrics dataset

**Additional context**
image

VinozzZ:

When using OTELMetrics, Refinery previously only emits metrics if they have been registered. This is different from how `LegacyMetrics` works. `LegacyMetrics` will emit data regardless registration during startup. 

This PR changes OTELMetrics to match its behavior with LegacyMetrics so that we don't accidentally create different behaviors when using these two different type of metrics

- fixes #1455 

- create a `init` function for each metrics type
- call init in both `Register` and actual metrics calls
- call register for missing `router` metrics data

### DIFF
--- a/metrics/otel_metrics.go
+++ b/metrics/otel_metrics.go
@@ -190,69 +190,31 @@ func (o *OTelMetrics) Register(metadata Metadata) {
 	o.lock.Lock()
 	defer o.lock.Unlock()
 
-	ctx := context.Background()
-
-	unit := string(metadata.Unit)
 	switch metadata.Type {
 	case Counter:
-		ctr, err := o.meter.Int64Counter(metadata.Name,
-			metric.WithUnit(unit),
-			metric.WithDescription(metadata.Description),
-		)
+		_, err := o.initCounter(metadata)
 		if err != nil {
-			o.Logger.Error().WithString("name", metadata.Name).Logf("failed to create counter")
+			o.Logger.Error().WithString("name", metadata.Name).Logf("failed to create counter. %s", err.Error())
 			return
 		}
-
-		// Give the counter an initial value of 0 so that OTel will send it
-		ctr.Add(ctx, 0)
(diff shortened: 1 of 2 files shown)


metrics/otel_metrics.go
```diff
@@ -263,59 +225,101 @@ func (o *OTelMetrics) Increment(name string) {
...
-		ctr.Add(context.Background(), 1)
-		o.values[name]++
+	var err error
+	ctr, ok := o.counters[name]
+	if !ok {
+		ctr, err = o.initCounter(Metadata{
+			Name: name,
+		})
+
+		if err != nil {
```
kentquirk, line 235: Given that we're ignoring the error anyway, why bother to have initCounter return an error?

kentquirk (commented):
Why this instead of making sure that we register all metrics?

metrics/otel_metrics.go
VinozzZ: I'm returning it for the Register call so that we can know what the error is during registration

VinozzZ:
> Why this instead of making sure that we register all metrics?

I haven't found a solution programmatically to make sure we register all metrics so I thought we can at least match the behavior between the two

kentquirk (approved):

PR #1469, opened 2025-01-15
feat(stress_relief): WIP maximum individual stress level as cluster stress level

VinozzZ:

During a scaling event, newly added refinery has much lower stress level as it ramps up. As a result, the cluster stress level will be pushed out of stress relief mode even though the rest of the cluster  is experiencing significant stress.

I would like to try out a solution by forcing the entire cluster to stay in stress relief mode as long as one node is stressed

-


VinozzZ:
we tried this dev build in our internal cluster. It didn't help with stabilizing the cluster during a scaling up event

PR #1470, opened 2025-01-15
fix: only load config and rules once

TylerHelmuth:
 
We found an issue where refinery, as part of its config load process on startup or reload, would load the config/rules (whether from file or http or any), multiple times. This was happening because we the config data was passed around as a reader and if we ever needed the data a second time (which we do for validation vs unmarshaling into config) we had to go get another reader.

- updates `newFileConfig` to first grab the data of all config locations and the use this data for all validation and unmarshalling. This prevents `newFileConfig` from needing to grab the data multiple times
- updated some unit tests, but most needed unchanged.

### DIFF
--- a/config/configLoadHelpers.go
+++ b/config/configLoadHelpers.go
@@ -1,7 +1,6 @@
 package config
 
 import (
-	"bytes"
 	"crypto/md5"
 	"encoding/hex"
 	"encoding/json"
@@ -58,7 +57,7 @@ func formatFromResponse(resp *http.Response) Format {
 }
 
 // getReaderFor returns an io.ReadCloser for the given URL or filename.
-func getReaderFor(u string) (io.ReadCloser, Format, error) {
+func getReaderFor(u string) ([]byte, Format, error) {
 	if u == "" {
 		return nil, FormatUnknown, fmt.Errorf("empty url")
 	}
@@ -68,7 +67,7 @@ func getReaderFor(u string) (io.ReadCloser, Format, error) {
 	}
 	switch uu.Scheme {
 	case "file", "": // we treat an empty scheme as a filename
-		r, err := os.Open(uu.Path)
+		r, err := os.ReadFile(uu.Path)
 		if err != nil {
 			return nil, FormatUnknown, err
 		}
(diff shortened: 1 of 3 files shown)


kentquirk (commented):
Personally, I think I would have done it differently by keeping the io.ReadCloser abstraction but changing it to a io.ReadSeekerCloser so it could be reset to the beginning to re-read it. I believe this would have been a smaller change, but also, I think what you did here is fine and don't need to force you to change it. 

However, because Reader is a well-known abstraction in Go with a specific meaning, and this is no longer one of those, if you prefer to do it this way, I'd like to change the name used in all these functions. Instead of getReaderFor, maybe getDataBufferFor? or getContentsFor? And similarly for other functions and the new type you created.

Also, there are a couple of wrong comments now. (Like, "we read the file twice").

TylerHelmuth:
I didn't know about `ReadSeekerCloser`, thats exactly what I wanted to use but couldnt find a way to reset a reader.  I'll switch to that

TylerHelmuth:
Reopenning because I would've had to make my own implementation of `io.ReadSeekCloser` for the `resp.Body` case since `bytes.NewReader` is only an `io.Reader`.

config/configLoadHelpers.go
```diff
@@ -58,7 +57,7 @@ func formatFromResponse(resp *http.Response) Format {
 }
 
 // getReaderFor returns an io.ReadCloser for the given URL or filename.
-func getReaderFor(u string) (io.ReadCloser, Format, error) {
+func getReaderFor(u string) ([]byte, Format, error) {
```
kentquirk, line 60: This is also not a reader. A bit more renaming, please?

config/file_config.go
```diff
@@ -453,14 +453,23 @@ func (e *FileConfigError) Error() string {
 // In order to do proper validation, we actually read the file twice -- once into
 // a map, and once into the actual config object.
 func newFileConfig(opts *CmdEnv) (*fileConfig, error) {
+	configReaders, err := getConfigDataForLocations(opts.ConfigLocations)
```
kentquirk, line 456: Same here.

kentquirk (approved):
Thank you! I'm happy now.

PR #1472, opened 2025-01-21
feat(experiment): make drop decision queue size configurable

VinozzZ:

The `cuckoo_addqueue_full` shows full when experimenting HPA. I would like to increase the queue size since now each refinery will hold all trace decisions in the cluster instead of just 1/Nth

benchmark with the queue size increased from 1000 to 10000
```
goos: darwin
goarch: arm64
pkg: github.com/honeycombio/refinery/collect/cache
cpu: Apple M2 Max
BenchmarkCuckooTraceChecker_Add-12                 	268866240	        8.203 ns/op	      0 B/op	      0 allocs/op
BenchmarkCuckooTraceChecker_AddParallel-12         	1957152	      603.1 ns/op	      2 B/op	      0 allocs/op
BenchmarkCuckooTraceChecker_Check-12               	47585214	       25.77 ns/op	      0 B/op	      0 allocs/op
BenchmarkCuckooTraceChecker_CheckParallel-12       	1930326	      615.8 ns/op	      0 B/op	      0 allocs/op
BenchmarkCuckooTraceChecker_CheckAddParallel-12    	 967039	     1296 ns/op	     11 B/op	      0 allocs/op
```

- add `DroppedQueueSize` config option
- add `incoming/peer_router_otlp` metric to track otlp traffic

### DIFF
--- a/collect/cache/cuckoo.go
+++ b/collect/cache/cuckoo.go
@@ -61,7 +61,7 @@ func NewCuckooTraceChecker(capacity uint, addQueueDepth uint, m metrics.Metrics)
 		current:  cuckoo.NewFilter(capacity),
 		future:   nil,
 		met:      m,
-		addch:    make(chan string, defaultAddQueueDepth),
+		addch:    make(chan string, addQueueDepth),
 	}
 	for _, metric := range cuckooTraceCheckerMetrics {
 		m.Register(metric)
--- a/collect/cache/cuckoo.go
+++ b/collect/cache/cuckoo.go
@@ -39,7 +39,7 @@ type CuckooTraceChecker struct {
 
 const (
 	// This is how many items can be in the Add Queue before we start blocking on Add.
-	AddQueueDepth = 1000
+	defaultAddQueueDepth = 1000
 	// This is how long we'll sleep between possible lock cycles.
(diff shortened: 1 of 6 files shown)


collect/cache/cuckoo.go
```diff
@@ -50,13 +50,16 @@ var cuckooTraceCheckerMetrics = []metrics.Metadata{
 	{Name: CurrentLoadFactor, Type: metrics.Gauge, Unit: metrics.Percent, Description: "the fraction of slots occupied in the current cuckoo filter"},
 }
 
-func NewCuckooTraceChecker(capacity uint, m metrics.Metrics) *CuckooTraceChecker {
+func NewCuckooTraceChecker(capacity uint, addQueueDepth uint, m metrics.Metrics) *CuckooTraceChecker {
```
TylerHelmuth, line 53: `addQueueDepth` param isn't used, should it be used on line 62?

route/route.go
TylerHelmuth, line 1: Are these changes related to drop queue size or fixing that bug where we don't increment the right metrics for OTLP ingest during stress relief?

collect/cache/cuckoo.go
VinozzZ: Good catch 😮‍💨

Issue #1477, opened 2025-01-29
logrus_error="can not add field \"reason\"" when sending trace


CalPully:
**Versions**

- Go: 1.23.5
- Refinery: 2.9.1

**Steps to reproduce**

1. Run the refinery docker image
2. Apply configuration using the DeterministicSampler with one or more Rules
3. Send traces to the refinery instance
4. Observe info level logs showing a logrus error "can not add field \"reason\""
  
**Additional context**

The logging library in use uses a `map[string]interface{}` to build the fields for logging. The library performs a check on the data provided that it is not a function or a pointer to a function when calling the `WithFields` method.  In the snippet linked below, `"reason"` is being set to the function `td.KeptReason` whereas I believe the intention is to set to the return of the function, `td.KeptReason()`.

https://github.com/honeycombio/refinery/blob/aa8e92a5568dd702c815116f986cc2a85e7585e9/collect/collect.go#L1017

The `sirupsen/logrus` code performing the check can be found here.

 https://github.com/sirupsen/logrus/blob/d1e6332644483cfee14de11099f03645561d55f8/entry.go#L125

kentquirk:
Closed by #1478.

PR #1478, opened 2025-01-29
fix: error from logrus when sending a trace

nic-littlepay:

Attempts to address issue honeycombio/refinery#1477 

In `collect.go`, it changes values passed to logger so as to avoid an error in logrus lib (see issue for repro steps). Begins passing a string value to `logFields` map where currently a function ref is being passed. From looking at [logrus source at this line](https://github.com/sirupsen/logrus/blob/master/entry.go#L135), it seems the library treats passing functions or function pointers as an error.

The code was passing `td.KeptReason` function pointer to logger, but I've made the assumption that the intended value was actually the `td.Reason` string. This assumption is based on the logFields map using key of `"reason"` (rather than it using `"keptReason"`) and also that it is `td.Reason` being subsequently passed to `sendableTrace{...}`. 

Was unclear where to add a unit test as anywhere I began writing one it just felt like I was testing logrus directly, rather than refinery, but happy to be advised on best place for a test.

### DIFF
--- a/collect/collect.go
+++ b/collect/collect.go
@@ -1014,7 +1014,7 @@ func (i *InMemCollector) send(ctx context.Context, trace *types.Trace, td *Trace
 	} else {
 		logFields["environment"] = td.SamplerSelector
 	}
-	logFields["reason"] = td.KeptReason
+	logFields["reason"] = td.Reason
 	if td.SamplerKey != "" {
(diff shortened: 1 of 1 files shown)


VinozzZ (approved):
Thanks for fixing the bug! You are correct that this field is intended for `td.Reason`

PR #1480, opened 2025-02-03
fix: only record attributes when otel tracing is enabled

VinozzZ:

In profiling data, the `Attributes()` is called even though otel tracing isn't enabled. This is a waste of CPU resource. We should only compute it when tracing is enabled

- check whether tracing is enabled in `otelutil`


MikeGoldsmith (approved):
Looks good. Do we have an estimate of how much CPU we were wasting in the profiling data?

PR #1483, opened 2025-02-07
fix: don't assume attribute type in oteltracing

VinozzZ:

When a non-string type value is passed in to the `AddTelemetryAttributeFunc` when `OTELTracing` is enabled, Refinery currently panics

- don't assume the attribute type

### DIFF
--- a/cmd/refinery/main.go
+++ b/cmd/refinery/main.go
@@ -249,6 +249,12 @@ func main() {
 				span.SetAttributes(attribute.String(key, v))
 			case bool:
 				span.SetAttributes(attribute.Bool(key, v))
+			case int:
+				span.SetAttributes(attribute.Int(key, v))
+			case float64:
+				span.SetAttributes(attribute.Float64(key, v))
+			default:
+				span.SetAttributes(attribute.String(key, fmt.Sprintf("%v", v)))
 			}
 		}
 	}
--- a/cmd/refinery/main.go
+++ b/cmd/refinery/main.go
@@ -244,7 +244,12 @@ func main() {
 		// add telemetry callback so husky can enrich spans with attributes
 		husky.AddTelemetryAttributeFunc = func(ctx context.Context, key string, value any) {
 			span := trace.SpanFromContext(ctx)
-			span.SetAttributes(attribute.String(key, value.(string)))
+			switch v := value.(type) {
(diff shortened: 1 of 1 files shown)


cmd/refinery/main.go
```diff
@@ -244,7 +244,12 @@ func main() {
 		// add telemetry callback so husky can enrich spans with attributes
 		husky.AddTelemetryAttributeFunc = func(ctx context.Context, key string, value any) {
 			span := trace.SpanFromContext(ctx)
-			span.SetAttributes(attribute.String(key, value.(string)))
+			switch v := value.(type) {
+			case string:
```
JamieDanielson, line 248: It seems like the previous version was intending to cast the value provided as a String. Are there unexpected results here if we allow a Bool vs casting a Bool to a string? Is there ever a case where we might get something other than a String or Bool (e.g. an Int)?

VinozzZ: we were asserting the Bool to a string which can panic if the assertion fails. I'm not sure what other types there could be but the change is at least making sure we are not going to panic

kentquirk: I feel that if we go this route we should at least have cases for generic int and float as well as bool.

But another option would be to say we're always going to send a String attribute and do
`span.SetAttributes(attribute.String(key, fmt.Sprintf("%v", v)))`

JamieDanielson: Aha, yes, I just realized this. The casting to a string is what was causing the issue. Here we're handling a string or bool properly, and seems it will basically ignore anything else (at least in local testing with otel-go that was the case). Thanks!

JamieDanielson: > But another option would be to say we're always going to send a String attribute and do
> `span.SetAttributes(attribute.String(key, fmt.Sprintf("%v", v)))`

I like this idea. Then we can decide later if we need to allow other data types.

kentquirk: OK, after talking to Jamie, we think we should add cases for int and float as well, and then add a default case that uses the sprintf technique to cover all the other types that could be sent. Then this won't crash and it will get the desired data to the destination even if it's ill-formed.

kentquirk (approved):
Love it. Thanks.

PR #1485, opened 2025-02-10
feat: Try sync.pool to manage events

MikeGoldsmith:

Tries out using a `sync.pool` to manage Event structs, which is the most commonly allocated and deallocated struct.

I chose to pool Event as that's where the variable size `map[string]any` lives and can also then be used for non-span data that gets passed through Refinery (eg regular events and logs with no Trace ID).

Could try pooling span instead if we wanted to make the interfaces a little cleaner.

-

### DIFF
--- a/route/route.go
+++ b/route/route.go
@@ -437,16 +437,7 @@ func (r *Router) requestToEvent(ctx context.Context, req *http.Request, reqBod [
 		return nil, err
 	}
 
-	return &types.Event{
-		Context:     ctx,
-		APIHost:     apiHost,
-		APIKey:      apiKey,
-		Dataset:     dataset,
-		Environment: environment,
-		SampleRate:  uint(sampleRate),
-		Timestamp:   eventTime,
-		Data:        data,
-	}, nil
+	return types.NewEvent(ctx, apiHost, apiKey, dataset, environment, uint(sampleRate), eventTime, data), nil
 }
 
 func (r *Router) batch(w http.ResponseWriter, req *http.Request) {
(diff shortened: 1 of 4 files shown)


route/route.go
```diff
@@ -437,16 +437,7 @@ func (r *Router) requestToEvent(ctx context.Context, req *http.Request, reqBod [
...
 
-	return &types.Event{
-		Context:     ctx,
-		APIHost:     apiHost,
-		APIKey:      apiKey,
-		Dataset:     dataset,
-		Environment: environment,
-		SampleRate:  uint(sampleRate),
-		Timestamp:   eventTime,
-		Data:        data,
```
VinozzZ, line 448: Does that mean the sync.Pool also manages the `Data` field or it just manages the pointer to it?
Since the bulk of the data is in the `Data` field, it would be great if we can reuse the allocation for that

MikeGoldsmith: As Event.Data is a map[string]interface{} (not a pointer), I think the pooled events will manage that too.

VinozzZ: Oh sorry, I meant that the map[string]interface{} is referenced by pointer not value. That means the `data` passed into the `type.NewEvent` method is a pointer to the map[string]interface{}. The creation of the allocation for map[string]interface{} is 3 lines above. I suspect that the `sync.Pool` is not managing the actual creation of the `data`

kentquirk:
It will be interesting to see if this shows a performance improvement (I have my doubts but also could be very wrong). How do you expect to validate whether this has had an effect?

PR #1491, opened 2025-02-19
Yingrong/refinery opamp health

VinozzZ:

Report Refinery health status through opamp

- run health check on a ticker in refinery opamp agent
- add tests
- I also started working on adding metrics and custom message capability. I probably should split those changes out into a separate PR

### DIFF
--- a/internal/health/mock.go
+++ b/internal/health/mock.go
@@ -1,0 +1,33 @@
+package health
+
+import "sync"
+
+type MockHealthReporter struct {
+	isAlive bool
+	isReady bool
+	mutex   sync.Mutex
+}
+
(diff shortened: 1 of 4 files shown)


internal/health/mock.go
```diff
@@ -0,0 +1,33 @@
+package health
```
codeboten, line 1: i'd rename this to `mock_test.go` to exclude it from being built in the binary

codeboten (approved):

internal/health/mock.go
VinozzZ: Unfortunately, if a file is named with `_test.go`, go disallow referencing any of its exported code from another package.

PR #1492, opened 2025-02-21
feat: send bytes_received through opamp custom message

VinozzZ:

- This PR enables Refinery to send `bytes_received` metrics through opamp messages

- create `usageTracker` in opamp agent to track `bytes_received` metrics reported from Refinery internal metrics
- `usageTracker` does all the calculation and conversion to otlp resources metrics
- add tests


agent/otel_metrics.go
```diff
@@ -0,0 +1,55 @@
...
+	d.Attributes().PutStr("signal", string(signal))
+	return nil
+}
+
+func convertFloat64ToInt64(value float64) (int64, error) {
+	if value > math.MaxInt64 {
+		return 0, fmt.Errorf("value %f is too large to convert to int64", value)
+	}
+	if value < math.MinInt64 {
+		return 0, fmt.Errorf("value %f is too small to convert to int64", value)
```
kentquirk, line 52: "too small" is probably the wrong phrase here -- MinInt64 is a very large negative number. 

I'm also wondering why you bothered to introduce this function at all. It's not actually wrong, but if you had just used int64(value) it would only error if a single report is more than about 9000 terabytes. Then the calling function wouldn't have to return errors either.

agent/usage_report.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+	ur.datapoints = ur.datapoints[:0]
+	return data, nil
+}
+
+type usageSignal string
+
+var (
+	signal_trace usageSignal = "trace"
+	signal_log   usageSignal = "log"
```
codeboten, line 70:
```suggestion
	signal_trace usageSignal = "traces"
	signal_log   usageSignal = "logs"
```

agent/otel_metrics.go
VinozzZ: If there's a overflow issue that causes a datapoint missing, I would like to be able to capture it in refinery's telemetry data. My thinking is that, that's money missing for honeycomb so I want to make sure it's visible

kentquirk: OK, fine, it's certainly not wrong. But if someone is overflowing an int64 the money missing would be *at least* in the millions of dollars..  I'd like to have that problem. :)

agent/agent.go
```diff
@@ -188,20 +196,99 @@ func (agent *Agent) Stop(ctx context.Context) {
...
+		case <-agent.ctx.Done():
+			// TODO: drain the existing reports
+			return
+		case <-timer.Chan():
+			if err := agent.sendUsageReport(); err != nil {
+				if errors.Is(err, errNoData) {
+					agent.logger.Debugf(context.Background(), "No data to report")
+					continue
+				}
+				agent.logger.Errorf(context.Background(), "MONEY STEALING. Could not send usage report: %v", err)
```
TylerHelmuth, line 243:
```suggestion
				agent.logger.Errorf(context.Background(), "Could not send usage report: %v", err)
```

agent/agent.go
```diff
@@ -165,6 +172,7 @@ func (agent *Agent) connect() error {
 	agent.logger.Debugf(context.Background(), "started opamp client")
 
 	go agent.healthCheck()
+	go agent.usageReport()
```
TylerHelmuth, line 175: What is the difference between healthCheck and usageReport?

VinozzZ: The`healthCheck` is the one that reports Refinery health through the opamp health check capability.
`usageReport` is to report the usage metrics to through the custom message. 
I wanted to separate the usage metrics calculation from the actual reporting interval. That's why I created a separate goroutine for reporting itself

agent/agent.go
```diff
@@ -188,20 +196,99 @@ func (agent *Agent) Stop(ctx context.Context) {
...
+			}
+
+			now := agent.clock.Now()
+			agent.usageTracker.Add(newTraceCumulativeUsage(traceUsage, now))
+			agent.usageTracker.Add(newLogCumulativeUsage(logUsage, now))
+		}
+	}
+}
+
+func (agent *Agent) usageReport() {
```
TylerHelmuth, line 228: Since this is specific to sending the report and not accumulating the metrics and we update the name to reflect that?

agent/usage_report.go
```diff
@@ -0,0 +1,116 @@
...
+// Add records the current cumulative usage and calculates the delta between the last update.
+func (ur *usageTracker) Add(data usage) {
+	ur.mut.Lock()
+	defer ur.mut.Unlock()
+
+	if data.val == 0 {
+		return
+	}
+
+	deltaTraceUsage := data.val - ur.lastUsageData[data.signal].val
```
TylerHelmuth, line 43: What happens if `ur.lastUsageData[data.signal].val` is 0 or if `ur.lastUsageData[data.signal]` is nil?

VinozzZ: When a new signal shows up for the first time, the `ur.lastUsageData[data.signal]` will return the zero value for `usage` type. The zero value of `usage` is 
```
usage{
    signal:    "", // Zero value for string
    val:       0,  // Zero value for float64
    timestamp: time.Time{}, // Zero value for time.Time
}
```
So when the lone `deltaTraceUsage := data.val -ur.lastUsageData[data.sginal].val` executes, it's effectively computing `data.val - 0` for a new signal that hasn't been seen before

TylerHelmuth (approved):

PR #1494, opened 2025-02-25
feat: Use Github Actions to build and publish docker images to GHCR

irvingpop:

- As of April 1st, 2025 [Dockerhub will further clamp down on pull rate limits for unauthenticated users](https://docs.docker.com/docker-hub/usage/), which might make things difficult for our customers with larger Refinery deployments.    
- In order to improve the resilience of customer deployments, let's dual-publish Refinery images to Github Container Registry (GHCR)
- One challenge with extending our CircleCI pipeline to do this is that [external auth to GHCR is limited to PATs only](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#authenticating-to-the-container-registry)

- This PR adds a Github Actions native workflow, which builds and publishes a Docker container exactly the same way as `build-docker.sh` and CircleCI, but with a greatly simplified configuration.

I tested this out in a forked repo, with the results here: https://github.com/irvingpop/refinery/pkgs/container/refinery

CleanShot 2025-02-24 at 16 21 16


.github/workflows/publish-docker-ghcr.yml
```diff
@@ -0,0 +1,43 @@
...
+          MINOR_VERSION=$(echo $VERSION | cut -d. -f1,2)
+          echo "MINOR_VERSION=${MINOR_VERSION}" >> $GITHUB_ENV
+          # Set the build ID for the binary
+          echo "GOFLAGS=-ldflags=-X=main.BuildID=${VERSION}" >> $GITHUB_ENV
+          # Set the source build date for ko
+          echo "SOURCE_DATE_EPOCH=$(git log -1 --format=%ct)" >> $GITHUB_ENV
+
+      - name: Build and push Docker images using ko
+        run: >
+          ko build
```
TylerHelmuth, line 36: @JamieDanielson had an idea that instead of building a new image, can this action be dependent on the cirlceci job completing and then pull the image and re-push it?

JamieDanielson: Also we'd want/need this to happen after a successful build and publish. So it looks like you can change the trigger to be on [release publish](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#release)

```yaml
on:
  release:
    types: [published]
```

irvingpop: Oh neat!  updated.

verythorough:
I'm sure y'all have already thought of this, but I wanted to note that once we have this option, we'll probably want to add it to the docs anywhere we mention using Docker Hub.

irvingpop:
Note:  if we do go with this PR,  let's pin all the Actions shas before merging.    I can take care of that if there's interest in going this route (although I understand the binary divergence isn't ideal, but FWIW the current dockerhub builds aren't signed)

MikeGoldsmith:
Sorry for the long wait. We wanted to make sure we are consistent with how we publish packages.

We're going to update the existing CircleCI workflow to publish the image to GHCR. I'll close this PR and will work on that update soon.

PR #1496, opened 2025-02-25
feat: record bytes_received metrics for traces and logs signal

VinozzZ:

- Add a new `bytes_received_*` metric for both trace and log signal so that we can retrieve it later in our opamp agent

- Record `bytes_received` only for `incoming` data before the logic to route peer traffic. This is to prevent double counting peer traffic.
- Store counter values in `multi_metrics` so that metrics value can be retrieved later

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -214,7 +214,7 @@ func TestReload(t *testing.T) {
 	watcher := &configwatcher.ConfigWatcher{
 		Config: c,
 		PubSub: pubsub,
-		Logger: &logger.StdoutLogger{},
+		Logger: &logger.NullLogger{},
 	}
 	watcher.Start()
 	defer watcher.Stop()
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/honeycombio/refinery/config"
 	"github.com/honeycombio/refinery/internal/configwatcher"
+	"github.com/honeycombio/refinery/logger"
 	"github.com/honeycombio/refinery/pubsub"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
@@ -213,6 +214,7 @@ func TestReload(t *testing.T) {
 	watcher := &configwatcher.ConfigWatcher{
 		Config: c,
 		PubSub: pubsub,
+		Logger: &logger.StdoutLogger{},
 	}
 	watcher.Start()
 	defer watcher.Stop()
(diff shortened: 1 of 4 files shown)


metrics/multi_metrics_test.go
```diff
@@ -91,10 +91,10 @@ func TestMultiMetrics_Register(t *testing.T) {
 	mm.Down("updown")
 	mm.Gauge("gauge", 42)
 
-	// counter should be 0 because it's not tracked by StoreMetrics
+	// counter should be 1 because it's tracked by StoreMetrics
```
VinozzZ, line 94: @kentquirk , do you know why we didn't store counters in the `StoreMetrics`?

kentquirk: They're not stored because they're reset on each metrics scrape, so they're not useful for stress relief.

https://github.com/honeycombio/refinery/blob/main/metrics/multi_metrics.go#L17

VinozzZ: Do you mind to point out where this values map resets? I don't think I understand how the reset works for them

kentquirk: The values map doesn't reset, but counters do. The value you read from the values map won't reflect the value sent to Honeycomb because that one resets but the local values map doesn't.

VinozzZ: Ok, that aligns with my understanding.I will update the comments so it's clear that this is also used by the opamp agent now

config/config_test.go
```diff
@@ -213,6 +214,7 @@ func TestReload(t *testing.T) {
 	watcher := &configwatcher.ConfigWatcher{
 		Config: c,
 		PubSub: pubsub,
+		Logger: &logger.NullLogger{},
```
codeboten, line 217: is this change needed?

route/route.go
```diff
@@ -602,6 +606,14 @@ func (r *Router) processEvent(ev *types.Event, reqID interface{}) error {
 		IsRoot:  isRootSpan(ev, r.Config),
 	}
 
+	if r.incomingOrPeer == "incoming" {
+		if span.Data["meta.signal_type"] == "log" {
+			r.Metrics.Count("bytes_received_logs", span.GetDataSize())
```
codeboten, line 611: are there any performance considerations around GetDataSize or is it returning a value that's already been calculated?

VinozzZ: That's a good point. I changed it to only track it when opamp is enabled like what we talked about

config/config_test.go
VinozzZ: Yes, because the test would panic if the logger is nil

codeboten (approved):

PR #1510, opened 2025-03-20
feat: Record the number of bytes per event

MikeGoldsmith:

Adds a new histogram metric to record the number of bytes per event as they are received.

- Add new histogram metrics for recording bytes per event as all event types are received
- Move DataSize field from span to event and store it when calling `GetDataSize`
- Update `GetDataSize` so it can recursively walk slices and maps
- Add unit tests and update existing tests with updated datasize attributes


types/event.go
```diff
@@ -33,6 +33,53 @@ type Event struct {
...
+		return 1
+	case float64, int64, int:
+		return 8
+	case string:
+		return len(value)
+	case []byte: // also catch []uint8
+		return len(value)
+	case []bool:
+		return len(value)
+	case []float64:
```
kentquirk, line 60: I don't think it hurts to look at all these slice types, but I don't think that in practice you're ever going to get anything other than a `[]any` which you don't handle here explicitly. You definitely need that.

kentquirk: I say that because the JSON unmarshaler won't do anything else and I'm pretty sure that Husky converts all array types into []any as well.

MikeGoldsmith: I did originally have `[]any` in the list, but in tests I never hit that case. Maybe I need to add a test that explicitly uses `[]any` as the data entry type.

MikeGoldsmith: Added case for `[]any` in [236113e](https://github.com/honeycombio/refinery/pull/1510/commits/236113e1708c4855c374fe523ac6f57270ca13b3)

kentquirk: Even if it doesn't go into the automated tests, it might be worth doing a manual test that sends, say, a []int in a field through the OTLP endpoint and see how it shows up in this function.

MikeGoldsmith: You're right about [Husky](https://github.com/honeycombio/husky/blob/main/otlp/common.go#L408) using []interface as for array types. I'll confirm with the JSON unmarshaller and may remove the other cases.

kentquirk (approved):
I think this does what it says it does. But I'm wondering if it might be useful to also add in something like `8*len(e.Data)` to try to account for the overhead of the Event object itself. 

That's not a blocker, and I'm not even sure it's helpful. I thought I'd suggest it.

PR #1515, opened 2025-03-27
docs: add missing metrics and extra details

JamieDanielson:

- missing some stress relief metrics
- missing some extra detail that may be useful

- add metric `libhoney_peer_queue_length` to docs
- add metric `libhoney_upstream_queue_length` to docs
- update notes on stress relief, redistribution, available memory, and negative operators.
- run `make all` to generate docs


config.md
```diff
@@ -1343,6 +1348,7 @@ SamplingRate is the sampling rate to use when Stress Relief is activated.
 All new traces will be deterministically sampled at this rate based only on the `traceID`.
 It should be chosen to be a rate that sends fewer samples than the average sampling rate Refinery is expected to generate.
 For example, if Refinery is configured to normally sample at a rate of 1 in 10, then Stress Relief should be configured to sample at a rate of at least 1 in 30.
+If this value is set too low it will drastically increase the amount of data sent to Honeycomb, which could overwhelm the upstream queue.
```
verythorough, line 1351: What does "too low" mean? 

Is "1 in 10" lower or higher than "1 in 30"? How far outside that range would be "too low"?

config/metadata/configMeta.yaml
```diff
@@ -1258,6 +1258,7 @@ groups:
           `MB` and `GiB`) and Kubernetes units (such as `M` and `Gi`) are
           supported. Fractional values with a suffix are supported. If
           `AvailableMemory` is set, `Collections.MaxAlloc` must not be defined.
+          A useful value for this setting is about 85% of the pod's total memory.
```
codeboten, line 1261: It would be great to have more information re. the 85% recommendation. "about 85%" makes me wonder why not 90, or 80

kentquirk: Maybe something like "A useful value is to leave 1-2GB of pod memory for overages. For typical configurations, this might be 85 to 90% of the pod's total memory."

config/metadata/configMeta.yaml
```diff
@@ -1307,8 +1308,12 @@ groups:
         summary: controls whether to transmit traces in cache to remaining peers during cluster scaling event.
         description: >
           If `true`, Refinery will NOT forward live traces in its cache to the rest of the peers when peers join or leave the cluster.
-          By disabling this behavior, it can help to prevent disruptive bursts of network traffic when large traces with long `TraceTimeout`
-          are redistributed.
+          Disabling redistribution can help to prevent disruptive bursts of network traffic when large traces with long `TraceTimeout`
+          are redistributed. However, it will also cause data loss during scale up/down events.
```
kentquirk, line 1312:
```suggestion
          are redistributed. However, it may also cause partial data loss of traces that were in flight when scaling occurred.
```

config/metadata/configMeta.yaml
```diff
@@ -1307,8 +1308,12 @@ groups:
...
+
+          Redistribution is intended to help prevent data loss and make whole trace decisions on scale up/down events.
```
kentquirk, line 1314:
```suggestion
          Redistribution is intended to help prevent data loss by reconsolidating traces onto the appropriate node after a scaling event.
```

tools/convert/templates/configV2.tmpl
```diff
@@ -2,7 +2,7 @@
 ## Honeycomb Refinery Configuration ##
 ######################################
 #
-# created {{ now }} from {{ .Input }} using a template generated on 2025-02-13 at 16:52:04 UTC
+# created {{ now }} from {{ .Input }} using a template generated on 2025-03-27 at 23:12:19 UTC
```
kentquirk, line 5: for files where the only change was the generation date, I usually just revert them rather than include them in the commit.

config.md
kentquirk: I wish we had just used sampling probability or a name other than sample rate!

Maybe something more elaborate like:
> If this value is configured to send more data to honeycomb during stress relief than the normal average sampling strategy, it may overwhelm the upstream send queue and have the opposite of the desired effect.

kentquirk (commented):
I'm not going to approve yet, since this is still draft, but it looks pretty good. I've put in some suggestions.

config/metadata/configMeta.yaml
```diff
@@ -1307,8 +1308,12 @@ groups:
...
+          This feature does incur additional resource usage. If the cluster does not have enough resource capacity headroom,
+          a redistribution event can cause the cluster to go into stress relief (if enabled).
```
MikeGoldsmith, line 1316: Re-ordered the sentences and removed the precent disruptive bursts statement. All forwarded spans contribute to the additional resources being used during a scale up/down event, not just long traces.

```suggestion
Redistribution is intended to help prevent data loss and make whole trace decisions during cluster scale up/down events by redistributing stored spans to peers.
During scale down events, all stored spans are forwarded to the new owning Refinery peer instance so it can shut down without dropping any spans.
During scale up events, only stored spans that belong to another Refinery peer instance are forwarded so that instance can make whole trace decisions.
Refinery uses additional system resources during scale up/down events.
If the cluster does not have enough resource capacity headroom, a redistribution event can cause the cluster to go into stress relief (if enabled).
Disabling redistribution will incur data loss during scale up/down events.
```

config/metadata/configMeta.yaml
MikeGoldsmith, line 1261: I think 85% is a bit conservative. We usually recommend 90-95% so it leaves some headroom to the OS but still gets most of it. I also like Kent's phrasing for leaving some headroom for the OS.

```suggestion
          A good starting value for this setting is 90% of the pod's total memory.
```

MikeGoldsmith (commented):
This is a great start, thank you @JamieDanielson

config.md
JamieDanielson: Used your wording @kentquirk

config/metadata/configMeta.yaml
JamieDanielson: Used your wording @kentquirk

config/metadata/configMeta.yaml
JamieDanielson: Makes sense, I've reorganized per your suggestion!

config/metadata/configMeta.yaml
JamieDanielson: updated based on another thread, and did 85-90% with a specific note of 1-2GB.

MikeGoldsmith (approved):

PR #1520, opened 2025-04-02
fix: update trace_id field to meta.trace_id

codeboten:

- This fixes a bug identified when span data processed by refinery includes a `trace_id` attribute. This change updates the field to `meta.trace_id` which should avoid collisions with existing instrumentation that sets `trace_id`.

- Update `trace_id` field for internal decisions to `meta.trace_id`
- Update the tests


kentquirk (approved):
This seems to be what we need. Thanks.

dustxd (approved):
Not sure if there's more context that I should be aware of but fix seems reasonable for problem description!

kentquirk (approved):
Sigh, I already approved it.

PR #1521, opened 2025-04-02
call UpdatedPeers callbacks on FilePeers config updates

wjordan:

- Make it possible to plug in non-Redis systems for dynamic peer discovery.
- Supporting DNS-based service-discovery is the more specific use-case. We have been running a [custom peer implementation](https://github.com/honeycombio/refinery/commit/34192c7930eec91bb317a7d74f405b7c0db9ea93) that periodically queries [Fly.io `.internal` DNS](https://fly.io/docs/networking/private-networking/#fly-io-internal-dns) for a list of peers, but with this PR we could instead write the peer list to the config file in a background process and keep the implementation simple.

- The change hooks up the existing `UpdatedPeersCallback` to the existing `file` peer-management type using a reload callback on the config file. When the `PeerManagement.Peers` entry is updated in the config file, it will trigger the registered UpdatedPeers callbacks.

### DIFF
--- a/internal/peer/file.go
+++ b/internal/peer/file.go
@@ -15,7 +15,8 @@ type FilePeers struct {
 	Metrics metrics.Metrics `inject:"metrics"`
 	Done    chan struct{}
 
-	id string
+	id        string
+	callbacks []func()
 }
 
 func (p *FilePeers) GetPeers() ([]string, error) {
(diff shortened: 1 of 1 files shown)


kentquirk:
Thanks, @wjordan. 

I like the philosophy of the change. We have a few people out this week, but we'll look into it together next week and let you know.

VinozzZ:
Thank you for submitting this change. After talking with the team, we think this is a good feature to add for the file peer.

Do you mind to update the corresponding doc section [here](https://github.com/honeycombio/refinery/blob/e939a9ccacc926aa64cbf222a8941aa2d48cced7/config/metadata/configMeta.yaml#L997) in this PR as well?

Issue #1524, opened 2025-04-09
When TraceLocalityMode is Distributed, dump incoming spans on a peer when stressed


mterhar:
**Is your feature request related to a problem? Please describe.**

Today, if you have Distributed mode enabled, the goal is for each refinery node in the cluster to take whatever spans it's given and hold them until a decision is made by the deciding node, then take the decided action. This works fine until a few pods in the cluster gets stressed. 

This can be due to a load balancer configuration or a client config that skips the load balancer or a dumb load balancer like iptables. 

**Describe the solution you'd like**

When in Distributed mode, it ignores the "individual_stress_level" metric and uses the cluster_stress_level instead. This means that instead of stressing out when it's memory gets full, it just drops the spans on the floor.  The way to see this is `incoming_router_dropped` metric goes up during these periods. 

Rather than incoming_router_dropped, it would be _nice_ if it could send the spans to another pod. It could even do a sanity check like if individual_stress_level > cluster_stress_level, send to it's proper-shard peer. If it is already on the proper shard peer, just do `ProcessSpanImmediately` on it. 

**Describe alternatives you've considered**

Alternatives:

1. Send a signal back to the load balancer to tell it to lay off delivering data to the node. A 429 response maybe?
2. Include something about the `r.Collector.AddSpan(span)` channel in the health measurement so a failure to add here would trigger to "processSpanImmediately()` function.

**Additional context**

PR #1525, opened 2025-04-10
fix: add a retry on span routing

mterhar:

- #1524 

- Adds error handling so if it can't add a span to the cache, it will send it elsewhere or process it immediately as a stress behavior


jhonmike (approved):
LGTM

mterhar:
Here's a chart showing the impact of this PR on a cluster that is 30 nodes and handling about a million spans per second.

VinozzZ:

PR #1527, opened 2025-04-14
fix: ensure SendKeyMode priority

codeboten:

- Prior to this change, the functionality enabled by SendKeyMode contained a bug that would prevent it from taking precedence over the x-honeycomb-team header if the header was not present.

- updated the code to check for `SendKeyMode` prior to validating headers
- added tests for both OTLP endpoints (http/grpc) for logs and traces

### DIFF
--- a/route/otlp_logs.go
+++ b/route/otlp_logs.go
@@ -2,7 +2,6 @@ package route
 
 import (
 	"context"
-	"errors"
 	"net/http"
 
 	huskyotlp "github.com/honeycombio/husky/otlp"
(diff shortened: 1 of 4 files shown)


route/otlp_logs.go
```diff
@@ -64,20 +66,18 @@ func (l *LogsServer) Export(ctx context.Context, req *collectorlogs.ExportLogsSe
...
 	apicfg := l.router.Config.GetAccessKeyConfig()
 	if err := apicfg.IsAccepted(ri.ApiKey); err != nil {
 		return nil, status.Error(codes.Unauthenticated, err.Error())
 	}
-	keyToUse, err := apicfg.GetReplaceKey(ri.ApiKey)
+	keyToUse, _ := apicfg.GetReplaceKey(ri.ApiKey)
 
-	if err != nil {
-		return nil, status.Error(codes.Unauthenticated, err.Error())
+	if err := ri.ValidateLogsHeaders(); err != nil && err != huskyotlp.ErrMissingAPIKeyHeader {
```
VinozzZ, line 75: Do we need to handle the case where `err == huskyotlp.ErrMissingAPIKeyHeader && len(keyToUse) ==0` ?

codeboten: That case is captured in `TranslateLogsRequest` at line 81, which calls ValidateLogsHeaders again

VinozzZ (approved):

PR #1535, opened 2025-04-18
fix: improve trace cache doc and metric

### LINKED ISSUE #1502: Outdated metric referenced in README.md
`collect_cache_buffer_overrun` was recently removed (https://github.com/honeycombio/refinery/pull/1452) but is still called out as an important metric to watch in the README. Is there a new metric we should be watching instead? 

https://github.com/honeycombio/refinery?tab=readme-ov-file#scaling-up

> Refinery uses bounded queues and circular buffers to manage allocating traces, so even under high volume memory use shouldn't expand dramatically. However, given that traces are stored in a circular buffer, when the throughput of traces exceeds the size of the buffer, things will start to go wrong. If you have statistics configured, a counter named **collect_cache_buffer_overrun** will be incremented each time this happens. The symptoms of this will be that traces will stop getting accumulated together, and instead spans that should be part of the same trace will be treated as two separate traces.

Side note: would be nice if there was an up-to-date runbook on how to operate a refinery cluster.

VinozzZ:

This PR updates the outdated `Scale Up` section of the README to make it more clear how to manage memory consumption in Refinery.

- fixes #1502 

- Updates README
- Add a new metric, `collector_cache_eviction` to signal cache eviction operation


collect/collect.go
```diff
@@ -270,6 +271,7 @@ func (i *InMemCollector) checkAlloc(ctx context.Context) {
 	if maxAlloc == 0 || mem.Alloc < uint64(maxAlloc) {
 		return
 	}
+	i.Metrics.Increment("collector_cache_eviction")
```
TylerHelmuth, line 274: Is it technically true that if we make it here we'll definitely evict something? An extreme example would be if the cache is empty we wouldnt, but this would still increment. More realistic would be everything matching the `countinue` conditions in the loop.

VinozzZ: `checkAlloc` is only called when `peer_queue` is empty. That would mean some spans are ingested and stored in the cache. I don't think it's possible this can increment without items in the cache

TylerHelmuth: Is it possible that non of the items in the cache would be evicted?

VinozzZ: I don't believe that would be possible in `concentrated` mode

PR #1537, opened 2025-04-21
maint: record sampler key cardinality as metric

### LINKED ISSUE #1365: Add trace key length as metric
The trace key builder was updated to have a maximum number of unique values. This issue tracks exposing the number of values used to build a trace key as a metric, likely a histogram.

---

              OK; the function here doesn't have access to a logger, so I had it return the key size it's using and log if it is the max value.

It occurs to me that we could now make a metric out of this, but I don't have time to do that. Please feel free if you think that's a good idea.

_Originally posted by @kentquirk in https://github.com/honeycombio/refinery/pull/1364#discussion_r1788003814_

VinozzZ:

- closes #1365 

- add a new metric, `<sampler-type>_trace_key_length`, for dynamic samplers to record the number of unique value in a trace key

### DIFF
--- a/sample/sample.go
+++ b/sample/sample.go
@@ -111,7 +111,7 @@ var samplerMetrics = []metrics.Metadata{
 	{Name: "_num_dropped", Type: metrics.Counter, Unit: metrics.Dimensionless, Description: "Number of traces dropped by configured sampler"},
 	{Name: "_num_kept", Type: metrics.Counter, Unit: metrics.Dimensionless, Description: "Number of traces kept by configured sampler"},
 	{Name: "_sample_rate", Type: metrics.Histogram, Unit: metrics.Dimensionless, Description: "Sample rate for traces"},
-	{Name: "_trace_key_length", Type: metrics.Histogram, Unit: metrics.Dimensionless, Description: "Number of unique trace keys"},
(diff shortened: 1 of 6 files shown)


sample/sample.go
```diff
@@ -111,6 +111,7 @@ var samplerMetrics = []metrics.Metadata{
 	{Name: "_num_dropped", Type: metrics.Counter, Unit: metrics.Dimensionless, Description: "Number of traces dropped by configured sampler"},
 	{Name: "_num_kept", Type: metrics.Counter, Unit: metrics.Dimensionless, Description: "Number of traces kept by configured sampler"},
 	{Name: "_sample_rate", Type: metrics.Histogram, Unit: metrics.Dimensionless, Description: "Sample rate for traces"},
+	{Name: "_trace_key_length", Type: metrics.Histogram, Unit: metrics.Dimensionless, Description: "Number of unique trace keys"},
```
kentquirk, line 114: I find this name confusing. At first I thought it was referring to the length of trace.trace_id. Could we rename it to "_sampler_key_cardinality", perhaps, and change the description to "Number of unique keys being tracked by the sampler"?

kentquirk (approved):

PR #1538, opened 2025-04-22
feat(pubsub): implement google pub/sub

lizthegrey:

- We mandate use of Redis, but some of our clients may prefer to use Google Pub/Sub for its increased durability to single region failure.

- Implements use of Google Pub/Sub as an option instead of Redis for peer to peer metadata communication. Future work: send node to node communication through PubSub as well with a custom Transmission.


config/metadata/configMeta.yaml
```diff
@@ -1047,11 +1052,44 @@ groups:
...
 
+  - name: GooglePeerManagement
+    title: "Google Peer Management"
+    description: >
+      controls how the Refinery cluster communicates between peers when using
+      Google Pub/Sub. Only applies when `PeerManagement.Type` is "google".
+
+    fields:
+      - name: Topic
+        v1group: PeerManagement
```
kentquirk, line 1063: These two `v1` lines can come out here and below; they only apply to settings that existed in 1.x versions of Refinery.

config/metadata/configMeta.yaml
```diff
@@ -1047,11 +1052,44 @@ groups:
...
+        v1name: GoogleTopic
+        type: string
+        valuetype: nonemptystring
+        example: ""
+        reload: false
+        envvar: REFINERY_GOOGLE_TOPIC
```
kentquirk, line 1069: If you want these envvars to work, they also have to be specified in cmdenv.go so they work on the command line too. IIRC, we don't read undecorated envvars separately. 

As of this month, we do allow string configuration values to look like `${ENVVAR}` which does substitute as expected.

kentquirk (commented):
Generally, we try not to mix the license updates and documentation regeneration with functional PRs because it makes them harder to review, and we'd be revisiting a lot of work. We usually do those only right before a release. 

If you'd be willing to revert the license update, it would be appreciated.

config/metadata/configMeta.yaml
lizthegrey: Thanks! The process of adding new config flags is a wee bit under-specified I fear so I did a fair bit of cargo cult copying and also doc regenerating.

PR #1541, opened 2025-04-23
feat(experiment): use the consistent hashing algorithm from the collector

VinozzZ:

this is for an experiment that I would like to run.

- use the consistent hashing algorithm from https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/loadbalancingexporter/consistent_hashing.go#L56 to do deterministic sharding


kentquirk:
Why not do this with both sharders implementing a common interface?

mterhar:
We have at least one known deployed refinery where they are using the Loadbalancing processor to separate their traces into 2 different refinery clusters to provide some increased availability but maintain trace consistency.

I am glad this is a configuration because I can see a lot of other situations where the alignment would be very valuable.

PR #1542, opened 2025-04-24
maint: prepare for release v2.9.5

VinozzZ:

- prepare for 2.9.5 release

- update CHANGELOG.md and RELEASE_NOTES.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,30 @@
 # Refinery Changelog
 
+## 2.9.5 2025-04-24
+
+### Features
+
+- feat: add experimental support for opamp (#1528) | [Alex Boten](https://github.com/AlexBoten)
+- feat: Support arbitrary envvar replacement in config strings (#1517) | [Kent Quirk](https://github.com/KentQuirk)
+
+### Fixes
+
+- fix: improve trace cache doc and metric (#1535) | [Yingrong Zhao](https://github.com/vinozzz)
+- fix: flush internal metrics on shutdown (#1534) | [Yingrong Zhao](https://github.com/vinozzz)
+- fix: add env var support for APIHost in telemetry cfg (#1536) | [Yingrong Zhao](https://github.com/vinozzz)
+- fix: ensure SendKeyMode priority (#1527) | [Alex Boten](https://github.com/AlexBoten)
+- fix: update trace_id field to meta.trace_id (#1520) | [Alex Boten](https://github.com/AlexBoten)
+
+### Maintenance
+
+- docs: some copyedits from docs (#1526) | [Jamie Danielson](https://github.com/Jamie Danielson)
+- docs: add missing metrics and extra details (#1515) | [Jamie Danielson](https://github.com/Jamie Danielson)
+- maint: Mark TraceLocalityMode as experimetnal (#1522) | [Mike Goldsmith](https://github.com/Mike Goldsmith)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,30 @@
 # Refinery Changelog
 
+## 2.9.5 2025-04-24
+
+### Features
+
+- feat: add experimental support for opamp (#1528) | [Alex Boten](https://github.com/AlexBoten)
```
JamieDanielson, line 7: May be worth adding an extra comment here reiterating that it is experimental and not intended for use

CHANGELOG.md
```diff
@@ -1,5 +1,30 @@
...
+
+- fix: improve trace cache doc and metric (#1535) | [Yingrong Zhao](https://github.com/vinozzz)
+- fix: flush internal metrics on shutdown (#1534) | [Yingrong Zhao](https://github.com/vinozzz)
+- fix: add env var support for APIHost in telemetry cfg (#1536) | [Yingrong Zhao](https://github.com/vinozzz)
+- fix: ensure SendKeyMode priority (#1527) | [Alex Boten](https://github.com/AlexBoten)
+- fix: update trace_id field to meta.trace_id (#1520) | [Alex Boten](https://github.com/AlexBoten)
+
+### Maintenance
+
+- docs: some copyedits from docs (#1526) | [Jamie Danielson](https://github.com/Jamie Danielson)
```
JamieDanielson, line 20:
```suggestion
- docs: some copyedits from docs (#1526) | [Jamie Danielson](https://github.com/JamieDanielson)
```
looks like the names in the maintenance section still need updates!

RELEASE_NOTES.md
```diff
@@ -2,6 +2,10 @@
 
 While [CHANGELOG.md](./CHANGELOG.md) contains detailed documentation and links to all the source code changes in a given release, this document is intended to be aimed at a more comprehensible version of the contents of the release from the point of view of users of Refinery.
 
+## Version 2.9.5
+
+This release includes improvements to trace cache documentation, internal metric flushing on shutdown, and enhanced metric visibility such as sampler key cardinality. It also introduces environment variable substitution in config strings and experimental support for OpAMP.
```
JamieDanielson, line 7: Similar to the comment above, I think we should be more explicit here that OpAMP is experimental and not intended or supported for customer use.

CHANGELOG.md
VinozzZ: I will add it in the release note. I don't think we typically add extra content in changelog. I could be wrong tho

JamieDanielson: We don't typically, but for folks who aren't aware of a separate release notes file (changelog is pretty standard), it seems worthwhile to point it out just in case.

VinozzZ: Maybe I have the wrong understanding there. I thought Change log is simply the list of changes introduced in a release. 

I added a note at the top section to specifically call out this.

JamieDanielson: > I thought Change log is simply the list of changes introduced in a release.

That is true. And in this repo it seems to serve only as that purpose. But for many repos in OSS it is the only source of truth for changes and people look only at that file, so if there's something critical to be aware of, it's useful to put it in here just in case. Thanks for updating!

JamieDanielson (approved):
🚀

PR #1563, opened 2025-05-14
fix: add _sum to legacy metrics histograms

mterhar:

- Legacy metrics were missing sums for histograms 

- Adds sums to histograms for legacy metrics


MikeGoldsmith (changes requested):
The changes looks good. Please could you add some tests to verify the expected metrics are output correctly?

PR #1564, opened 2025-05-14
maint: Add autogenerated comment to config templates

MikeGoldsmith:

Updates the generated config and rules files to include a markdown comment at the top of the file to indicate they are generated and should not be edited manually.

I confirmed our docs website can handle these markdown comments.

- Update config and rules templates to add an auto generated message
- Regenerate config and rules files to get updated content

### DIFF
--- a/refinery_rules.md
+++ b/refinery_rules.md
@@ -1,3 +1,5 @@
+<!-- Do not edit manually. This is an AUTO-GENERATED file. -->
+
 ## Refinery Rules file
 
 The Refinery `rules` file is a YAML file.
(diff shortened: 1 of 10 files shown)


refinery_rules.md
```diff
@@ -1,3 +1,5 @@
+<!-- Do not edit manually. This is an AUTO-GENERATED file. -->
```
JamieDanielson, line 1: This is not auto-generated. The content is auto-generated in Refinery, then copy-pasted here. See #2545 for an example of the last update made.

JamieDanielson (approved):
Just added a suggestion to include this note in all the templates instead of just some

PR #1568, opened 2025-05-23
maint: Add stronger warning about TraceLocalityMode

verythorough:

- Folks don't always notice the warning that TraceLocalityMode is currently experimental, and are surprised to learn it's not supported. After consulting with support teams, we've selected update wording that's made to be more noticeable, similar to what we've done with a more recently added experimental/internal feature, OpAMP support.

- This PR follows the pattern of https://github.com/honeycombio/refinery/pull/1522, updating `config/metadata/configMeta.yaml` and `refinery_config.md`, then using `make all` in `tools/convert` to generate the other three files.

### DIFF
--- a/config.md
+++ b/config.md
@@ -3,7 +3,7 @@
 # Honeycomb Refinery Configuration Documentation
 
 This is the documentation for the configuration file for Honeycomb's Refinery.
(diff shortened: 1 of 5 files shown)


kentquirk (approved):
Looks correct! Thanks.

PR #1569, opened 2025-05-29
fix: use pooling for zstd decoder

VinozzZ:

When there are more CPU available, we should allow more decoder to work through incoming requests instead of blocking them

- use `sync.Pool` for zstd decoder


kentquirk (changes requested):
Unfortunately, I've figured out this is going to still have the memory leak problem.

In [this section](https://github.com/klauspost/compress/tree/master/zstd#allocation-less-operation), zstd compressor says that you *must* call Close() before freeing the compressor.  But sync.Pool doesn't have a destroy method, so we can't guarantee that happens. And you can't reuse something you've called Close() on. 

This is why we had been using a channel full of decoders. They never get closed. 

One option would be to make a bigger channel. Maybe this would work better if we had, say 10 (or even 50) of these instead of 4. We're running them single-threaded so it might be fine to have a lot.

Another would be to try a pool library that supports a destructor. [This one](https://github.com/jolestar/go-commons-pool?tab=readme-ov-file#go-commons-pool) has a custom factory that does that, but it's the only one I have been able to find that does. It looks pretty good, though.

In any case, this PR as it stands shouldn't go forward. Sorry. :(

VinozzZ:
Thank you @kentquirk for finding the memory leak! I wasn't able to see the memory grow with my local test, but that could just be the fact that the load creating many decoders.

> One option would be to make a bigger channel. Maybe this would work better if we had, say 10 (or even 50) of these instead of 4. We're running them single-threaded so it might be fine to have a lot

I believe this code is being run concurrently. Each http request is being served with a goroutine which then will call this `getMaybeCompressedBody` function. 
I'm tempted to use this [automaxprocs](https://github.com/uber-go/automaxprocs) like we do for our internal services to allow Refinery to set its GOMAXPROC based on the container CPU quota if it exists. And then we can configure the number of decoder to be the number of CPUs

VinozzZ:
>I worry that automaxprocs won't be enough. The author of the zstd library is very concerned about maximum throughput in a streaming environment, but that's not our problem here, so I don't think it matters so much if we have a bunch of them

I'm also worry that if we max out CPU on decoding request payload during a spike of traffic, it would OOMKill Refinery sooner than what we expected.

I'm going to try running with 10 in the load test environment to see if it makes a difference

VinozzZ:
Unfortunately, I don't see a significant performance improvement from this change. I'm going to revert it in kibble and dogfood

PR #1573, opened 2025-06-03
fix: use sensible default value for HealthCheckTimeout and MaxExpiredTraces

VinozzZ:

The current default configuration in Refinery sets:

* HealthCheckTimeout = 3s
* MaxExpiredTraces = 5000

During load testing, we observed that processing a single expired trace through the sampler and flushing logic takes about 5ms. If Refinery were to actually process the full batch of 5000 traces in one sendExpiredTraces iteration, it would take about 25s. This processing time far exceeds the 3s default for HealthCheckTimeout

- Decrease the default value for `MaxExpiredTraces` to 3000
- Adjust `HealthCheckTimeout` value based on `MaxExpiredTraces`
- Increase default `HealthCheckTimeout` to 15s

### DIFF
--- a/config.md
+++ b/config.md
@@ -3,7 +3,7 @@
 # Honeycomb Refinery Configuration Documentation
 
 This is the documentation for the configuration file for Honeycomb's Refinery.
-It was automatically generated on 2025-05-23 at 20:21:03 UTC.
+It was automatically generated on 2025-06-03 at 21:32:34 UTC.
 
 ## The Config file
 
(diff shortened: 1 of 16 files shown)


config.md
```diff
@@ -391,12 +391,14 @@ MaxExpiredTraces is the maximum number of expired traces to process.
 This setting controls how many traces are processed when it is time to make a sampling decision.
 Up to this many traces will be processed every `SendTicker` duration.
 If this number is too small it will mean Refinery is spending less time calculating sampling decisions, resulting in data arriving at Honeycomb slower.
+Additionally, MaxExpiredTraces indirectly affects the system’s health check behavior.
+The HealthCheckTimeout will be automatically adjusted based on this value to ensure health checks remain accurate relative to the configured processing load.
 If your `collector_collect_loop_duration_ms` is above 3 seconds it is recommended to reduce this value and the `SendTicker` duration.
```
kentquirk, line 396: Do we want to change this advice?

config/file_config.go
```diff
@@ -871,6 +871,24 @@ func (f *fileConfig) GetSamplerConfigForDestName(destname string) (any, string)
...
+	configured := time.Duration(f.mainConfig.Collection.HealthCheckTimeout)
+	var timeout time.Duration
+	if f.mainConfig.Traces.MaxExpiredTraces != 0 {
+		// assume each traces take 5ms to process, so we can use that
+		// as a timeout for the health check.
+		timeout = time.Duration(f.mainConfig.Traces.MaxExpiredTraces) * time.Millisecond * 5
+	}
+
+	if configured < timeout {
+		return timeout
```
robbkidd, line 887: What do you think of a log message (INFO? WARN?) when Refinery has determined a higher HealthCheckTimeout than what was configured?

VinozzZ: Hmm, that's a good point. We don't have access to logger here. Maybe the caller can just log out the value?

robbkidd: The caller could, but wouldn't be able to include any statement about whether the timeout was configured or computed. 🤔

VinozzZ: That is fair. I believe it's ok. If the healthcheck failed due to timeout, we can find the log about the current timeout value through logs. I believe users shouldn't care about this unless there's a healthcheck failure. I have updated the error log in the healthcheck code to include the timeout value

PR #1574, opened 2025-06-04
feat: only send decision spans when needed

MikeGoldsmith:

-

-

### DIFF
--- a/collect/collect.go
+++ b/collect/collect.go
@@ -539,35 +539,9 @@ func (i *InMemCollector) redistributeTraces(ctx context.Context) {
 			continue
 		}
 
-		for _, sp := range trace.GetSpans() {
-
-			sp.SetSendBy(trace.SendBy)
-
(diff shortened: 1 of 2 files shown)


collect/collect.go
```diff
@@ -543,7 +545,12 @@ func (i *InMemCollector) redistributeTraces(ctx context.Context) {
 
 			if !i.Config.GetCollectionConfig().TraceLocalityEnabled() {
 				dc := i.createDecisionSpan(sp, trace, newTarget)
-				i.PeerTransmission.EnqueueEvent(dc)
+				// only forward spans that contain fields that can be used
+				// by a sampler to make a decision
+				if dc.GetDataSize() > 0 {
```
VinozzZ: I believe `GetDataSize()` will always be greater than 0 since there's metadata added in the decision spans

MikeGoldsmith: Ah, you're right. I read it as though it would only contain fields that the sample needed but actually is a combination of some metadata fields and sampler fields. Thanks.

PR #1575, opened 2025-06-04
fix: TraceKey used in dynamic sampler should not create a new key if no key fields are found

VinozzZ:

Currently, if no key fields are found, the dynamic sampler will create a new key with ,+trace_length

- return an empty string as the key if no key fields are found

Open question:
- Should we still sample the trace through a dynamic sampler if no key fields are found? Should the trace simply be dropped?


sample/trace_key.go
```diff
@@ -100,7 +105,7 @@ outer:
 		}
 	}
 
-	if d.useTraceLength {
+	if d.useTraceLength && fieldCount > 0 {
```
kentquirk: I feel like if the user has specified to use trace length, that should be respected even if the other fields are empty. It's an edge case, but it feels semi-plausible?

VinozzZ: Then they would be essentially creating keys based on trace length. I imagine that would be a very high cardinality field

kentquirk: That's the point of that flag; if they use it, the keys have at least the cardinality of trace length. But there are predictable systems where that might be a good idea. I.e., most traces are either 10 or 14 spans, but if there's an issue you might get a 100-span trace -- you want to keep those.

VinozzZ: I see. That makes sense. I will revert this change

kentquirk:
I think the behavior here will be that if a trace ends up with a blank key, we don't count them and would always use 1 for the sample rate. In other words, it treats every trace that doesn't have a key as special. 

But is that what users actually want? I feel like expected behavior is that if a trace doesn't have any key fields, we should put those traces into a single bucket all together. I could imagine a situation where they want to EMAThroughput sample all the endpoints, and anything that doesn't match an endpoint might actually be a large fraction of traces. They would have a hard time reaching their target sample rates.

What problem are you trying to fix with this?

VinozzZ:
> I think the behavior here will be that if a trace ends up with a blank key, we don't count them and would always use 1 for the sample rate. In other words, it treats every trace that doesn't have a key as special.
> 
> But is that what users actually want? I feel like expected behavior is that if a trace doesn't have any key fields, we should put those traces into a single bucket all together. I could imagine a situation where they want to EMAThroughput sample all the endpoints, and anything that doesn't match an endpoint might actually be a large fraction of traces. They would have a hard time reaching their target sample rates.
> 
> What problem are you trying to fix with this?

I believe the behavior here is all traces without any key fields on them will be put into the blank trace key bucket.

Maybe my understanding of how dynamic sampler should work is not correct. Based on current documentation, it looks like only traces that contains key fields are handed to the sampler. When I was doing testing, I was surprised to see an empty key as `,,,` present. I put up this to make it explicit that if it's an empty key, it should be an empty string. Maybe the doc can also be included to call out that traces that doesn't contain any key fields are also sampled through a single bucket

kentquirk (approved):
Thanks for putting up with my pickiness!

VinozzZ:
> Thanks for putting up with my pickiness!

I appreciate your thoroughness!!

PR #1580, opened 2025-06-05
test: improve existing integration benchmark, add new test and benchmark for Router.batch()

ianwilkes:
We need benchmarks which roughly approximate production bottlenecks in order to evaluate the effectiveness of optimizations.

Improves the pre-existing integration benchmarks to be more internally efficient, and use msgpack+zstd which is more typical today. Note these still output json which isn't typical in production so that still needs to be fixed.

Also adds a new test and benchmark for Router.batch() which is the primary consumer of both CPU and memory, at least in our internal cluster.

### DIFF
--- a/route/route_test.go
+++ b/route/route_test.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"compress/gzip"
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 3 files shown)


route/route_test.go
```diff
@@ -923,3 +924,201 @@ func TestProcessEventMetrics(t *testing.T) {
...
+	w := &discardResponseWriter{}
+
+	b.ResetTimer()
+	for b.Loop() {
+		req.Body = io.NopCloser(bytes.NewReader(batchMsgpack))
+
+		router.batch(w, req)
+
+		// Drain spans to prevent channel from filling up
+		for len(mockCollector.Spans) > 0 {
```
VinozzZ, line 1120: Should this draining logic be outside of `b.Loop()` so that it's not included in the benchmark result?

VinozzZ (approved):
LGTM. I just have one question about the router's benchmark test

route/route_test.go
ianwilkes: It will have a negligible impact on the benchmark results - pulling a few items out of the a channel is very fast. You see this kind of thing a lot in benchmarks where it's just too difficult to extract some piece of cleanup logic. It usually only matters for very tight benchmarks of very fast operations, which this definitely is not.

Also, arguably the cost of pulling events from the channel is part of the cost of this code. In a more intensely optimized environment you might find that the channel overhead itself was a bottleneck and improve efficiency by passing batches of events through the channel instead, as retriever does during queries.

PR #1582, opened 2025-06-09
test: add benchmark test for collect loop

VinozzZ:

We would like to measure performance of different sampling configurations with various span count per trace to identify bottlenecks and optimize the collector's performance.

- Added `BenchmarkCollectorWithSamplers` to test the performance of different sampler configurations:
  - Deterministic sampler
  - Dynamic sampler
  - EMA dynamic sampler
  - Rules-based sampler

- Implemented matrix testing across three workload scenarios:
  - Small traces (100 spans per trace)
  - Medium traces (10,000 spans per trace)
  - Large traces (1,000,000 spans per trace)

### DIFF
--- a/collect/collect_test.go
+++ b/collect/collect_test.go
@@ -7,6 +7,7 @@ import (
 	"runtime"
 	"strconv"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
@@ -2562,7 +2563,7 @@ func BenchmarkCollectorWithSamplers(b *testing.B) {
 	}{
 		{"small_traces", 100},
 		{"medium_traces", 10000},
-		{"large_traces", 1000000},
+		{"large_traces", 1_000_000},
 	}
 
 	// Different sampler configurations to test
@@ -2571,18 +2572,18 @@ func BenchmarkCollectorWithSamplers(b *testing.B) {
 		config interface{}
 	}{
 		{
-			"deterministic_sampler",
+			"deterministic",
 			&config.DeterministicSamplerConfig{SampleRate: 1},
 		},
 		{
-			"dynamic_sampler",
+			"dynamic",
 			&config.DynamicSamplerConfig{
 				SampleRate: 1,
 				FieldList:  []string{"sampler-field-1", "sampler-field-2"},
 			},
 		},
 		{
-			"ema_dynamic_sampler",
+			"emadynamic",
 			&config.EMADynamicSamplerConfig{
 				GoalSampleRate: 1,
 				FieldList:      []string{"sampler-field-1", "sampler-field-2"},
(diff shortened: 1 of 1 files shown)


collect/collect_test.go
```diff
@@ -2552,3 +2552,210 @@ func TestSpanLimitSendByPreservation(t *testing.T) {
...
+// BenchmarkCollectorWithSamplers runs benchmarks for different sampler configurations
+func BenchmarkCollectorWithSamplers(b *testing.B) {
+	// Common test scenarios to run for each sampler
+	scenarios := []struct {
+		name          string
+		spansPerTrace int
+	}{
+		{"small_traces", 100},
+		{"medium_traces", 10000},
+		{"large_traces", 1000000},
```
ianwilkes: nit: it's easier to read numbers like this with separators: `1_000_000`

collect/collect_test.go
```diff
@@ -2552,3 +2552,210 @@ func TestSpanLimitSendByPreservation(t *testing.T) {
...
+	}
+
+	// Different sampler configurations to test
+	samplerConfigs := []struct {
+		name   string
+		config interface{}
+	}{
+		{
+			"deterministic_sampler",
```
ianwilkes: The full benchmark identifiers are awfully long there. Since samplers are kind of implied here, how about remove the `_sampler` suffix from all these and maybe even rename the function to just `BenchmarkCollector`

collect/collect_test.go
```diff
@@ -2552,3 +2552,210 @@ func TestSpanLimitSendByPreservation(t *testing.T) {
...
+				collector := setupBenchmarkCollector(b, sampler.config, sender)
+				defer collector.Stop()
+
+				b.StopTimer()
+
+				// use b.N as the number of traces to create
+				totalSpans := b.N * scenario.spansPerTrace
+
+				// Setup done channel that waits for all spans to be processed
+				done := make(chan struct{})
```
ianwilkes: Conventionally when you're just waiting for completion, use a `sync.WaitGroup` instead of a channel. It does the same thing but the intention is easier to read.

collect/collect_test.go
```diff
@@ -2552,3 +2552,210 @@ func TestSpanLimitSendByPreservation(t *testing.T) {
...
+}
+
+func (c *mockSender) Flush() {
+	// No-op for mock sender
+}
+
+func (c *mockSender) RegisterMetrics() {
+	// No-op for mock sender
+}
+func (c *mockSender) waitForCount(b *testing.B, target int) {
```
ianwilkes: spacing

collect/collect_test.go
```diff
@@ -2552,3 +2552,210 @@ func TestSpanLimitSendByPreservation(t *testing.T) {
...
+			benchName := fmt.Sprintf("%s/%s", sampler.name, scenario.name)
+			b.Run(benchName, func(b *testing.B) {
+
+				sender := &mockSender{
+					eventQueue: make(chan *types.Event, 10000),
+				}
+				collector := setupBenchmarkCollector(b, sampler.config, sender)
+				defer collector.Stop()
+
+				b.StopTimer()
```
ianwilkes: doesn't matter much, but move this to the top of this function

PR #1583, opened 2025-06-09
fix: use noop span when tracing is disabled

VinozzZ:

The go.opentelemetry.io/otel/trace/noop tracer still allocates by copying the parent context and setting the noop span as the current span. While this behavior is generally fine, it introduces unnecessary allocations in processSpan, which is a hot path in our trace ingestion pipeline.

These context allocations are significant enough to appear in memory profiles during benchmarking. 

- return a global noop span if tracing is not enabled
- remove some not crucial instrumentation from `processSpan` function


ianwilkes (approved):
Annoyingly the go otel SDK does have a `nonRecordingSpan` type which is really what you'd want here, but there's no way to make one manually. The implementation is pretty small though so if we end up doing this in more than one place, it might be worth copying over so we can use it, rather that endless populating this `noopSpan` which contains internal locks which may slow us down.

PR #1584, opened 2025-06-09
fix: only calculate span count and annotation type once

VinozzZ:

Benchmarking revealed that AnnotationType was being called excessively — once for every span, per count type. This happens because each span metadata calculation iterates through the entire span list and re-evaluates the annotation type for each span.

This PR improves performance by calculating the annotation type once per span and caching the result. Subsequent metadata calculations reuse the cached value.

- cache `AnnotationType` value on each span
- cache `spanCount`, `spanEventCount`, and `spanLinkCount` on traces

Before:
Screenshot 2025-06-09 at 6 47 22 PM

After:

Screenshot 2025-06-09 at 6 54 32 PM

### DIFF
--- a/types/event.go
+++ b/types/event.go
@@ -114,6 +114,10 @@ type Trace struct {
 	// This is used to memoize the impact calculation so that it doesn't get
 	// calculated over and over during a sort.
 	totalImpact int
+
+	spanCount      uint32
+	spanEventCount uint32
+	spanLinkCount  uint32
 }
 
 // AddSpan adds a span to this trace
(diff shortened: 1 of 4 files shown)


types/event.go
```diff
@@ -171,47 +175,48 @@ func (t *Trace) SetKeptReason(reason uint) {
 	t.keptReason = reason
 }
 
+func (t *Trace) spanCounts() (spanCount, spanEventCount, spanLinkCount uint32) {
```
MikeGoldsmith: Having an internal and external func with the same name can be confusing. I think using a verb, like 
`calculate`, make it more readable as you know it's going to do something. It could be something else like `get`.

```suggestion
func (t *Trace) calculateSpanCounts() (spanCount, spanEventCount, spanLinkCount uint32) {
```

types/event.go
```diff
@@ -324,25 +330,32 @@ func (sp *Span) GetDataSize() int {
...
 	// SpanAnnotationTypeSpanEvent is the type for a span event.
 	SpanAnnotationTypeSpanEvent
 	// SpanAnnotationTypeLink is the type for a span link.
 	SpanAnnotationTypeLink
 )
 
-// AnnotationType returns the type of annotation this span is.
+// GetSpanAnnotationType returns the type of annotation this span is.
 func (sp *Span) AnnotationType() SpanAnnotationType {
+	if sp.annotationType != SpanAnnotationTypeUnSet {
```
MikeGoldsmith, line 360: An alternative to doing a lazy load of annotation type, we could set it when we first create the span type but I don't think it would be much different from this performance wise.

MikeGoldsmith (approved):
Looks good! ~45% reduction in average time is huge.

I've added one small suggestion on func name but not blocking.

types/event.go
VinozzZ: I thought about doing that too. The reason why I decided not to do it is because only kept traces need to know about annotation type. If we do it when we first create the span type, it will be done for both kept and dropped spans

ianwilkes (commented):
One artifact of this approach is that calling AnnotationType() and the various trace methods which use it are no longer concurrency-safe - is it guaranteed that these will only be called on a given span/trace from a single goroutine? If yes, document this restriction. If no, then I think pre-setting the values is probably safest.

Also, question about how this is used, the AnnotationType() method itself is looking for string field values and returning numeric constants, but there are a couple of places in the code (like ExtractDecisionContext()) where we set the field value with the numeric code directly. Is that intentional or is that a bug?

VinozzZ:
> One artifact of this approach is that calling AnnotationType() and the various trace methods which use it are no longer concurrency-safe - is it guaranteed that these will only be called on a given span/trace from a single goroutine? If yes, document this restriction. If no, then I think pre-setting the values is probably safest.
> 
> Also, question about how this is used, the AnnotationType() method itself is looking for string field values and returning numeric constants, but there are a couple of places in the code (like ExtractDecisionContext()) where we set the field value with the numeric code directly. Is that intentional or is that a bug?

I believe the Span object itself is not expected to be used by multiple goroutines. I can document this restriction

The numeric value of `AnnotationType` is never exported to honeycomb. The DecisionContext is only used for peer communication. I think it's safe to use the numeric code directly

types/event.go
```diff
@@ -171,47 +175,48 @@ func (t *Trace) SetKeptReason(reason uint) {
 	t.keptReason = reason
 }
 
+func (t *Trace) calculateSpanCounts() (spanCount, spanEventCount, spanLinkCount uint32) {
```
ianwilkes: Given that this is only used to set the three fields, you could have this method do that directly and skip the return values, which would be marginally less code/complexity.

ianwilkes (approved):
OK

PR #1585, opened 2025-06-11
perf: optimize trace key generation with hash-based deduplication

VinozzZ:

The trace key generation process was creating unnecessary memory allocations and putting pressure on garbage collection due to excessive string creation. This was particularly problematic for large traces with many spans that contains the same value for key fields.

- Replaced string concatenation and `fmt.Sprintf` operations with a reusable byte buffer to minimize allocations
-  use uint64 values as map keys instead of strings
- add unit test

## Benchmark result
This is the benchmark result from running the `collect` loop which exercises this function
```
goos: darwin
goarch: arm64
pkg: github.com/honeycombio/refinery/collect
cpu: Apple M2 Max
                                                     │   old.txt    │          new-trace-key.txt          │
                                                     │    sec/op    │   sec/op     vs base                │
CollectorWithSamplers/deterministic/small_traces-12     79.72µ ± 3%   79.82µ ± 2%        ~ (p=0.971 n=10)
CollectorWithSamplers/deterministic/medium_traces-12    8.009m ± 4%   7.960m ± 3%        ~ (p=0.481 n=10)
CollectorWithSamplers/deterministic/large_traces-12     939.1m ± 3%   928.8m ± 3%        ~ (p=0.739 n=10)
CollectorWithSamplers/dynamic/small_traces-12          103.20µ ± 3%   87.85µ ± 4%  -14.87% (p=0.000 n=10)
CollectorWithSamplers/dynamic/medium_traces-12          9.409m ± 3%   8.266m ± 4%  -12.15% (p=0.000 n=10)
CollectorWithSamplers/dynamic/large_traces-12            1.173 ± 2%    1.063 ± 5%   -9.37% (p=0.000 n=10)
CollectorWithSamplers/emadynamic/small_traces-12       100.94µ ± 3%   81.08µ ± 3%  -19.67% (p=0.000 n=10)
CollectorWithSamplers/emadynamic/medium_traces-12       9.328m ± 2%   7.825m ± 2%  -16.11% (p=0.000 n=10)
CollectorWithSamplers/emadynamic/large_traces-12         1.153 ± 1%    1.019 ± 2%  -11.59% (p=0.000 n=10)
CollectorWithSamplers/rulesbased/small_traces-12        74.59µ ± 1%   75.36µ ± 7%        ~ (p=0.123 n=10)
CollectorWithSamplers/rulesbased/medium_traces-12       7.383m ± 1%   7.136m ± 2%   -3.34% (p=0.000 n=10)
CollectorWithSamplers/rulesbased/large_traces-12        927.6m ± 3%   922.9m ± 2%        ~ (p=0.218 n=10)
geomean                                                 9.222m        8.519m        -7.62%

                                                     │    old.txt    │          new-trace-key.txt           │
                                                     │     B/op      │     B/op      vs base                │
CollectorWithSamplers/deterministic/small_traces-12     6.193Ki ± 0%   6.194Ki ± 0%        ~ (p=0.700 n=10)
CollectorWithSamplers/deterministic/medium_traces-12    617.2Ki ± 0%   617.2Ki ± 0%        ~ (p=0.382 n=10)
CollectorWithSamplers/deterministic/large_traces-12     73.44Mi ± 0%   73.43Mi ± 0%        ~ (p=0.247 n=10)
CollectorWithSamplers/dynamic/small_traces-12          20.352Ki ± 0%   7.224Ki ± 0%  -64.50% (p=0.000 n=10)
CollectorWithSamplers/dynamic/medium_traces-12         1482.3Ki ± 0%   618.3Ki ± 0%  -58.28% (p=0.000 n=10)
CollectorWithSamplers/dynamic/large_traces-12          157.37Mi ± 0%   73.45Mi ± 0%  -53.33% (p=0.000 n=10)
CollectorWithSamplers/emadynamic/small_traces-12       20.461Ki ± 0%   7.306Ki ± 0%  -64.29% (p=0.000 n=10)
CollectorWithSamplers/emadynamic/medium_traces-12      1482.3Ki ± 0%   618.4Ki ± 0%  -58.29% (p=0.000 n=10)
CollectorWithSamplers/emadynamic/large_traces-12       157.37Mi ± 0%   73.44Mi ± 0%  -53.33% (p=0.000 n=10)
CollectorWithSamplers/rulesbased/small_traces-12        6.983Ki ± 0%   6.983Ki ± 0%        ~ (p=0.954 n=10)
CollectorWithSamplers/rulesbased/medium_traces-12       618.0Ki ± 0%   617.9Ki ± 0%   -0.00% (p=0.001 n=10)
CollectorWithSamplers/rulesbased/large_traces-12        73.44Mi ± 0%   73.44Mi ± 0%   -0.00% (p=0.043 n=10)
geomean                                                 1.044Mi        684.9Ki       -35.91%

                                                     │   old.txt   │           new-trace-key.txt           │
                                                     │  allocs/op  │  allocs/op   vs base                  │
CollectorWithSamplers/deterministic/small_traces-12     326.0 ± 0%    326.0 ± 0%        ~ (p=1.000 n=10) ¹
CollectorWithSamplers/deterministic/medium_traces-12   30.05k ± 0%   30.05k ± 0%        ~ (p=1.000 n=10)
CollectorWithSamplers/deterministic/large_traces-12    3.002M ± 0%   3.002M ± 0%        ~ (p=0.469 n=10)
CollectorWithSamplers/dynamic/small_traces-12           769.0 ± 0%    355.0 ± 0%  -53.84% (p=0.000 n=10)
CollectorWithSamplers/dynamic/medium_traces-12         70.10k ± 0%   30.09k ± 0%  -57.08% (p=0.000 n=10)
CollectorWithSamplers/dynamic/large_traces-12          7.002M ± 0%   3.002M ± 0%  -57.13% (p=0.000 n=10)
CollectorWithSamplers/emadynamic/small_traces-12        771.0 ± 0%    357.0 ± 0%  -53.70% (p=0.000 n=10)
CollectorWithSamplers/emadynamic/medium_traces-12      70.11k ± 0%   30.09k ± 0%  -57.08% (p=0.000 n=10)
CollectorWithSamplers/emadynamic/large_traces-12       7.002M ± 0%   3.002M ± 0%  -57.13% (p=0.000 n=10)
CollectorWithSamplers/rulesbased/small_traces-12        335.0 ± 0%    335.0 ± 0%        ~ (p=1.000 n=10) ¹
CollectorWithSamplers/rulesbased/medium_traces-12      30.06k ± 0%   30.06k ± 0%   -0.00% (p=0.005 n=10)
CollectorWithSamplers/rulesbased/large_traces-12       3.002M ± 0%   3.002M ± 0%   -0.00% (p=0.050 n=10)
geomean                                                47.35k        31.40k       -33.68%
```

benchmark result from trace_key.build function itself
```
goos: darwin
goarch: arm64
pkg: github.com/honeycombio/refinery/sample
cpu: Apple M2 Max
                                         │ old-trace-key-build.txt │       new-trace-key-build.txt       │
                                         │         sec/op          │   sec/op     vs base                │
TraceKeyBuild/small_trace_fields-12                   19.736µ ± 3%   4.659µ ± 1%  -76.40% (p=0.000 n=10)
TraceKeyBuild/large_trace_fields-12                   191.90m ± 3%   73.89m ± 1%  -61.49% (p=0.000 n=10)
TraceKeyBuild/small_trace_many_fields-12               56.11µ ± 2%   13.98µ ± 1%  -75.08% (p=0.000 n=10)
TraceKeyBuild/with_root_fields-12                     19.697µ ± 2%   4.850µ ± 0%  -75.38% (p=0.000 n=10)
TraceKeyBuild/high_cardinality-12                      19.52µ ± 1%   10.04µ ± 0%  -48.57% (p=0.000 n=10)
geomean                                                152.2µ        47.20µ       -68.99%

                                         │ old-trace-key-build.txt │        new-trace-key-build.txt        │
                                         │          B/op           │     B/op      vs base                 │
TraceKeyBuild/small_trace_fields-12                   13809.0 ± 0%     515.0 ± 0%   -96.27% (p=0.000 n=10)
TraceKeyBuild/large_trace_fields-12                88005373.0 ± 0%     652.0 ± 1%  -100.00% (p=0.000 n=10)
TraceKeyBuild/small_trace_many_fields-12             30.717Ki ± 0%   1.031Ki ± 0%   -96.64% (p=0.000 n=10)
TraceKeyBuild/with_root_fields-12                     14370.0 ± 0%     595.0 ± 0%   -95.86% (p=0.000 n=10)
TraceKeyBuild/high_cardinality-12                    13.284Ki ± 0%   1.383Ki ± 0%   -89.59% (p=0.000 n=10)
geomean                                               92.13Ki          785.3        -99.17%

                                         │ old-trace-key-build.txt │       new-trace-key-build.txt       │
                                         │        allocs/op        │ allocs/op   vs base                 │
TraceKeyBuild/small_trace_fields-12                    434.00 ± 0%   17.00 ± 0%   -96.08% (p=0.000 n=10)
TraceKeyBuild/large_trace_fields-12                4000035.00 ± 0%   18.00 ± 0%  -100.00% (p=0.000 n=10)
TraceKeyBuild/small_trace_many_fields-12              1284.00 ± 0%   47.00 ± 0%   -96.34% (p=0.000 n=10)
TraceKeyBuild/with_root_fields-12                      437.00 ± 0%   19.00 ± 0%   -95.65% (p=0.000 n=10)
TraceKeyBuild/high_cardinality-12                       319.0 ± 0%   101.0 ± 0%   -68.34% (p=0.000 n=10)
geomean                                                3.151k        30.77        -99.02%
```

### DIFF
--- a/sample/trace_key.go
+++ b/sample/trace_key.go
@@ -116,9 +116,9 @@ type distinctValue struct {
 	fields []string
 	values []map[uint64]string
 
-	// totalUniqueCount keeps track of how many unique values we've seen so far
+	// totalUniqueCount keeps track of how many unique values we've seen so far for a trace key.
 	totalUniqueCount int
-	// maxDistinctValue is the maximum number of distinct values we will store
+	// maxDistinctValue is the maximum number of distinct values we will store for a trace key.
 	maxDistinctValue int
 }
 
--- a/sample/trace_key.go
+++ b/sample/trace_key.go
@@ -57,9 +57,7 @@ func (d *traceKey) build(trace *types.Trace) (string, int) {
 	spans := trace.GetSpans()
 	uniques := d.distinctValue
 	uniques.Reset(d.fields, maxKeyLength)
-	defer func() {
-		d.keyBuilder.Reset()
-	}()
+	d.keyBuilder.Reset()
 outer:
 	for i, field := range d.fields {
 		for _, span := range spans {
--- a/sample/trace_key.go
+++ b/sample/trace_key.go
@@ -1,13 +1,11 @@
 package sample
 
 import (
-	"encoding/binary"
+	"bytes"
 	"fmt"
-	"math"
 	"sort"
 	"strconv"
 	"strings"
-	"sync"
 
 	"github.com/dgryski/go-metro"
 	"github.com/honeycombio/refinery/types"
@@ -23,6 +21,8 @@ type traceKey struct {
 	fields         []string
 	rootOnlyFields []string
 	useTraceLength bool
+	keyBuilder     *bytes.Buffer
+	distinctValue  *distinctValue
 }
 
 func newTraceKey(fields []string, useTraceLength bool) *traceKey {
@@ -43,6 +43,8 @@ func newTraceKey(fields []string, useTraceLength bool) *traceKey {
 		fields:         nonRootFields,
 		rootOnlyFields: rootOnlyFields,
 		useTraceLength: useTraceLength,
+		distinctValue:  &distinctValue{buf: make([]byte, 0, 1024)},
+		keyBuilder:     &bytes.Buffer{},
 	}
 }
 
@@ -53,12 +55,12 @@ func (d *traceKey) build(trace *types.Trace) (string, int) {
 
 	// for each field, for each span, get the value of that field
 	spans := trace.GetSpans()
-	uniques := distinctValuePool.Get().(*distinctValue)
+	uniques := d.distinctValue
 	uniques.init(d.fields, maxKeyLength)
 	defer func() {
 		// reset the distinctValue for reuse
 		uniques.Reset()
-		distinctValuePool.Put(uniques)
+		d.keyBuilder.Reset()
 	}()
 outer:
 	for i, field := range d.fields {
@@ -77,7 +79,8 @@ outer:
 	// ok, now we have a map of fields to a list of all unique values for that field.
 	// (unless it was huge, in which case we have a bunch of them)
 
-	var key strings.Builder
+	// please change this so that strings.Builder can be reused between calls
+
 	for i := range d.fields {
 		values := uniques.Values(i)
 		// if there's no values for this field, skip it
@@ -87,40 +90,31 @@ outer:
 		var prevStr string
 		for _, str := range values {
 			if str != prevStr {
-				key.WriteString(str)
-				key.WriteRune('•')
+				d.keyBuilder.WriteString(str)
+				d.keyBuilder.WriteRune('•')
 				fieldCount += 1
 			}
 			prevStr = str
 		}
 		// get ready for the next element
-		key.WriteRune(',')
+		d.keyBuilder.WriteRune(',')
 	}
 
 	if trace.RootSpan != nil {
 		for _, field := range d.rootOnlyFields {
 			if val, ok := trace.RootSpan.Data[field]; ok {
-				key.WriteString(fmt.Sprintf("%v,", val))
+				d.keyBuilder.WriteString(fmt.Sprintf("%v,", val))
 				fieldCount += 1
 			}
 		}
 	}
 
 	if d.useTraceLength {
-		key.WriteString(strconv.FormatInt(int64(len(spans)), 10))
+		d.keyBuilder.WriteString(strconv.FormatInt(int64(len(spans)), 10))
 		fieldCount += 1
 	}
 
-	return key.String(), fieldCount
-}
-
-// Pool for reusing distinctValue objects.
-var distinctValuePool = &sync.Pool{
-	New: func() any {
-		return &distinctValue{
-			buf: make([]byte, 0, 1024),
-		}
-	},
+	return d.keyBuilder.String(), fieldCount
 }
 
 // distinctValue keeps track of distinct values for a set of fields.
@@ -217,17 +211,13 @@ func (d *distinctValue) AddAsString(value any, fieldIdx int) bool {
 	case string:
 		d.buf = append(d.buf, []byte(v)...)
 	case int:
-		d.buf = binary.BigEndian.AppendUint64(d.buf, uint64(v))
+		d.buf = strconv.AppendInt(d.buf, int64(v), 10)
 	case int64:
-		d.buf = binary.BigEndian.AppendUint64(d.buf, uint64(v))
+		d.buf = strconv.AppendInt(d.buf, v, 10)
 	case float64:
-		d.buf = binary.BigEndian.AppendUint64(d.buf, math.Float64bits(v))
+		d.buf = strconv.AppendFloat(d.buf, v, 'f', -1, 64)
 	case bool:
-		if v {
-			d.buf = append(d.buf, '1')
-		} else {
-			d.buf = append(d.buf, '0')
-		}
+		d.buf = strconv.AppendBool(d.buf, v)
 	default:
 		d.buf = append(d.buf, fmt.Sprintf("%v", v)...)
 	}
@@ -238,7 +228,7 @@ func (d *distinctValue) AddAsString(value any, fieldIdx int) bool {
 		if d.totalUniqueCount >= d.maxDistinctValue {
 			return false
 		}
-		d.values[fieldIdx][hash] = fmt.Sprintf("%v", value)
+		d.values[fieldIdx][hash] = string(d.buf)
 		return true
 	}
 
--- a/sample/trace_key.go
+++ b/sample/trace_key.go
@@ -74,8 +74,11 @@ outer:
 
 	var key strings.Builder
 	for i := range d.fields {
-		// if there's no values for this field, skip it
 		values := uniques.Values(i)
+		// if there's no values for this field, skip it
+		if len(values) == 0 {
+			continue
+		}
 		var prevStr string
 		for _, str := range values {
 			if str != prevStr {
--- a/sample/trace_key.go
+++ b/sample/trace_key.go
@@ -1,12 +1,16 @@
 package sample
 
 import (
+	"encoding/binary"
 	"fmt"
+	"maps"
+	"math"
+	"slices"
 	"sort"
 	"strconv"
 	"strings"
 
-	"github.com/honeycombio/refinery/generics"
+	"github.com/dgryski/go-metro"
 	"github.com/honeycombio/refinery/types"
 )
 
@@ -47,27 +51,21 @@ func newTraceKey(fields []string, useTraceLength bool) *traceKey {
 // returns the number of values used to build the key
 func (d *traceKey) build(trace *types.Trace) (string, int) {
 	fieldCount := 0
-	// fieldCollector gets all values from the fields listed in the config, even
-	// if they happen multiple times.
-	fieldCollector := make(map[string][]string)
 
 	// for each field, for each span, get the value of that field
 	spans := trace.GetSpans()
-	uniques := generics.NewSetWithCapacity[string](maxKeyLength)
+	uniques := newDistinctValue(d.fields, maxKeyLength)
 outer:
-	for _, field := range d.fields {
+	for i, field := range d.fields {
 		for _, span := range spans {
 			if val, ok := span.Data[field]; ok {
-				u := fmt.Sprintf("%s/%v", field, val)
 				// don't bother to add it if we've already seen it
-				if uniques.Contains(u) {
-					continue
-				}
-				uniques.Add(u)
-				if len(uniques) >= maxKeyLength {
+				if uniques.totalUniqueCount >= maxKeyLength {
 					break outer
 				}
-				fieldCollector[field] = append(fieldCollector[field], fmt.Sprintf("%v", val))
+				if uniques.AddAsString(val, i) {
+					continue
+				}
 			}
 		}
 	}
@@ -75,14 +73,9 @@ outer:
 	// (unless it was huge, in which case we have a bunch of them)
 
 	var key strings.Builder
-	for _, field := range d.fields {
+	for i := range d.fields {
 		// if there's no values for this field, skip it
-		values, ok := fieldCollector[field]
-		if !ok || len(values) == 0 {
-			continue
-		}
-		// sort and collapse list
-		sort.Strings(values)
+		values := uniques.Values(i)
 		var prevStr string
 		for _, str := range values {
 			if str != prevStr {
@@ -111,4 +104,93 @@ outer:
 	}
 
 	return key.String(), fieldCount
+}
+
+// distinctValue keeps track of distinct values for a set of fields.
+// It stores the unique values as strings.
+type distinctValue struct {
+	buf    []byte
+	fields []string
+	values []map[uint64]string
+
+	// totalUniqueCount keeps track of how many unique values we've seen so far
+	totalUniqueCount int
+	// maxDistinctValue is the maximum number of distinct values we will store
+	maxDistinctValue int
+}
+
+func newDistinctValue(fields []string, maxDistinctValue int) *distinctValue {
+	d := &distinctValue{
+		buf:              make([]byte, 0, 1024),
+		fields:           fields,
+		values:           make([]map[uint64]string, len(fields)),
+		maxDistinctValue: maxDistinctValue,
+	}
+
+	for i := range d.values {
+		d.values[i] = make(map[uint64]string, maxDistinctValue)
+	}
+
+	return d
+
+}
+
+// Values returns the distinct values for a given field index.
+// It returns a sorted slice of strings containing the unique values for that field.
+func (d *distinctValue) Values(fieldIdx int) []string {
+	if fieldIdx < 0 || fieldIdx >= len(d.values) {
+		return nil
+	}
+
+	// Get the map for the specified field index
+	valueMap := d.values[fieldIdx]
+	if len(valueMap) == 0 {
+		return nil
+	}
+
+	values := slices.Collect(maps.Values(valueMap))
+	sort.Strings(values)
+
+	return values
+}
+
+// AddAsString adds a value to the distinct values for a given field index.
+// It returns true if the value was added, false if it was already present or if the maxDistinctValue limit was reached.
+func (d *distinctValue) AddAsString(value any, fieldIdx int) bool {
+	if value == nil {
+		return false
+	}
+
+	d.buf = d.buf[:0] // reset the buffer for each new value
+
+	switch v := value.(type) {
+	case string:
+		d.buf = append(d.buf, []byte(v)...)
+	case int:
+		d.buf = binary.BigEndian.AppendUint64(d.buf, uint64(v))
+	case int64:
+		d.buf = binary.BigEndian.AppendUint64(d.buf, uint64(v))
+	case float64:
+		d.buf = binary.BigEndian.AppendUint64(d.buf, math.Float64bits(v))
+	case bool:
+		if v {
+			d.buf = append(d.buf, '1')
+		} else {
+			d.buf = append(d.buf, '0')
+		}
+	default:
+		d.buf = append(d.buf, fmt.Sprintf("%v", v)...)
+	}
+
+	hash := metro.Hash64(d.buf, 0)
+	if _, exists := d.values[fieldIdx][hash]; !exists {
+		d.totalUniqueCount++
+		if d.totalUniqueCount >= d.maxDistinctValue {
+			return false
+		}
+		d.values[fieldIdx][hash] = fmt.Sprintf("%v", value)
+		return true
+	}
+
+	return false
 }
--- a/sample/trace_key.go
+++ b/sample/trace_key.go
@@ -55,20 +55,16 @@ func (d *traceKey) build(trace *types.Trace) (string, int) {
 
 	// for each field, for each span, get the value of that field
 	spans := trace.GetSpans()
-	uniques := d.distinctValue
-	uniques.Reset(d.fields, maxKeyLength)
+	d.distinctValue.Reset(d.fields, maxKeyLength)
 	d.keyBuilder.Reset()
 outer:
 	for i, field := range d.fields {
 		for _, span := range spans {
 			if val, ok := span.Data[field]; ok {
-				// don't bother to add it if we've already seen it
-				if uniques.totalUniqueCount >= maxKeyLength {
+				if d.distinctValue.totalUniqueCount >= maxKeyLength {
 					break outer
 				}
-				if uniques.AddAsString(val, i) {
-					continue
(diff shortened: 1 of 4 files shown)


kentquirk (approved):
I really like this. 


One caution is that it limits the cardinality of any given field to 10. I'd like to at least put that value in a constant for clarity. I'm wondering if there's any situation where this becomes an unwanted breaking change. Probably not, because in those cases the combinatorics for the sampler key means we'll end up with too many buckets. However, we should note this in the release notes and say that we expect it to stablize sampler performance in some edge cases.

UPDATE: The above is WRONG, I misread tests as the real code. 

Also, it would be cool to have a benchmark you could run before/after to see the improvements. That's always good ammo for future changes like this.

sample/trace_key.go
```diff
@@ -47,42 +51,34 @@ func newTraceKey(fields []string, useTraceLength bool) *traceKey {
...
+	for i := range d.fields {
+		values := uniques.Values(i)
 		// if there's no values for this field, skip it
-		values, ok := fieldCollector[field]
-		if !ok || len(values) == 0 {
+		if len(values) == 0 {
 			continue
 		}
-		// sort and collapse list
-		sort.Strings(values)
```
TylerHelmuth, line 85: sorting was important before to ensure the list of values was always the same when possible. Does the new algorithm not need that assurance anymore?

TylerHelmuth: nvm I see the sort in `Values`

sample/trace_key.go
```diff
@@ -112,3 +108,92 @@ outer:
...
+		d.buf = append(d.buf, fmt.Sprintf("%v", v)...)
+	}
+
+	hash := metro.Hash64(d.buf, 0)
+	if _, exists := d.values[fieldIdx][hash]; !exists {
+		d.totalUniqueCount++
+		if d.totalUniqueCount >= d.maxDistinctValue {
+			return false
+		}
+		d.values[fieldIdx][hash] = fmt.Sprintf("%v", value)
```
TylerHelmuth: If we're trying to avoid calling `Sprintf`, is there a way we can use what we learned about `value` type fromt he switch statement above to avoid this call whenever possible?

VinozzZ: Regardless what type `value` is, we are going to format it into a string in order to compose the trace_key so this step is needed as long as we want to use a human readable sample key.
Technically, we could just use the uint64 hash directly to create the sample key, but that requires change to the dynsampler_go library. Or we can use the hash as a string and then only create the human readable sample key when a trace is sampled. 
I didn't go with that complexity because the most costly part is the `fmt.Sprintf` creating the map key for checking uniqueness. In this case, we are only calling `fmt.Sprintf` at most 100 times per trace_key

sample/trace_key.go
```diff
@@ -1,12 +1,16 @@
...
 	"fmt"
+	"maps"
+	"math"
+	"slices"
 	"sort"
 	"strconv"
 	"strings"
 
-	"github.com/honeycombio/refinery/generics"
+	"github.com/dgryski/go-metro"
```
ianwilkes, line 10: You'll want to run go mod tidy since this is now a direct dependency. Unfortunate that we depend on both this _and_ its main competitor xxhash.

sample/trace_key.go
```diff
@@ -112,3 +108,92 @@ outer:
...
+
+	// totalUniqueCount keeps track of how many unique values we've seen so far for a trace key.
+	totalUniqueCount int
+	// maxDistinctValue is the maximum number of distinct values we will store for a trace key.
+	maxDistinctValue int
+}
+
+func newDistinctValue(fields []string, maxDistinctValue int) *distinctValue {
+	d := &distinctValue{
+		buf:              make([]byte, 0, 1024),
```
ianwilkes: Given that this whole struct is short-lived, it itself is going to create a log of garbage, especially with a default 1KB buffer.

I suggest adding a Reset() method which clears all the maps and then using a sync.Pool for these, which will make the whole thing amortize out to zero allocation. You can set the fields slice and extend the values slice if the one in the pool isn't big enough.

sample/trace_key.go
```diff
@@ -112,3 +108,92 @@ outer:
...
+		d.buf = binary.BigEndian.AppendUint64(d.buf, math.Float64bits(v))
+	case bool:
+		if v {
+			d.buf = append(d.buf, '1')
+		} else {
+			d.buf = append(d.buf, '0')
+		}
+	default:
+		d.buf = append(d.buf, fmt.Sprintf("%v", v)...)
+	}
```
ianwilkes: I think a good goal here is to have identical behavior to the prior version, but this doesn't. Since it's just for the hash, I think it _usually_ won't matter, but previously integer `1` would I think be treated as identical to string `"1"`, but that's no longer the case here. This is the sort of change which has bitten us on retriever before, as customers scratch their heads over weird artifacts related to their own heterogenous types.

So it may be worth mimicing the behavior of "%v" here, which you should be able to do with `strconv.Append*` methods, which should be fast enough not to be a problem here.

Also the fact that you didn't have to change any existing tests implies that that the unique trace key behavior isn't well tested, now would be a good time to fix that.

sample/trace_key.go
```diff
@@ -112,3 +108,92 @@ outer:
...
+		return nil
+	}
+
+	// Get the map for the specified field index
+	valueMap := d.values[fieldIdx]
+	if len(valueMap) == 0 {
+		return nil
+	}
+
+	values := slices.Collect(maps.Values(valueMap))
```
ianwilkes: `slices.Collect` is going to allocate a new slice every time; if you take the pool approach, you could retain and re-use a slice buffer for this instead.

ianwilkes (changes requested):
Whoops, approve was the wrong button. I'm concerned about the behavioral change to how we determine uniqueness, especially cases where two values are considered identical but produce different trace keys. Definitely needs to see clear testing of exactly what is, and is not, an identical key when this code is presented w/ similar values of different types.

sample/trace_key.go
VinozzZ: Since this function is called within a single goroutine, it's ok to just reuse a single instance and call Reset() after each use

sample/trace_key.go
```diff
@@ -40,75 +43,194 @@ func newTraceKey(fields []string, useTraceLength bool) *traceKey {
...
-	fieldCollector := make(map[string][]string)
 
 	// for each field, for each span, get the value of that field
 	spans := trace.GetSpans()
-	uniques := generics.NewSetWithCapacity[string](maxKeyLength)
+	uniques := d.distinctValue
+	uniques.init(d.fields, maxKeyLength)
+	defer func() {
+		// reset the distinctValue for reuse
+		uniques.Reset()
```
ianwilkes: Seems like this should be implicit with init()

A convention I see a lot is to have Reset() take arguments to set up for the new task in one combined operation.

sample/trace_key.go
```diff
@@ -40,75 +43,194 @@ func newTraceKey(fields []string, useTraceLength bool) *traceKey {
...
-	// fieldCollector gets all values from the fields listed in the config, even
-	// if they happen multiple times.
-	fieldCollector := make(map[string][]string)
 
 	// for each field, for each span, get the value of that field
 	spans := trace.GetSpans()
-	uniques := generics.NewSetWithCapacity[string](maxKeyLength)
+	uniques := d.distinctValue
+	uniques.init(d.fields, maxKeyLength)
+	defer func() {
```
ianwilkes: My arbitrary preference is to do resets before using the thing, not afterward. Here you've got a little of both, which is awkward.

sample/trace_key.go
```diff
@@ -40,75 +43,194 @@ func newTraceKey(fields []string, useTraceLength bool) *traceKey {
...
+	values       []map[uint64]string
+	valuesBuffer []string
+
+	// totalUniqueCount keeps track of how many unique values we've seen so far for a trace key.
+	totalUniqueCount int
+	// maxDistinctValue is the maximum number of distinct values we will store for a trace key.
+	maxDistinctValue int
+}
+
+func (d *distinctValue) init(fields []string, maxDistinctValue int) {
```
ianwilkes: This is way more complicated than it needs to be. It looks like `distinctValue.fields` is read-only, which means you don't need your own copy at all, and should just be able to do `d.fields = fields`. (I really, really wish go had immutables to make this explicit - they did it for strings but refuse to do it for other types)

Also, when you _do_ need a local copy of a slice in a pre-allocated buffer, you can just do
```
mySlice = append(mySlice[:0], theirSlice...)
```
instead of all this.

sample/trace_key.go
```diff
@@ -40,75 +43,194 @@ func newTraceKey(fields []string, useTraceLength bool) *traceKey {
...
+
+	if len(d.values) < len(fields) {
+		// if we don't have enough values, allocate more
+		for i := len(d.values); i < len(fields); i++ {
+			d.values = append(d.values, make(map[uint64]string))
+		}
+	} else {
+		// if we have more values than fields, trim the excess
+		d.values = d.values[:len(fields)]
+	}
```
ianwilkes: Having separate Reset() and init() methods makes this hard to follow, it will be simpler if you merge them. Something like:
```
for len(d.values) < len(fields) {
  d.values = append(d.values, make(map[uint64]string))
}
d.values = d.values[:len(fields)]
for i := range d.values {
  clear(d.values[i])
}
```
(this way there are never nils in d.values)

Slightly sub-optimal because if the field length moves around a lot you end up re-creating a lot of maps, but I suspect we don't care. Does the field length change much? Ever?

sample/trace_key.go
```diff
@@ -40,75 +43,194 @@ func newTraceKey(fields []string, useTraceLength bool) *traceKey {
...
+
+func (d *distinctValue) Reset() {
+	// Reset the fields and values but do not reallocate them
+	for i := range d.fields {
+		d.fields[i] = ""
+	}
+	for i := range d.values {
+		if d.values[i] != nil {
+			for k := range d.values[i] {
+				delete(d.values[i], k)
```
ianwilkes: As I noted above you should use `clear()` - deleting entries from a map one by one is very slow.

sample/trace_key.go
```diff
@@ -40,75 +43,194 @@ func newTraceKey(fields []string, useTraceLength bool) *traceKey {
...
+
+	d.buf = d.buf[:0] // reset the buffer for each new value
+
+	switch v := value.(type) {
+	case string:
+		d.buf = append(d.buf, []byte(v)...)
+	case int:
+		d.buf = strconv.AppendInt(d.buf, int64(v), 10)
+	case int64:
+		d.buf = strconv.AppendInt(d.buf, v, 10)
```
ianwilkes, line 174: I _think_ you can do this in one case., casting an int64 to an int64 is allowed.
```
case int, int64:
```

sample/trace_key_test.go
```diff
@@ -160,3 +162,341 @@ func createTestTrace(t *testing.T, spans []testSpan) *types.Trace {
...
+		expected := []string{"apple", "banana", "cherry"}
+		require.EqualValues(t, expected, values)
+	})
+
+	t.Run("Type conversion", func(t *testing.T) {
+		values := dv.Values(1)
+		require.Len(t, values, 2)
+		require.Contains(t, values, "42")
+		require.Contains(t, values, "true")
+	})
```
ianwilkes: Since `values` is sorted you should be able to compare equality of the whole slice, which IMO is a more robust test.

sample/trace_key_test.go
```diff
@@ -160,3 +162,341 @@ func createTestTrace(t *testing.T, spans []testSpan) *types.Trace {
...
+			require.Equal(t, 1, dv.totalUniqueCount)
+
+			// All equal values should be considered duplicates
+			for i := 1; i < len(tt.equalValues); i++ {
+				added := dv.AddAsString(tt.equalValues[i], 0)
+				require.False(t, added,
+					"Value %v should be considered a duplicate of %v",
+					tt.equalValues[i], tt.equalValues[0])
+				require.Equal(t, 1, dv.totalUniqueCount,
+					"Count should remain 1 after adding duplicate %v", tt.equalValues[i])
```
ianwilkes: nit: this formatting is a mess. You have two choices: all one line, or all separate lines.
```
foo(
  a,
  b,
)
```

sample/trace_key_test.go
```diff
@@ -160,3 +162,341 @@ func createTestTrace(t *testing.T, spans []testSpan) *types.Trace {
...
+						}
+					}
+				}
+
+				trace.AddSpan(span)
+			}
+
+			b.ResetTimer()
+
+			for i := 0; i < b.N; i++ {
```
ianwilkes: `for b.Loop()`

sample/trace_key_test.go
```diff
@@ -160,3 +162,341 @@ func createTestTrace(t *testing.T, spans []testSpan) *types.Trace {
...
+	})
+
+	t.Run("Empty field", func(t *testing.T) {
+		// Field that has no values
+		values := dv.Values(2)
+		require.Nil(t, values)
+	})
+}
+
+func BenchmarkTraceKeyBuild(b *testing.B) {
```
ianwilkes, line 360: nit: generally nice to have benchmarks at the bottom, not mixed in w/ the tests

sample/trace_key_test.go
```diff
@@ -160,3 +162,341 @@ func createTestTrace(t *testing.T, spans []testSpan) *types.Trace {
...
+			}
+
+			b.ResetTimer()
+
+			for i := 0; i < b.N; i++ {
+				key, _ := generator.build(trace)
+				// Force evaluation to prevent the compiler from optimizing away the call
+				if len(key) == 0 {
+					b.Fatal("Empty key returned")
+				}
```
ianwilkes: I'm 99% certain you don't need to do this for a function call.

sample/trace_key_test.go
```diff
@@ -160,3 +162,341 @@ func createTestTrace(t *testing.T, spans []testSpan) *types.Trace {
...
+			wantSuccess: false,
+			wantCount:   5, // Unchanged
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got := dv.AddAsString(tt.value, tt.fieldIdx)
+			require.Equal(t, tt.wantSuccess, got)
+			require.Equal(t, tt.wantCount, dv.totalUniqueCount)
```
ianwilkes: This really isn't testing much. Maybe these cases could be merged into the other big test, with more robust checks on the results?

ianwilkes (changes requested):
Good, just some cleanup left. I think the tests are a little raw.

sample/trace_key.go
VinozzZ: When I put them into one case, `v` becomes a type `any`. I will need to do a type assertion again

sample/trace_key_test.go
VinozzZ: You are correct!

sample/trace_key.go
VinozzZ: The field length only changes when a sampling rules is reconfigured by users. I don't think it happens that often

sample/trace_key.go
```diff
@@ -40,75 +43,162 @@ func newTraceKey(fields []string, useTraceLength bool) *traceKey {
...
+				}
 			}
 		}
 	}
 	// ok, now we have a map of fields to a list of all unique values for that field.
 	// (unless it was huge, in which case we have a bunch of them)
 
-	var key strings.Builder
-	for _, field := range d.fields {
+	// please change this so that strings.Builder can be reused between calls
```
ianwilkes: ?

sample/trace_key.go
```diff
@@ -40,75 +43,162 @@ func newTraceKey(fields []string, useTraceLength bool) *traceKey {
...
 func (d *traceKey) build(trace *types.Trace) (string, int) {
 	fieldCount := 0
-	// fieldCollector gets all values from the fields listed in the config, even
-	// if they happen multiple times.
-	fieldCollector := make(map[string][]string)
 
 	// for each field, for each span, get the value of that field
 	spans := trace.GetSpans()
-	uniques := generics.NewSetWithCapacity[string](maxKeyLength)
+	uniques := d.distinctValue
```
ianwilkes: On further consideration I'm not sure assigning this to a local is a good practice. Just using d.distinctValue throughout is a good reminder that this is a re-used resource, and will also avoid a regression if we ever change this to a non-pointer type.

sample/trace_key.go
```diff
@@ -40,75 +43,162 @@ func newTraceKey(fields []string, useTraceLength bool) *traceKey {
...
+	for len(d.values) < len(fields) {
+		d.values = append(d.values, make(map[uint64]string))
+	}
+
+	d.values = d.values[:len(fields)]
+	for i := range d.values {
+		clear(d.values[i])
+	}
+
+	// Reset the total unique count
```
ianwilkes: These are some incredibly low-value comments. Just because LLMs want to put this stuff everywhere doesn't mean it's useful.

sample/trace_key_test.go
```diff
@@ -137,6 +140,332 @@ func TestKeyLimits(t *testing.T) {
...
+						continue
+					}
+					dv.AddAsString(v, fieldIdx)
+				}
+			}
+
+			// each field should have expected unique counts and values
+			expectedTotalCount := 0
+			for fieldIdx, expectedCount := range tt.expectedCounts {
+				require.Equal(t, expectedCount, len(dv.values[fieldIdx]), "Unexpected count for field index %d", fieldIdx)
```
ianwilkes: generally I would expect `assert` here, not `require`, since a failure here doesn't blow up the whole test.

sample/trace_key_test.go
```diff
@@ -137,6 +140,332 @@ func TestKeyLimits(t *testing.T) {
...
+				expectedTotalCount += expectedCount
+			}
+
+			require.Equal(t, expectedTotalCount, dv.totalUniqueCount, "Unexpected unique count")
+			for fieldIdx, expectedValues := range tt.expectedValues {
+				values := dv.Values(fieldIdx)
+				require.ElementsMatch(t, expectedValues, values, "Unexpected distinct values")
```
ianwilkes: I think you should go further here and assert that the values match what is produced by `fmt.Sprintf("%v")`, in addition to the expected values in the the test cases. In case we make a mistake there.

sample/trace_key.go
VinozzZ: Ugh, the rebase messed up some commits. I'm gna take another look and clean it up

sample/trace_key.go
VinozzZ: I'm pretty sure this is because the rebase I did messed up the code. cleaning it up now

PR #1586, opened 2025-06-13
perf: add lazy-loading MessagePack payload for events

ianwilkes:
Refinery is not efficient! In production profiles we often spend over 50% of CPU time allocating and freeing memory, and a good chunk of the rest deserializing messagepack data.

This is the first (and hopefully largest) change on the way to cutting down the amount of time refinery spends deserializing events. In place of the ubiquitous `map[string]any`, this keeps the original serialized data with the event throughout its lifetime, and interrogates it when necessary to retrieve individual fields. The theory here is that ultimately refinery needs to access only a small fraction of fields, and unpacking the rest into distinct scalars on the heap is unnecessary. This is an approach which has served us well with other services.

Added here are the initial container types needed to make this happen, and all the boilerplate interface changes that go with that. Unfortunately I had to bring in a new message pack library, because our old one doesn't know how to work with flat `[]byte` slices efficiently, but it's also not a drop-in replacement. (The old library also hasn't had a commit in two years and is looking like it may be abandoned.)

But, this PR itself does not improve performance. To see initial gains, we'll need to optimize how we do messagepack deserialization, and pre-request some of our fields instead of demanding them all one-by-one with Get(). And this PR is big enough as it is. Longer term we'll also want to leverage this pre-serialized data during re-transmission to peers, and handle otel data efficiently in some way (probably direct transcoding). For the record I think it would be better if we were using otel for refinery's outbound transmission, but that's too big a change if we want to get this done as well.

This is only one behavioral change here, which is that the event size calculation now uses the serialized data when available, which will definitely produce somewhat different numbers than what we do today. Since this didn't break any tests I'm assuming that ok.


types/event_test.go
```diff
@@ -57,16 +59,15 @@ func TestSpan_GetDataSizeSlice(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
+			sliceData := make([]any, tt.num)
+			for i := range tt.num {
```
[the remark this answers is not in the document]
ianwilkes: Wow.

transmit/transmit.go
```diff
@@ -97,7 +97,12 @@ func (d *DefaultTransmission) EnqueueEvent(ev *types.Event) {
 		WithString("api_host", ev.APIHost).
 		WithString("dataset", ev.Dataset).
 		Logf("transmit sending event")
-	libhEv := d.builder.NewEventSized(len(ev.Data))
+	// Calculate approximate size for the libhoney event
+	dataSize := 0
+	for range ev.Data.All() {
```
MikeGoldsmith: Can we use the updated `ev.GetDataSize()` that resolves back to the new Payload's `GetDataSize` here?

types/payload.go
```diff
@@ -0,0 +1,235 @@
...
+		p.memoizedFields = make(map[string]any, len(keys))
+	}
+
+	keysToFind := make(map[string]struct{}, len(keys))
+	for _, key := range keys {
+		if _, ok := p.memoizedFields[key]; !ok {
+			keysToFind[key] = struct{}{}
+		}
+	}
+	if len(keysToFind) == 0 {
```
MikeGoldsmith, line 58: Maybe move this check above map init so we don't create with 0 length.

types/payload.go
```diff
@@ -0,0 +1,235 @@
...
+				break
+			}
+			key := string(keyBytes)
+			p.memoizedFields[key] = value
+			keysFound++
+		}
+	}
+}
+
+func (p *Payload) Get(key string) any {
```
MikeGoldsmith, line 116: I think returning `(any, bool)` like we used to get from a map would make using this func easier. There are many occasions where we try to get a value then immediately do `value != nil`.

types/payload.go
```diff
@@ -0,0 +1,235 @@
...
+	p.memoizedFields[key] = value
+}
+
+func (p *Payload) IsEmpty() bool {
+	return len(p.memoizedFields) == 0 && p.msgpMap == nil
+}
+
+// All() allows easily iterating all values in the Payload, but this is very
+// NOT EFFICIENT relative to getting a subset of values using Get. Don't use
+// this in non-test code unless you have to other choice.
```
MikeGoldsmith, line 160:
```suggestion
// this in non-test code unless you have to other choice.
// We only expect this to happen when transmitting sampled events using Libhoney.
```

MikeGoldsmith (approved):
This looks really good. I think having Payload.Get return a exists bool would make using it easier.

I've left a couple of other small suggestions, none are blocking.

types/msgp_payload_map_test.go
```diff
@@ -0,0 +1,329 @@
...
+		skipCount++
+		// Don't call any Value methods - test that skipping works
+	}
+	assert.Equal(t, len(inputData), skipCount)
+
+	// Test skipping some, but not all values
+	iter, err = payloadMap.Iterate()
+	require.NoError(t, err)
+
+	clear(gotData)
```
VinozzZ, line 103: This is super cool. I didn't know go has a built-in clear function

types/payload.go
```diff
@@ -0,0 +1,235 @@
...
+	}
+
+	for {
+		keyBytes, _, err := iter.NextKey()
+		if err != nil {
+			break
+		}
+
+		if string(keyBytes) == key {
+			value, err := iter.ValueAny()
```
VinozzZ, line 135: What's the performance tradeoff here by just using ValueAny()? What about using the Value method for concrete types?

transmit/transmit.go
ianwilkes: Yes, good catch. GetDataSize() was a late addition.

types/payload.go
ianwilkes: I think that depends on whether we actually want `nil` values or not. In this PR I assumed we actually don't want them (honeycomb will discard them on ingest) so this essentially pretends they don't exist. I probably should have noted that as a behavior change in the PR description.

Switching to the two-return form of Get() implies that you can get back a `nil` value and should do something with it, but the code here doesn't seem to really consider that. So, do you have an opinion there? Is there some reason why we want to distinguish "value is null" from "value does not exist at all"? If they're semantically the same I'd rather not have both states. (And returning `exists = false` any time the value is `nil` departs from the map semantic which I think is worse.)

types/payload.go
ianwilkes: You pay the cost the first time you put the concrete type into the interface variable, and most of the refinery code is using `any` for field values anyway, so there's no benefit to using a concrete accessor at this stage. I'm hopeful that eventually we'll have a need for typed versions of Get(), especially for numbers.

types/payload.go
MikeGoldsmith: Right, we need to decide what the second return parameter would mean. I believe maps use it to indicate the key exists, not that the value was not nil so you'd also need to do a nil check against the value. Implementing a different behaviour than the way maps work would 100% cause confusion and misuse.

I agree we should not expect nil values and treat that as invalid. So yeah, let's stick with a single return param.

Maybe we consider adding an Exists func that stream lines the Get and check for nil?

ianwilkes: @VinozzZ has found that we include nil values in the trace key, meaning they have to be allowed to exist in the payload, so I'm going to implement Exists(). The two-argument form makes updating all the test code a nightmare.

types/payload.go
ianwilkes: keysToFind is the set of values which haven't been memoized yet, so we don't know if the value is zero-length until we get here, unless the keys param itself is zero-length, which I'm assuming won't generally happen.

PR #1589, opened 2025-06-17
fix: wait for all goroutines inside collector to finish during shutdown

VinozzZ:

Making sure the collector has properly shutdown

- Add a WaitGroup for all goroutines created inside the collector


collect/collect.go
```diff
@@ -105,6 +105,7 @@ type InMemCollector struct {
 
 	sampleTraceCache cache.TraceSentCache
 
+	wg                sync.WaitGroup
```
ianwilkes: I might call this `shutdownWG` or `stopWG` to make it clear what this is for.

ianwilkes (approved):

kentquirk (approved):

PR #1590, opened 2025-06-17
perf: use concrete type for Counter, Gauge, and Histogram metrics

VinozzZ:

Use concrete type as input when recording Counter, Gauge, and Histogram metrics. This is to prevent allocation on the heap caused by converting a concrete type to `interface{}`

- Use `float64` as input type for Gauge and Histogram
- Use `int64` as input type for Counter
- Implement a wrapper for libhoney internal metrics since transmission.Metrics requires input being passed in as `interface{}`

### DIFF
--- a/collect/collect.go
+++ b/collect/collect.go
@@ -268,7 +268,7 @@ func (i *InMemCollector) checkAlloc(ctx context.Context) {
 
 	var mem runtime.MemStats
 	runtime.ReadMemStats(&mem)
-	i.Metrics.Gauge("memory_heap_allocation", int64(mem.Alloc))
+	i.Metrics.Gauge("memory_heap_allocation", float64(mem.Alloc))
 	if maxAlloc == 0 || mem.Alloc < uint64(maxAlloc) {
 		return
 	}
@@ -300,7 +300,7 @@ func (i *InMemCollector) checkAlloc(ctx context.Context) {
 	// or just run out of traces to delete.
 
 	cacheSize := len(allTraces)
-	i.Metrics.Gauge("collector_cache_size", cacheSize)
+	i.Metrics.Gauge("collector_cache_size", float64(cacheSize))
 
 	totalDataSizeSent := 0
 	tracesSent := generics.NewSet[string]()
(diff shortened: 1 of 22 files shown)


collect/collect.go
```diff
@@ -483,7 +483,7 @@ func (i *InMemCollector) redistributeTraces(ctx context.Context) {
 	redistrubutionStartTime := i.Clock.Now()
 
 	defer func() {
-		i.Metrics.Histogram("collector_redistribute_traces_duration_ms", i.Clock.Now().Sub(redistrubutionStartTime).Milliseconds())
+		i.Metrics.Histogram("collector_redistribute_traces_duration_ms", float64(i.Clock.Now().Sub(redistrubutionStartTime).Milliseconds()))
```
kentquirk: i just noticed the typo! (redistribution). You didn't introduce it, but if it's local, maybe fix?

metrics/metrics.go
```diff
@@ -35,50 +35,23 @@ import (
...
-	Gauge(name string, val interface{})     // for gauges
-	Count(name string, n interface{})       // for counters
-	Histogram(name string, obs interface{}) // for histogram
-	Up(name string)                         // for updown
-	Down(name string)                       // for updown
-	Get(name string) (float64, bool)        // for reading back a counter or a gauge
-	Store(name string, val float64)         // for storing a rarely-changing value not sent as a metric
+	Increment(name string)              // for counters
+	Gauge(name string, val float64)     // for gauges
+	Count(name string, n int64)         // for counters
```
kentquirk, line 40: We'd avoid a lot of type conversions if this were just an int instead of int64.

kentquirk (approved):
I think this is good. Original design was trying to avoid changing the interface, but it seems worthwhile.

metrics/metrics.go
VinozzZ: Unfortunately, the dynsampler-go returns int64 for its metrics. I'm not sure it's always safe to convert it into an int
```
	// GetMetrics returns a map of metrics about the sampler's performance.
	// All values are returned as int64; counters are cumulative and the names
	// always end with "_count", while gauges are instantaneous with no particular naming convention.
	// All names are prefixed with the given string.
	GetMetrics(prefix string) map[string]int64
```

kentquirk: Ah, ok. It *is* always safe in the context of anything but a 32-bit Go, but that's a good reason. Thanks.

PR #1591, opened 2025-06-17
maint: use collect.Start() in collector_test.go

VinozzZ:

We have lot of boilerplate code in collector test that manually starts a valid collector. This is hard to maintain and keep in sync with the production code

- switch to collector.Start in tests


kentquirk (approved):
I won't stand in the way, but I do have two concerns:
* Does this properly isolate the things we need it to, or have these tests now become integration tests?
* I'm slightly worried it will cause a noticeable change in overall test runtime. Did you check?

VinozzZ:
> I won't stand in the way, but I do have two concerns:
> 
> * Does this properly isolate the things we need it to, or have these tests now become integration tests?
> * I'm slightly worried it will cause a noticeable change in overall test runtime. Did you check?

I always appreciate your feedback!

- I believe this is still properly isolating things we need based on the configs we provide with less manual setups
- I checked the test time from this change. It doesn't seem to have an impact on the overall test runtime

PR #1592, opened 2025-06-17
perf: cleanup superfluous allocations during event unmarshal

ianwilkes:
Waste in the existing unmarshal() implementation remaining from the following PR:
- #1586

Skips the expensive Reader-based vmihailenco/msgpack.Decoder implementation for the common case of messagepack batches. Also de-pointerifies Payload's internal MsgpPayloadMap which was another heap allocation for no good reason. This is in no way comprehensive; there is other fat to trim in the batch ingest path.

```
benchmark                   old ns/op     new ns/op     delta
BenchmarkRouterBatch-12     5857          3724          -36.42%

benchmark                   old allocs     new allocs     delta
BenchmarkRouterBatch-12     69             44             -36.23%

benchmark                   old bytes     new bytes     delta
BenchmarkRouterBatch-12     5535          3550          -35.86%
```

### DIFF
--- a/route/batched_event.go
+++ b/route/batched_event.go
@@ -1,0 +1,120 @@
+package route
+
+import (
+	"bytes"
+	"time"
+
+	"github.com/honeycombio/refinery/types"
+	"github.com/tinylib/msgp/msgp"
+)
+
+type batchedEvent struct {
+	Timestamp        string        `json:"time"`
+	MsgPackTimestamp *time.Time    `msgpack:"time,omitempty"`
+	SampleRate       int64         `json:"samplerate" msgpack:"samplerate"`
+	Data             types.Payload `json:"data" msgpack:"data"`
+}
+
+func (b *batchedEvent) getEventTime() time.Time {
+	if b.MsgPackTimestamp != nil {
+		return b.MsgPackTimestamp.UTC()
+	}
+
+	return getEventTime(b.Timestamp)
+}
+
+func (b *batchedEvent) getSampleRate() uint {
+	if b.SampleRate == 0 {
+		return defaultSampleRate
+	}
+	return uint(b.SampleRate)
+}
+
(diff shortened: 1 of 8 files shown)


route/batched_event.go
```diff
@@ -0,0 +1,120 @@
...
+
+// UnmarshalMsg implements msgp.Unmarshaler
+func (b *batchedEvents) UnmarshalMsg(bts []byte) (o []byte, err error) {
+	var totalValues uint32
+	totalValues, bts, err = msgp.ReadArrayHeaderBytes(bts)
+	if err != nil {
+		err = msgp.WrapError(err)
+		return
+	}
+	if cap(*b) >= int(totalValues) {
```
VinozzZ, line 106: In what scenario will the `batchedEvents` be reused?

route/route.go
```diff
@@ -439,7 +440,7 @@ func (r *Router) requestToEvent(ctx context.Context, req *http.Request, reqBod [
 	}
 
 	data := map[string]interface{}{}
-	err = unmarshal(req, bytes.NewReader(reqBod), &data)
+	err = unmarshal(req, reqBod, &data)
```
VinozzZ, line 443: We are not implementing a custom unmarshaler here. Is it because a single event is probably small enough to just deserialized it into memory?

route/batched_event.go
ianwilkes: It won't be, at this time anyway, but as a general practice deserialization routines should consider the possibility unless there's a good reason not too. With this code we _could_ use a pool to re-use these and may one day want to.

route/route.go
ianwilkes: Who even uses this endpoint anymore? I'm assuming nobody with enough volume to care about performance is sending un-batched events, just as I'm assuming they aren't sending JSON.

PR #1593, opened 2025-06-18
fix: wait for goroutines to finish before shutting down

VinozzZ:

While this issue may not affect deployed instances of Refinery, it impacts the developer experience when running tests. Previously, some goroutines were not properly shut down after calling the shutdown function. This resulted in noisy and cluttered profiles during test runs, making it difficult to analyze performance or debug issues effectively.

- wait for goroutines created by various components to finish before return from `Stop()`

### DIFF
--- a/collect/cache/cuckoo.go
+++ b/collect/cache/cuckoo.go
@@ -94,6 +94,7 @@ func NewCuckooTraceChecker(capacity uint, m metrics.Metrics) *CuckooTraceChecker
 func (c *CuckooTraceChecker) Stop() {
 	// stop the goroutine that drains the add channel
 	close(c.done)
+	close(c.addch)
 	c.shutdownWG.Wait()
 
 	// make sure we drain the channel one last time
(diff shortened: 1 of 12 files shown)


collect/cache/cuckoo.go
```diff
@@ -59,26 +62,46 @@ func NewCuckooTraceChecker(capacity uint, m metrics.Metrics) *CuckooTraceChecker
...
 			}
 		}
 	}()
 
 	return c
 }
 
+func (c *CuckooTraceChecker) Stop() {
+	// stop the goroutine that drains the add channel
+	close(c.done)
```
ianwilkes, line 96: You should close c.addch here also. Assuming everything is shutting down in the correct order, you shouldn't get any more calls to Add(). If things are not shutting down, you'll find out because it will panic!

Note doing so means you'll also need to use the two-value channel read in drain to detect if it's closed.

ianwilkes (approved):

PR #1595, opened 2025-06-20
feat: draft DirectTransmit to replace libhoney for trace data transmission

ianwilkes:
Libhoney is old and tries to be safe and have good ergonomics and support a lot of options, and that means it also comes with a lot of overhead. It's also not easily extensible, should we want to improve refinery peer-to-peer communication.

This is a first iteration of a wholesale replacement for libhoney (for refinery's use only). It's not nearly as much code as libhoney, because it doesn't do nearly as much stuff, but doing these sends directly make some jobs - like handling send responses - _much_ simpler. It's not ready to go yet though, note the TODOs here. I think this PR is big enough already, and would like to land this and extend it in further PRs. Note I'm not actually using this new code yet; the goal is to get to a correct implementation (which this isn't because it doesn't honor max event or batch byte sizes), then improve the performance and scalability of the implementation.

This PR also extends the integration tests with a test API server. In the past we didn't really need that since we leaned on libhoney's tested transmission library, but rolling our own means we need our own testing.

I also removed the Flush() method from Transmission, because AFAICT it doesn't do anything that we weren't doing already in Stop().


transmit/direct_transmit.go
```diff
@@ -0,0 +1,377 @@
...
+		if !ok {
+			// TODO buffer size - use larger buffer to avoid blocking
+			ch := make(chan *types.Event, 100)
+			d.transmitQueues[key] = ch
+
+			d.shutdownWG.Add(1)
+			go d.batchEvents(ch)
+			q = ch
+		}
+		d.mutex.Unlock()
```
VinozzZ: This is probably a nit. Can this `Unlock` be moved to after line 105?

transmit/direct_transmit.go
```diff
@@ -0,0 +1,377 @@
...
+}
+
+// implements msgp.Marshaler, based on generated code.
+// Note though that this uses AppendTimeExt (the messagepack standard),
+// vs the default AppendTime() which uses a library-specific extension.
+// Since we don't use this library universally, AppendTime() will create
+// unreadable timestamps.
+func (z *batchedEvent) MarshalMsg(b []byte) (o []byte, err error) {
+	o = b
+	// map header, size 3
```
VinozzZ: Are we writing the map header here?

transmit/direct_transmit.go
```diff
@@ -0,0 +1,377 @@
...
+	d.mutex.RLock()
+	q, ok := d.transmitQueues[key]
+	d.mutex.RUnlock()
+
+	if !ok {
+		d.mutex.Lock()
+		q, ok = d.transmitQueues[key]
+		if !ok {
+			// TODO buffer size - use larger buffer to avoid blocking
+			ch := make(chan *types.Event, 100)
```
VinozzZ, line 106: The size of this channel is probably the equivalent of `PendingWorkCapacity` in libhoney?

transmit/direct_transmit.go
```diff
@@ -0,0 +1,377 @@
...
+			d.transmitQueues[key] = ch
+
+			d.shutdownWG.Add(1)
+			go d.batchEvents(ch)
```
VinozzZ, line 114: We probably also want to have a limit for the amount of concurrent batches?

transmit/direct_transmit.go
```diff
@@ -0,0 +1,377 @@
...
+		WithString("api_host", ev.APIHost).
+		WithString("dataset", ev.Dataset).
+		Logf("transmit sending event")
+
+	// Store enqueue time for queue time metrics
+	ev.EnqueuedUnixMicro = time.Now().UnixMicro()
+
+	// TODO this may end up on the heap, if so, either construct it inline, or make it
+	// a sub-struct of Event (which is already on the heap).
+	key := transmitKey{
```
VinozzZ, line 89: If the amount of traffic per key combination is unevenly distributed, the dataset has large amount of traffic will be capped by a single batch processing goroutine. That seems not ideal

transmit/direct_transmit_test.go
```diff
@@ -0,0 +1,502 @@
...
+	}
+
+	// Wait for all events to be processed (11 total)
+	require.Eventually(t, func() bool {
+		receivedEvents := testServer.getEvents()
+		return len(receivedEvents) == 11
+	}, 200*time.Millisecond, 10*time.Millisecond)
+
+	// Verify all events were received
+	receivedEvents := testServer.getEvents()
```
VinozzZ: What's the difference between this check and the require.Eventually check above?

transmit/direct_transmit_test.go
```diff
@@ -0,0 +1,502 @@
...
+	err := dt.Start()
+	require.NoError(t, err)
+	defer dt.Stop()
+
+	now := time.Now().UTC()
+
+	// Create events for multiple datasets and scenarios
+	var allEvents []*types.Event
+
+	// Dataset A: 5 events (should create 2 batches: 3 + 2)
```
VinozzZ, line 377: Is there a way to verify the batching logic works as expected as well?

transmit/direct_transmit.go
ianwilkes: Yes, it's in the first append. I'll update the comment to make that clearer.

transmit/direct_transmit.go
ianwilkes: Not directly, because that controls (I think) the total number of pending events, while this is a per-channel value. Because there could be hundreds of thousands of these channels, the value can't be that larger; I'd like to leave this here as a placeholder and figure out the right way to tune it later.

transmit/direct_transmit.go
ianwilkes: Yes, I should have noted in the PR description. This entire queue and dispatch mechanism is a placeholder to allow end-to-end testing, and is NOT viable for production use yet. My plan to to shift to a finite set of workers which are able to share load.

transmit/direct_transmit.go
ianwilkes: This is bounded by the number of keys - so probably ~1M in the very worst case of a huge cluster and thousands of datasets. That is maybe too many goroutines, so, yeah.

transmit/direct_transmit_test.go
ianwilkes: that is just some AI slop!

transmit/direct_transmit_test.go
ianwilkes: I added an assertion to the test server that the batch size doesn't exceed the maximum.

ianwilkes:
The structural concerns are correct; this basic data flow is just minimum functionality and will need to change. I could iterate on that in this PR, but would rather land it and go from there to limit the size of a single diff. What I'd like to do next is integrate this new implementation with our other tests, so that moving forward I can have confidence about its correctness as the plumbing changes.

VinozzZ (approved):

PR #1597, opened 2025-06-20
maint: remove metro hash and use wyhash instead

VinozzZ:

We already have a hashing algorithm for hashing. We don't need another one

- remove metro hash in trace_key.go and use existing wyhash instead


kentquirk (approved):
Lol, I probably should have asked you about this on the other PR. I figured it had been suggested by someone else.

PR #1600, opened 2025-06-24
feat: add support for REFINERY_HONEYCOMB_LOGGER_ADDITIONAL_FIELDS to honeycomb logger

TylerHelmuth:

- Allows the honeycomb logger to support adding attributes from the `OTEL_RESOURCE_ATTRIBUTES` env var to all exported logs.

- add logic for extracting attributes from the `OTEL_RESOURCE_ATTRIBUTES` env var

### DIFF
--- a/logger/honeycomb.go
+++ b/logger/honeycomb.go
@@ -81,6 +81,7 @@ func (h *HoneycombLogger) Start() error {
 	}
 	h.libhClient = libhClient
 
+	h.addResourceAttributes()
 	h.libhClient.AddField("refinery_version", h.Version)
 	if hostname, err := os.Hostname(); err == nil {
 		h.libhClient.AddField("hostname", hostname)
@@ -89,7 +90,6 @@ func (h *HoneycombLogger) Start() error {
 	h.libhClient.AddDynamicField("process_uptime_seconds", func() interface{} {
 		return time.Since(startTime) / time.Second
 	})
-	h.addResourceAttributes()
 
 	h.builder = h.libhClient.NewBuilder()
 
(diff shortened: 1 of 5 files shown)


logger/honeycomb.go
```diff
@@ -103,6 +105,31 @@ func (h *HoneycombLogger) Start() error {
 	return nil
 }
 
+// Add support for the OTEL_RESOURCE_ATTRIBUTES env var.
```
VinozzZ: I'm sorry that I didn't realize this is what you were planning to do when we talked about it. Mixing OTel config with honeycomb config is not ideal. I think it would be confusing for users to know what otel env config is supported by honeycomb logger/metrics v.s. the otel one.

Is it possible to support setting client's scope attribute for honeycomb logger through another environment variable instead of the one for otel?

TylerHelmuth: Ya we can name it whatever we want.

TylerHelmuth: Want to scope it to honeycome? Like `REFINERY_HONEYCOMB_LOGGER_ADDITIONAL_FIELDS`?

VinozzZ: Yes, that sounds good to me. Then we can go through the normal environment variable flow and add this as a config option under the HoneycombLogger

TylerHelmuth: Do you want me to update the config as well to have the option or only allow env var for now?

VinozzZ: Yes, we can have full config + env var support for adding additional fields to honeycomb logger

TylerHelmuth: Updated to have full config support. The implementation now relies on `go-flags` implementation for env/flag maps, which uses `:` as the delimiter.

VinozzZ (approved):

PR #1601, opened 2025-06-24
perf: memoize fields that don't exist on a payload

VinozzZ:

To avoid repeatedly iterating over the event payload for non-existing keys, we should remember the keys we have looked for and differentiate that from existing fields with nil value.

- track non-existing fields with `missingFields` in `Payload` struct

## Benchmark Results
Old
```
goos: darwin
goarch: arm64
pkg: github.com/honeycombio/refinery/types
cpu: Apple M2 Max
BenchmarkPayload/get_unknown         	1813033	     6572 ns/op	      0 B/op	      0 allocs/op
BenchmarkPayload/exist_unknown       	1807669	     6631 ns/op	      0 B/op	      0 allocs/op
```
New
```
goos: darwin
goarch: arm64
pkg: github.com/honeycombio/refinery/types
cpu: Apple M2 Max
BenchmarkPayload/get_unknown         	1000000000	       10.00 ns/op	      0 B/op	      0 allocs/op
BenchmarkPayload/exist_unknown       	1000000000	       10.00 ns/op	      0 B/op	      0 allocs/op
```

### DIFF
--- a/types/payload.go
+++ b/types/payload.go
@@ -142,10 +142,6 @@ func (p *Payload) Exists(key string) bool {
 		}
 	}
 
-	if p.missingFields == nil {
-		p.missingFields = make(map[string]struct{})
-	}
-	p.missingFields[key] = struct{}{}
 	return false
 }
 


kentquirk (approved):
Looks fine, but if you're adding a benchmark, it would always be cool to show the before/after results.

types/payload.go
```diff
@@ -151,6 +182,11 @@ func (p *Payload) Get(key string) any {
 		}
 	}
 
+	if p.missingFields == nil {
+		p.missingFields = make(map[string]struct{})
+	}
+	p.missingFields[key] = struct{}{}
```
ianwilkes: I don't think we should implicitly memoize - that has a non-trivial cost. The assumption should be that a field is accessed once only, and if it's going to be more than that, it's the subject of a MemoizeFields call.

ianwilkes (commented):
TestPayload doesn't have much coverage of this stuff, you should probably add a case there with a call MemoizeFields() on fields which do and don't exist.

types/payload.go
```diff
@@ -121,16 +142,29 @@ func (p *Payload) Exists(key string) bool {
...
```
ianwilkes: again, I don't think we should memoize by default

ianwilkes (approved):

types/payload.go
VinozzZ: Oops, sorry I missed this one

PR #1604, opened 2025-06-25
perf: switch to an async pool for DirectTranmission batch sends, test changes

ianwilkes:
The existing DirectTransmission implementation sends batches one at a time, which bottlenecks easily.

This sets up a concurrency pool (basically, one goroutine per POST, with a cap) - this is a little less efficient than a classic worker pool but avoid spinning up large numbers of goroutines when there's nothing for them to do. This involves some test changes since events no longer show up in deterministic order. The insane practice of one goroutine per host/api-key/dataset is still in place, for now.

Also switches integration tests away from libhoney to DirectTransmission, to ensure correctness as I change the code. I expect to remove DefaultTransmission soon.

The improved dispatch is mostly visible in the single-target transmission benchmark, although I don't yet consider this code optimized. High-card benchmark suffers a bit due to the extra overhead of goroutine dispatch:

```
BenchmarkTransmissionComparison/high_card/direct-12      988           1108          +12.09%
BenchmarkTransmissionComparison/low_card/direct-12       2970          931           -68.65%
```


app/app_test.go
```diff
@@ -671,11 +612,12 @@ func TestEventsEndpoint(t *testing.T) {
...
 
 	var apps [2]*App
-	var senders [2]*transmission.MockSender
+	var senders [2]*transmit.MockTransmission
 	for i := range apps {
 		var graph inject.Graph
 		basePort := 13000 + (i * 2)
-		senders[i] = &transmission.MockSender{}
+		senders[i] = &transmit.MockTransmission{}
+		senders[i].Start()
```
VinozzZ: I believe the `Start` is called in `newStartedApp`

transmit/direct_transmit.go
```diff
@@ -147,6 +160,12 @@ func (d *DirectTransmission) Stop() error {
 	}
 	d.transmitQueues = nil
 	d.shutdownWG.Wait()
+
+	if d.batchPool != nil {
+		d.batchPool.Wait()
+	}
+	d.batchPool = nil
```
VinozzZ, line 167: I'm curious why we want to assign `nil` value to the `batchPool` on shutdown?

transmit/direct_transmit.go
```diff
@@ -359,11 +378,10 @@ func (d *DirectTransmission) sendBatch(wholeBatch []*types.Event) {
 func (d *DirectTransmission) batchEvents(in <-chan *types.Event) {
 	defer d.shutdownWG.Done()
 
-	var batch []*types.Event
 	for ev := range in {
 		// Start a new batch, and wait for the first event.
-		batch = batch[:0]
-		batch = append(batch, ev)
+		batch := make([]*types.Event, 1, d.MaxBatchSize)
```
VinozzZ, line 383: Is the reason we can’t reuse the underlying array because each worker needs its own independent copy of the batched events?

VinozzZ (approved):

app/app_test.go
ianwilkes: Correct!

transmit/direct_transmit.go
ianwilkes: Because that way if something wasn't shut down properly, it will probably panic. Which is better than it silently dropping events and never being noticed forever.

transmit/direct_transmit.go
ianwilkes: Because we hand it to sendBatch(), which now runs concurrently. So we can't overwrite the contents of the slice, because that would mangle the data we gave to sendBatch(). Now, the slice is created, filled, passed to sendBatch(), and then becomes garbage.

A more efficient approach would recycle these via a pool, and I may do that at some point.

PR #1605, opened 2025-06-25
perf: more scalable DirectTransmission batching

ianwilkes:
In the original implementation, goroutine and timer proliferation were a concern.

Note: based on #1604

This replaces a whole bunch of channel and goroutine contention with a whole bunch of lock contention. I think that's going to be ok because the stuff we do while holding the lock is very cheap, and ultimately less than what libhoney was doing. It bechmarks well - the remaining allocation here is mostly in the http transport (including the test server itself) and there's not much to be done about that.

```
benchmark                                               old ns/op     new ns/op     delta
BenchmarkTransmissionComparison/high_card/direct-12     1101          702           -36.23%
BenchmarkTransmissionComparison/low_card/direct-12      894           640           -28.39%

benchmark                                               old allocs     new allocs     delta
BenchmarkTransmissionComparison/high_card/direct-12     3              2              -33.33%
BenchmarkTransmissionComparison/low_card/direct-12      3              2              -33.33%

benchmark                                               old bytes     new bytes     delta
BenchmarkTransmissionComparison/high_card/direct-12     5035          3050          -39.42%
BenchmarkTransmissionComparison/low_card/direct-12      4864          3211          -33.98%
```

### DIFF
--- a/transmit/direct_transmit.go
+++ b/transmit/direct_transmit.go
@@ -12,6 +12,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/jonboulle/clockwork"
 	"github.com/sourcegraph/conc/pool"
 	"github.com/tinylib/msgp/msgp"
 	"github.com/vmihailenco/msgpack/v5"
@@ -59,6 +60,7 @@ type DirectTransmission struct {
 	// constructed, not injected
 	Metrics   metrics.Metrics
 	Transport *http.Transport
+	Clock     clockwork.Clock
 
 	// Type is peer or upstream, and used only for naming metrics
 	Name string
@@ -67,12 +69,13 @@ type DirectTransmission struct {
 	MaxBatchSize int
 	BatchTimeout time.Duration
 
-	// Slice-based batching
-	eventBatches map[transmitKey]*eventBatch
-	batchMutex   sync.RWMutex
-	batchPool    *pool.Pool
-	ticker       *time.Ticker
-	done         chan struct{}
+	// Double-buffer slice-based batching
+	batches    [2]map[transmitKey]*eventBatch
+	hotIndex   int // 0 or 1, indicates which map is currently hot
+	batchMutex sync.RWMutex
+	batchPool  *pool.Pool
+	done       chan struct{}
+	shutdownWG sync.WaitGroup
 
 	httpClient *http.Client
 	userAgent  string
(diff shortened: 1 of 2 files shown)


transmit/direct_transmit.go
```diff
@@ -103,41 +118,52 @@ func (d *DirectTransmission) EnqueueEvent(ev *types.Event) {
 		Logf("transmit sending event")
 
 	// Store enqueue time for queue time metrics
-	ev.EnqueuedUnixMicro = time.Now().UnixMicro()
+	ev.EnqueuedUnixMicro = d.Clock.Now().UnixMicro()
 
-	// TODO this may end up on the heap, if so, either construct it inline, or make it
```
VinozzZ, line 108: Does this mean it's not allocated on the heap?

transmit/direct_transmit.go
```diff
@@ -103,41 +118,52 @@ func (d *DirectTransmission) EnqueueEvent(ev *types.Event) {
...
+				startTime: d.Clock.Now(),
+			}
+			d.eventBatches[key] = batch
 		}
+		d.batchMutex.Unlock()
+	}
+
+	// Add event to batch
+	batch.mutex.Lock()
+	if batch.events == nil {
```
VinozzZ, line 146: How would this be possible with the Lock held on line 134?

transmit/direct_transmit_test.go
```diff
@@ -535,10 +539,26 @@ func TestDirectTransmission(t *testing.T) {
...
-	// Send all events, the stop to make sure everything is fully flushed.
-	for _, event := range allEvents {
+	// Send all but one of our events.
+	for _, event := range allEvents[:len(allEvents)-1] {
 		dt.EnqueueEvent(event)
 	}
+
+	// At this point, time hasn't advanced so we should't get any incomplete batches.
+	require.Eventually(t, func() bool {
+		return len(testServer.getEvents()) == len(allEvents)-6
```
VinozzZ, line 552: Can we add a comment about how we came up with this calculation? 6 includes the dropped event due to size, 4 waiting for timeout, and 1 that hasn't been enqueued

VinozzZ (approved):

transmit/direct_transmit.go
ianwilkes: Doesn't show up in the profiler, so I guess not!

VinozzZ:
I'm amazed by the improvements from switching to locks from channels

transmit/direct_transmit.go
ianwilkes: Well, for one thing at this point that lock is no longer held. But more importantly, the batch may have a nil slice because it gets set to nil when the batch is dispatched. This is a good catch though, in fact I shouldn't be initializing it above here - only here.

PR #1606, opened 2025-06-25
feat: switch to DirectTransmission in refinery main

ianwilkes:
I think it's time to try it out.

Note: based on #1605 - last of the series.

Adds support for zstd compression and enables DirectTransmission in refinery proper. Also contains a bit of miscellaneous cleanup.

### DIFF
--- a/cmd/refinery/main.go
+++ b/cmd/refinery/main.go
@@ -180,7 +180,7 @@ func main() {
 		true,
 	)
 	peerTransmission := transmit.NewDirectTransmission(
-		upstreamMetricsRecorder,
+		peerMetricsRecorder,
 		peerTransport,
 		"peer",
 		int(c.GetTracesConfig().GetMaxBatchSize()),
--- a/cmd/refinery/main.go
+++ b/cmd/refinery/main.go
@@ -22,8 +22,6 @@ import (
 	"github.com/facebookgo/startstop"
 	"github.com/google/uuid"
 	"github.com/honeycombio/husky"
-	libhoney "github.com/honeycombio/libhoney-go"
-	"github.com/honeycombio/libhoney-go/transmission"
 	"github.com/jonboulle/clockwork"
 	"github.com/sirupsen/logrus"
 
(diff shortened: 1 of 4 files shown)


transmit/direct_transmit.go
```diff
@@ -334,7 +364,20 @@ func (d *DirectTransmission) sendBatch(wholeBatch []*types.Event) {
...
-		req, err := http.NewRequest("POST", apiURL.String(), bytes.NewReader(packed))
+
+		var reader io.Reader
+		if d.enableCompression {
+			zBufPtr := batchBufferPool.Get().(*[]byte)
+			zBuf := zstdEncoder.EncodeAll(packed, *zBufPtr)
+			reader = bytes.NewReader(zBuf)
+
+			*zBufPtr = zBuf[:0]
+			defer batchBufferPool.Put(zBufPtr)
```
VinozzZ, line 372: Should we put it back per batch instead of the whole function is finished?

cmd/refinery/main.go
```diff
@@ -172,46 +170,23 @@ func main() {
...
+	upstreamTransmission := transmit.NewDirectTransmission(
+		upstreamMetricsRecorder,
+		upstreamTransport,
+		"upstream",
+		int(c.GetTracesConfig().GetMaxBatchSize()),
+		time.Duration(c.GetTracesConfig().GetBatchTimeout()),
+		true,
+	)
+	peerTransmission := transmit.NewDirectTransmission(
+		upstreamMetricsRecorder,
```
VinozzZ: this should be peerMetricsRecorder

transmit/direct_transmit.go
ianwilkes: It doesn't really matter as long as it goes back eventually, and given the various error handling below, this is simpler.

cmd/refinery/main.go
ianwilkes: oops

VinozzZ (approved):

PR #1607, opened 2025-06-26
perf: prefetch span attributes during ingest and sampling

VinozzZ:

Fetch multiple data from the event payload is faster than doing it one at a time

TODO: I need to add tests for the changes

### DIFF
--- a/collect/collect.go
+++ b/collect/collect.go
@@ -1318,8 +1318,12 @@ func (i *InMemCollector) createDecisionSpan(sp *types.Span, trace *types.Trace, 
 
 	dc := sp.ExtractDecisionContext()
 	// extract all key fields from the span
-	keyFields := sampler.GetKeyFields()
-	sp.Data.MemoizeFields(keyFields...)
+	keyFields, nonRootFields := sampler.GetKeyFields()
+	if sp.IsRoot {
+		sp.Data.MemoizeFields(keyFields...)
+	} else {
+		sp.Data.MemoizeFields(nonRootFields...)
+	}
 	for _, keyField := range keyFields {
 		// Less efficient than a two-return version of Get(), so consider adding
 		// that to the Payload interface if these becomes a hotspot.
@@ -1529,6 +1533,17 @@ func (i *InMemCollector) makeDecision(ctx context.Context, trace *types.Trace, s
 	if sampler, found = i.datasetSamplers[samplerSelector]; !found {
 		sampler = i.SamplerFactory.GetSamplerImplementationForKey(samplerSelector, isLegacyKey)
 		i.datasetSamplers[samplerSelector] = sampler
+	}
+
+	// prepopulate spans with key fields
+	nonRootFields, rootFields := sampler.GetKeyFields()
+	allFields := append(nonRootFields, rootFields...)
+	for _, sp := range trace.GetSpans() {
+		if sp.IsRoot {
+			sp.Data.MemoizeFields(allFields...)
+		} else {
+			sp.Data.MemoizeFields(nonRootFields...)
+		}
 	}
 
 	startGetSampleRate := i.Clock.Now()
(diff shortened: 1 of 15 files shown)


ianwilkes:
I decided to try going further and put our core metadata fields in the struct as first-class fields. This is reasonably ergonomic from the outside, but very verbose on the inside, so I have mixed feelings about it. It generates fewer allocations per field, but ultimately memoizes more fields than are used in the route benchmark, so it looks like a wash on memory but should do better in production, I think. It's definitely faster vs performance_work_branch:

```
benchmark                old ns/op     new ns/op     delta
BenchmarkRouterBatch     44137         18956         -57.05%

benchmark                old allocs     new allocs     delta
BenchmarkRouterBatch     61             62             +1.64%

benchmark                old bytes     new bytes     delta
BenchmarkRouterBatch     65020         66156         +1.75%
```

In theory, this should make life easier and faster for the collector, but I'm open to the idea that maybe this is more complexity than is warranted and we should just stick with memoization. Profiling shows that a lot of of allocations here are unrelated to this specific code; there's some general cleanup we need to do in route.go. Most of the CPU time is just reading all the bytes, and the messsagepack parse; there may be more improvement possible on both those fronts.

I'm not too confident in this list of fields either. I asked claude for advice and here's what I got, this may be a good starting point if we want to go further with this:

Fields that are commonly used but NOT in the dedicated list:

1. "meta.refinery.original_sample_rate" - Used frequently in collect.go to preserve the original sample rate
2. "meta.span_count" - Used 6 times in collect.go for root span enrichment
3. "meta.event_count" - Used 3 times for root span enrichment
4. "meta.span_event_count" - Used 3 times for root span enrichment
5. "meta.span_link_count" - Used 3 times for root span enrichment
6. "meta.refinery.reason" - Used 3 times to record sampling decisions
7. "meta.refinery.local_hostname" - Used 3 times for debugging/tracking

Recommendations:

1. Consider adding meta.refinery.original_sample_rate to the dedicated fields list since it's used on every sampled span and is important for rate calculations.
2. Maybe add the count fields (meta.span_count, meta.event_count, etc.) if AddCountsToRoot is commonly enabled, as these would be accessed frequently on root spans.
3. Keep as-is the other fields like meta.refinery.reason and meta.refinery.local_hostname since they're used less frequently and only in specific scenarios.

ianwilkes:
Taking this a bit further to only do a single pass on msgpack data, in the ideal batch/msgpack case. I'd love to see how this performs in prod.

```
benchmark                old ns/op     new ns/op     delta
BenchmarkRouterBatch     44137         12024         -72.76%
```

collect/collect.go
```diff
@@ -554,10 +554,10 @@ func (i *InMemCollector) redistributeTraces(ctx context.Context) {
 
 			sp.APIHost = newTarget.GetAddress()
 
-			if v := sp.Data.Get("meta.refinery.forwarded"); v != nil {
-				sp.Data.Set("meta.refinery.forwarded", fmt.Sprintf("%s,%s", v, i.hostname))
+			if sp.Data.MetaRefineryForwarded != "" {
```
MikeGoldsmith, line 557: I think it's odd that some properties use the accessors, eg .Value and .Get(), .Set, but others do not. I think it would be best if we consolidate onto a single pattern.

types/payload.go
```diff
@@ -18,12 +414,198 @@ type Payload struct {
...
+
+	p.hasExtractedMetadata = true
+	return nil
+}
+
+// NewPayload creates a new Payload from a map of fields. This is not populate
+// metadata fields; to do this, you MUST call ExtractMetadata.
 func NewPayload(data map[string]any) Payload {
-	return Payload{
+	p := Payload{
```
MikeGoldsmith, line 605: nit: don't think we need this change

MikeGoldsmith (approved):
This is a big PR but I understand the theme of what's happening. I like that we're trying to get all of the most common & consistent fields in one iteration of the underlying data.

collect/collect.go
VinozzZ: Yes! we have a follow up PR that adds the rest of the metadata into this pattern

types/payload.go
VinozzZ: I will address this in the follow up PR

PR #1609, opened 2025-07-01
perf: promote all write-only metadata fields as first class fields in Payload struct

VinozzZ:

Writing to a type aware struct field is faster than writing to a `map[string]any`.

- create constants for all metadata fields that's write-only in the `collector`
- use bigger payload for collector benchmark test data
- create fields on Payload struct for all write-only metadata fields


sample/rules_test.go
```diff
@@ -518,103 +518,6 @@ func TestRules(t *testing.T) {
 			ExpectedRate:      10,
 			ExpectedKeyFields: []string{"test"},
 		},
-		{
```
VinozzZ, line 521: This test is deleted because this is no longer a supported feature

app/app_test.go
```diff
@@ -137,6 +137,7 @@ func (t *testAPIServer) handleBatch(w http.ResponseWriter, r *http.Request) {
 		}
 		if dataMap, ok := eventData["data"].(map[string]interface{}); ok {
 			event.Data = types.NewPayload(dataMap)
+			event.Data.ExtractMetadata([]string{"trace.trace_id"}, []string{"trace.parent_id"})
```
MikeGoldsmith, line 140: There should be OTel semconv constants we can use for these field names.

types/payload.go
```diff
@@ -535,52 +650,117 @@ func (p *Payload) Get(key string) any {
...
+	case MetaRefineryShutdownSend:
+		if v, ok := value.(bool); ok {
+			p.MetaRefineryShutdownSend.Set(v)
+		}
+		return
+	case MetaRefinerySampleKey:
+		if v, ok := value.(string); ok {
+			p.MetaRefinerySampleKey = v
+		}
+		return
```
MikeGoldsmith: Is this because we always want to return when we get a match on a known field name because we don't need to memoize it?

VinozzZ: Yes, we don't need to waste space for duplicated information

app/app_test.go
VinozzZ: This is actually from the default value for `TraceIdFieldNames` and `ParentIdFieldNames` config options

PR #1610, opened 2025-07-02
perf: use a buffer pool for http body reads

ianwilkes:
It sure is expensive allocating these big buffers for every HTTP read, especially given that their size is initially unknown (at least when compressed.) We also copied compressed bodies twice, which was totally unnecessary.

We also did not apply any body size limit to uncompressed payloads.

The main change here is to keep a pool of buffers for the HTTP bodies, allowing us to easily avoid potentially large allocations. It also restructures the read so that we only read the body contents at most once, and applies the size limit to uncompressed payloads.

For our optimized Payload path, this does mean we need to make new copies of the relevant sections of the body for storage with the Payload. However, this is better than what we did before, because it's less memory, the size is known a priori, and it solves the problem of a single Event hanging onto a reference to the entire buffer from a much larger batch.

We could take this further and also recycle Payload's messagepack buffer, but that requires accurately identifying each and every place where an Event goes out of scope - a job I'd rather leave to the garbage collector, at least for now.

### DIFF
--- a/route/route.go
+++ b/route/route.go
@@ -692,7 +692,7 @@ var httpBodyBufferPool = sync.Pool{
 	},
 }
 
-// When finished with the returned buffer, callers should return it to bufferPool.
+// When finished with the returned buffer, callers should return it to httpBodyBufferPool.
 // Reads the body and immediately closes it, releasing resources as soon as possible.
 func (r *Router) readAndCloseMaybeCompressedBody(req *http.Request) (*bytes.Buffer, error) {
 	defer req.Body.Close()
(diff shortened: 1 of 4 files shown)


route/route.go
```diff
@@ -721,16 +717,19 @@ func (r *Router) getMaybeCompressedBody(req *http.Request) (io.Reader, error) {
...
-
-		reader = buf
+		reader = zReader
 	default:
 		reader = req.Body
 	}
-	return reader, nil
+
+	buf := httpBodyBufferPool.Get().(*bytes.Buffer)
+	buf.Reset()
```
MikeGoldsmith: What do you think about moving the `Reset` into a release function as it's possible a brand new buffer is returned here and wouldn't need to be Reset? Callers of `readAndCloseMaybeCompressedBody` would then defer the release func instead of the `Put` back into the pool.

MikeGoldsmith (approved):
Looks good to me. Left one suggestion regarding moving the `Reset` into a release func.

route/route.go
ianwilkes: Sure.

PR #1611, opened 2025-07-03
fix: fully initialize Payload object during unmarshaling

VinozzZ:

Currently, the Payload object isn’t fully initialized until ExtractMetadata is explicitly called. This creates a risk: if a developer forgets to call ExtractMetadata, the Payload instance may appear valid but lack critical metadata — leading to subtle bugs.

- Modified Payload to hold a reference to the Config object.
- Enabled automatic metadata extraction during unmarshaling by invoking ExtractMetadata within the unmarshal logic.
- Standardized Payload initialization through NewPayload and NewEmptyPayload constructors.
- Applied the same initialization logic for JSON format.
- Exception: OTEL data deserialization happens in Husky, so ExtractMetadata must still be called manually after Husky returns parsed results.

While reviewing the transmission logic, I found two metadata fields that shouldn’t always be forwarded to Honeycomb:
- meta.trace_id: Only used in distributed mode. Now replaced with the configured trace ID field name for decision spans, ensuring meta.trace_id is never sent externally.
- meta.refinery.root: Also used only for internal peer communication. Updated logic to only send this field if its value is true.


route/batched_event.go
```diff
@@ -13,6 +15,7 @@ type batchedEvent struct {
 	MsgPackTimestamp *time.Time    `msgpack:"time,omitempty"`
 	SampleRate       int64         `json:"samplerate" msgpack:"samplerate"`
 	Data             types.Payload `json:"data" msgpack:"data"`
+	cfg              config.Config `json:"-" msgpack:"-"`
```
kentquirk, line 18: I don't think you need the struct tag here saying to ignore the field, because the field is not exported and is ignored anyway. But it also doesn't hurt to have it, if only for documentation.

types/payload.go
```diff
@@ -241,11 +245,20 @@ func (p *Payload) extractMetadataFromBytes(data []byte, traceIdFieldNames, paren
 // This MUST be called manually after creating or unmarshaling a non-empty Payload
```
kentquirk: Is this comment still true, or should it be modified?

types/payload.go
```diff
@@ -241,11 +245,20 @@ func (p *Payload) extractMetadataFromBytes(data []byte, traceIdFieldNames, paren
...
 	if p.hasExtractedMetadata {
 		return nil
 	}
 
+	var parentIdFieldNames, traceIdFieldNames []string
+	if p.config != nil {
+		parentIdFieldNames = p.config.GetParentIdFieldNames()
+		traceIdFieldNames = p.config.GetTraceIdFieldNames()
+	}
+	// buf := make([]byte, 1024)
```
kentquirk: delete leftover code?

types/payload.go
```diff
@@ -275,6 +288,7 @@ func (p *Payload) ExtractMetadata(traceIdFieldNames, parentIdFieldNames []string
 				}
 			case MetaRefineryRoot:
 				if v, ok := value.(bool); ok {
+					fmt.Println("memoized", v)
```
kentquirk: delete, I think

types/payload.go
```diff
@@ -367,14 +381,15 @@ func (p *Payload) ExtractMetadata(traceIdFieldNames, parentIdFieldNames []string
...
-		_, err := p.extractMetadataFromBytes(p.msgpMap.rawData, traceIdFieldNames, parentIdFieldNames)
+		_, err := p.extractMetadataFromBytes(p.msgpMap.rawData)
 		if err != nil {
 			return err
 		}
 	}
 
 	// After processing all fields, set root status if needed and not already set
 	if needRootDecision && !p.MetaRefineryRoot.HasValue && p.MetaSignalType != "log" {
+		fmt.Println("extractmetadata end", !foundParentId)
```
kentquirk: again, probably not wanted

types/payload.go
```diff
@@ -383,41 +398,35 @@ func (p *Payload) ExtractMetadata(traceIdFieldNames, parentIdFieldNames []string
 }
 
 // NewPayload creates a new Payload from a map of fields. This is not populate
-// metadata fields; to do this, you MUST call ExtractMetadata.
-func NewPayload(data map[string]any) Payload {
+// metadata fields
+func NewPayload(data map[string]any, config config.Config) Payload {
```
kentquirk: you could possibly combine these into a single function that takes a data parameter that could be nil; this could minimize confusion about when to use which. (then the comment should be more complete)

types/payload.go
```diff
@@ -495,6 +505,53 @@ func (p *Payload) MemoizeFields(keys ...string) {
 }
 
 func (p *Payload) Exists(key string) bool {
+	// if the key is a metadata field, check the dedicated field
```
kentquirk, line 1108: I understand why this exists, but I really don't like it for its high long-term maintenance cost. We have to remember to do this for metadata forever.

kentquirk (approved):
A bunch of little things, but no big objections.

types/payload.go
VinozzZ: I added the helper function to the metadatafield list

route/batched_event.go
```diff
@@ -54,19 +84,12 @@ func (b *batchedEvent) UnmarshalMsg(bts []byte) (o []byte, err error) {
...
 				b.MsgPackTimestamp = nil
 			} else {
 				if b.MsgPackTimestamp == nil {
 					b.MsgPackTimestamp = new(time.Time)
 				}
 				*b.MsgPackTimestamp, bts, err = msgp.ReadTimeBytes(bts)
-				if err != nil {
-					err = msgp.WrapError(err, "MsgPackTimestamp")
-					return
-				}
```
ianwilkes: why did this come out?

route/route.go
```diff
@@ -606,13 +598,12 @@ func (r *Router) processEvent(ev *types.Event, reqID interface{}) error {
 	r.Metrics.Histogram(r.incomingOrPeer+"_router_event_bytes", float64(ev.GetDataSize()))
 
 	// An error here is effectively a parsing error, so return it up the stack.
-	err := ev.Data.ExtractMetadata(r.Config.GetTraceIdFieldNames(), r.Config.GetParentIdFieldNames())
-	if err != nil {
+	if err := ev.Data.ExtractMetadata(); err != nil {
```
ianwilkes, line 601: I thought the idea was to get rid of this call entirely - which cases do we still need it for?

ianwilkes (approved):
Good progress - too bad this is so verbose but I think it's a sound investment.

route/route.go
VinozzZ: This is for otlp data

route/batched_event.go
VinozzZ: oops, I will add it back

PR #1612, opened 2025-07-03
maint: remove unused LibhoneyMetricsWrapper

VinozzZ:

Since we are no longer using libhoney as the tramission library, we don't need to maintain a libhoney specific metrics wrapper anymore.

- remove unused LibhoneyMetricsWrapper


kentquirk (approved):
We like red diffs.

PR #1615, opened 2025-07-06
perf: remove dynamic string creation for metric names

VinozzZ:

When a string is dynamically created, it's allocated on the heap. This creates pressure on the GC which is biggest performance issue Refinery is currently battling with

- Only compute necessary metric names once during startup
- store the computed result as struct fields for quicker access

## Benchmarks
```
**OLD**
goos: darwin
goarch: arm64
pkg: github.com/honeycombio/refinery/route
cpu: Apple M2 Max
BenchmarkRouterBatch-12    	5418073	     8647 ns/op	  17322 B/op	     41 allocs/op
PASS
ok  	github.com/honeycombio/refinery/route	47.190s

**NEW**
goos: darwin
goarch: arm64
pkg: github.com/honeycombio/refinery/route
cpu: Apple M2 Max
BenchmarkRouterBatch-12    	5644776	     8416 ns/op	  17092 B/op	     33 allocs/op
PASS
ok  	github.com/honeycombio/refinery/route	47.861s
```

### DIFF
--- a/collect/collect.go
+++ b/collect/collect.go
@@ -434,7 +434,7 @@ func (i *InMemCollector) collect() {
 				span.End()
 				return
 			}
-			i.processSpan(ctx, sp, "peer")
+			i.processSpan(ctx, sp, types.RouterTypeIncoming)
 		default:
 			select {
 			case msg, ok := <-i.dropDecisionMessages:
@@ -472,14 +472,14 @@ func (i *InMemCollector) collect() {
 					span.End()
 					return
 				}
-				i.processSpan(ctx, sp, "incoming")
+				i.processSpan(ctx, sp, types.RouterTypeIncoming)
 			case sp, ok := <-i.fromPeer:
 				if !ok {
 					// channel's been closed; we should shut down.
 					span.End()
 					return
 				}
-				i.processSpan(ctx, sp, "peer")
+				i.processSpan(ctx, sp, types.RouterTypePeer)
 			case <-i.reload:
 				i.reloadConfigs()
 			}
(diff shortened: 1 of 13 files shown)


VinozzZ:
> Two things:
> 
> * Given this change, I think it makes sense to go further and change incomingOrPeer to either a bool (isIncoming) or an enum with two constants. There's little sense in carrying a string around when we don't use it for anything any more other than logging.
> * The change to the sampler constants is a good start, but their GetMetrics functions all do multiple string concatenations every call. It's probably worth an equivalent PR there to fix that -- and it might be worth changing it. Maybe we don't need to pass in the prefix to the samplers at all? At the very least, it could be passed in once rather than on every call.

I changed the `incomingOrPeer` into an enum with two constants.
I'm going to do a release for dynsampler-go and then update Refinery to the latest version in another PR

collect/collect.go
```diff
@@ -789,7 +796,12 @@ func (i *InMemCollector) processSpan(ctx context.Context, sp *types.Span, source
 	var spanForwarded bool
 	// if this trace doesn't belong to us and it's not in sent state, we should forward a decision span to its decider
 	if !trace.Sent && !isMyTrace {
-		i.Metrics.Increment(source + "_router_peer")
+		switch source {
+		case "peer":
```
ianwilkes: Why is source still a string? Shouldn't you also use the enum here?

ianwilkes (approved):

collect/collect.go
VinozzZ: It was because of circular dependency. I just moved the RouterType into the `types` package and updated the code here to reuse that

types/router_types.go
```diff
@@ -0,0 +1,28 @@
...
+	RouterTypePeer
+)
+
+func (rt RouterType) String() string {
+	switch rt {
+	case RouterTypeIncoming:
+		return "incoming"
+	case RouterTypePeer:
+		return "peer"
+	default:
```
kentquirk: This is going to be annoyingly nitpicky, so to make it clear I'm fine if you don't change anything.

You've set up a situation here where there's a possibility that the value is out of the defined range (although you'd have to explicitly assert an integer to get it). I was originally thinking that if it was an int, you might use it as an index into a slice of strings.

But these `Is...` functions treat it like a bool, except that out-of-range values are treated differently in the two cases.

I'd love to see one of:
* Only have either IsPeer or IsIncoming
* Define one of them using == and the other using !=
* Define one of them in terms of the other
* Use a bool instead of an int as the base type

kentquirk (approved):
One nit.

types/router_types.go
VinozzZ: Thank you for pointing it out! I didn't realize the inconsistency. I changed the RouterType to be a boolean. This is much simpler. I like it

PR #1617, opened 2025-07-10
perf: reduce lock contention for internal metrics

VinozzZ:

Refinery supports three metrics backends: libhoney (Honeycomb), OpenTelemetry (OTel), and Prometheus. Users can enable one or more of these backends simultaneously through the MultiMetrics wrapper.

Currently, each metrics implementation—including MultiMetrics—maintains its own internal locks and in-memory copies of metrics. This leads to unnecessary duplication, especially for metrics critical to the Stress Relief system, where we only need a single shared in-memory representation.

Additionally:
- OTel and Prometheus implementations are already thread-safe and do not require extra synchronization.
- The libhoney implementation previously relied on a mutex for concurrency control. This PR updates it to use atomic operations instead, reducing lock contention and improving performance under load.

- Replace mutex with atomic operation for all metrics except histogram in libhoney metrics
- Remove inmemory metrics value in both OTel and Prometheus since MultiMetrics is storing them already
- Only hold locks for map access to retrieve a metric by name in OTel and Prometheus metrics
- Enable msgpack encoding for libhoney metrics and logger.
- Add tests and benchmark tests

## Benchmark
```
goos: darwin
goarch: arm64
pkg: github.com/honeycombio/refinery/metrics
cpu: Apple M2 Max
                                         │   old.txt    │                new.txt                │
                                         │    sec/op    │    sec/op     vs base                 │
ConcurrentAccess/ConcurrentCounters-12     437.2n ± ∞ ¹   354.0n ± ∞ ¹        ~ (p=1.000 n=1) ²
ConcurrentAccess/ConcurrentGauges-12       353.7n ± ∞ ¹   238.9n ± ∞ ¹        ~ (p=1.000 n=1) ²
ConcurrentAccess/ConcurrentHistograms-12   482.7n ± ∞ ¹   356.8n ± ∞ ¹        ~ (p=1.000 n=1) ²
ConcurrentAccess/ConcurrentMixed-12        483.6n ± ∞ ¹   368.9n ± ∞ ¹        ~ (p=1.000 n=1) ²
geomean                                    435.9n         324.8n        -25.48%
```

### DIFF
--- a/metrics/legacy.go
+++ b/metrics/legacy.go
@@ -8,6 +8,7 @@ import (
 	"runtime"
 	"sort"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	libhoney "github.com/honeycombio/libhoney-go"
@@ -43,15 +44,22 @@ type LegacyMetrics struct {
 }
 
 type counter struct {
-	lock sync.Mutex
 	name string
 	val  int64
 }
 
-type gauge struct {
-	lock sync.Mutex
-	name string
-	val  float64
+func (c *counter) add(delta int64) {
+	atomic.AddInt64(&c.val, delta)
+}
+
+// get atomically loads the value of the counter.
+func (c *counter) get() int64 {
+	return atomic.LoadInt64(&c.val)
+}
+
+// report atomically swaps the value of the counter to 0 and returns the previous value.
+func (c *counter) report() int64 {
+	return atomic.SwapInt64(&c.val, 0)
 }
 
 type histogram struct {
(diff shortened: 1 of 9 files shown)


metrics/legacy.go
```diff
@@ -43,15 +44,22 @@ type LegacyMetrics struct {
 }
 
 type counter struct {
-	lock sync.Mutex
 	name string
 	val  int64
```
ianwilkes: It would be clearer and safer to use an `atomic.Int64` here, and the atomic data types in general for values.

metrics/legacy.go
```diff
@@ -242,25 +283,15 @@ func (h *LegacyMetrics) reportToHoneycomb(ctx context.Context) {
...
-				count.lock.Unlock()
+				ev.AddField(count.name, count.report())
 			}
 
 			for _, updown := range h.updowns {
-				updown.lock.Lock()
-				ev.AddField(updown.name, updown.val)
-				// updown.val = 0   // updowns are never reset to 0
-				updown.lock.Unlock()
+				ev.AddField(updown.name, updown.report())
```
ianwilkes: The old code says updowns are never reset, but that's exactly what your report() method does. Seems like this should have caused a test failure.

metrics/legacy.go
```diff
@@ -61,9 +69,41 @@ type histogram struct {
...
+	return atomic.LoadInt64(&u.val)
+}
+
+func (u *updown) add(n int64) {
+	atomic.AddInt64(&u.val, n)
+}
+
+type gauge struct {
+	name string
+	val  uint64
```
ianwilkes: Add a comment that this is really a float in disguise.

metrics/prometheus.go
```diff
@@ -84,85 +80,81 @@ func (p *PromMetrics) Register(metadata Metadata) {
...
 		}
 	}
 }
 
 func (p *PromMetrics) Store(name string, val float64) {
-	p.lock.Lock()
-	defer p.lock.Unlock()
-
-	p.values[name] = val
+	return
```
ianwilkes: Wait what? Is this because prometheus doesn't support Store()? Add a comment explaining it's a no-op and the reason.

ianwilkes (changes requested):
Nice to clean up some unnecessary steps but I definitely think the old-school function-based atomics should be considered deprecated.

metrics/legacy.go
VinozzZ: Good catch! I added a test for it

metrics/prometheus.go
VinozzZ: I created a separate interface for concrete metrics implementation. The thinking here is that only the metrics wrapper needs to store an in-memory version of metrics. All the concrete implementation shouldn't worry about that

ianwilkes (approved):

Issue #1616, opened 2025-07-09
Batch processing fails if a single event's data field is null


egalvanibooking:
## Problem
When calling the `batch` endpoint, if any event in the payload has its data field set to `null`, the entire batch fails to process, even if the other events are correctly defined.

Because of some unexpected behaviour in the pre-processing logic on our side, some events were send to refinery with `data: null`. This led to the following error log, and complete batch processing failure. 

```
level=error msg="handler returning error" error.err="assignment to entry in nil map" error.msg="caught panic" error.stack_trace="goroutine 36143477 [running]: runtime/debug.Stack() 
.... 
github.com/honeycombio/refinery/route.addIncomingUserAgent(...) /home/builder/go/refinery/route/route.go:1081 github.com/honeycombio/refinery/route.(*Router).batch(0xc0002e18e8, {0x12a44b0, 0xc2ae2c3e00}, 0xc04db6d400) /home/builder/go/refinery/route/route.go:509 +0x8fa
```

After checking the source code, this appears to be the flow:
1. [calling](https://github.com/honeycombio/refinery/blob/main/route/route.go#L480) `unmarshal` on an event with `data: null` results in the `Data` field being set to `nil`
2. the `addIncomingUserAgent` method triggers a panic [when assigning](https://github.com/honeycombio/refinery/blob/main/route/route.go#L1112) `ev.Data["meta.refinery.incoming_user_agent"] = userAgent` 
3. since there is no error handing around that, the full batch fails 

The refinery version we are using is `2.9.1` but also the new versions seem affected by this behaviour. 

## Ideas
- Skip or emit specific error for events with empty / null data field

Would love your thoughts on this! Happy to open a PR if this fix makes sense.

PR #1618, opened 2025-07-14
fix: send http status code 400 for empty event payload

VinozzZ:

fix: #1616 

- return 400 for empty event payload for libhoney endpoints


TylerHelmuth (approved):

robbkidd (approved):
Ooooh yeah, test cases. ❤️

PR #1619, opened 2025-07-14
fix: address race condition in managing locks for internal OTel metrics

robbkidd:

- We fell victim to the classic blunder of concurrency: read lock release before read-write lock without a reread. This introduced a bug in managing the registration and state-changes for OTel metrics. 

- Moved the lock management into `getOrInit(FooType)` functions and checked for a metrics existence again after taking the read-write lock.
- Added lots of tests.
- And some benchmarking.

### DIFF
--- a/metrics/otel_metrics.go
+++ b/metrics/otel_metrics.go
@@ -34,6 +34,7 @@ type OTelMetrics struct {
 
 	meter        metric.Meter
 	shutdownFunc func(ctx context.Context) error
+	testReader   sdkmetric.Reader
 
 	lock             sync.RWMutex
 	counters         map[string]metric.Int64Counter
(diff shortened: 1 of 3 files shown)


metrics/otel_metrics.go
```diff
@@ -129,12 +130,18 @@ func (o *OTelMetrics) Start() error {
 		return err
 	}
 
+	var reader sdkmetric.Reader
+
+	reader = sdkmetric.NewPeriodicReader(exporter,
```
kentquirk: Suggestion: put this inside the other case of the testreader conditional, rather than creating it and immediately overwriting it?

metrics/otel_metrics.go
```diff
@@ -226,183 +230,180 @@ func (o *OTelMetrics) Register(metadata Metadata) {
...
+	o.lock.RUnlock()
+
+	if ok { // yey, the guage exists; return it
+		return g, nil
 	}
 
-	// Give the counter an initial value of 0 so that OTel will send it
-	ctr.Add(context.Background(), 0)
-	o.counters[metadata.Name] = ctr
+	// oh, so sad; gotta make the guage
```
kentquirk:
```suggestion
	// oh, so sad; gotta make the gauge
```

robbkidd: 🤦🏻‍♂️

metrics/otel_metrics.go
robbkidd: Sure! I was following what I thought was a Go idiom of straight-happy-path/branching-difference, but I'm happy to either-or in [an if-else instead](https://github.com/honeycombio/refinery/pull/1619/commits/debe32c90bd894c97f8f6358cd7d233dd8f7fd20).

kentquirk: Yeah, you're not wrong on the idiom, but in this case it's an expensive happy path, so I thought it worth doing that.

PR #1620, opened 2025-07-15
perf: memoize sampling fields in router

VinozzZ:

-

-


types/event.go
```diff
@@ -351,6 +350,39 @@ func (sp *Span) CacheImpact(traceTimeout time.Duration) int {
...
+		// Check the prefix pattern "hc[a-z]ic_"
+		if key[:2] != "hc" || key[3:6] != "ic_" {
+			return false
+		}
+		if key[2] < 'a' || key[2] > 'z' {
+			return false
+		}
+		if key[3] != 'i' || key[4] != 'c' || key[5] != '_' {
+			return false
+		}
```
ianwilkes: you already checked for this above

types/event.go
```diff
@@ -351,6 +350,39 @@ func (sp *Span) CacheImpact(traceTimeout time.Duration) int {
...
+		for i := 0; i < keyLen; i++ {
+			c := key[i]
+			if !((c >= '0' && c <= '9') || (c >= 'a' && c <= 'f')) {
+				return false
+			}
+		}
+		return true
+	} else if keyLen == 64 {
+		// Check the prefix pattern "hc[a-z]ic_"
+		if key[:2] != "hc" || key[3:6] != "ic_" {
```
ianwilkes: you ought to add a benchmark for IsLegacyAPIKey - I think the compiler will optimize away the implicit string instantiations here but you'll want to be sure.

types/event.go
```diff
@@ -351,6 +350,36 @@ func (sp *Span) CacheImpact(traceTimeout time.Duration) int {
 	return multiplier * sp.GetDataSize()
 }
 
-func IsLegacyAPIKey(apiKey string) bool {
-	return huskyotlp.IsClassicApiKey(apiKey)
+func IsLegacyAPIKey(key string) bool {
+	keyLen := len(key)
```
robbkidd:
```suggestion
	// TODO: implement this optimization in husky and go back to calling it from there?
	keyLen := len(key)
```

config/file_config.go
```diff
@@ -850,6 +850,35 @@ func (f *fileConfig) GetAllSamplerRules() *V2SamplerConfig {
...
+		isLegacyKey = true
+	}
+
+	if isLegacyKey {
+		if prefix := f.GetDatasetPrefix(); prefix != "" {
+			return fmt.Sprintf("%s.%s", prefix, samplerKey)
+		}
+	}
+
+	return samplerKey
```
ianwilkes: This seems needlessly convoluted. How about
```
if !IsLegacyAPIKey(apiKey) {
	return environment
}
if prefix := f.GetDatasetPrefix(); prefix != "" {
	return fmt.Sprintf("%s.%s", prefix, dataset)
}
return dataset
```

config/file_config.go
```diff
@@ -850,6 +850,35 @@ func (f *fileConfig) GetAllSamplerRules() *V2SamplerConfig {
...
+}
+
+func (f *fileConfig) GetSamplingKeyFieldsForDestName(samplerKey string) []string {
+	sampler, ok := f.rulesConfig.Samplers[samplerKey]
+	if !ok {
+		sampler, ok = f.rulesConfig.Samplers["__default__"]
+		if !ok {
+			return nil
+		}
+	}
```
ianwilkes, line 875: I would expect this to require a mutex, as in GetSamplerConfigForDestName below - or is this entirely read-only at this point, allowing lock-free access?

config/file_config.go
```diff
@@ -1088,3 +1117,37 @@ func (f *fileConfig) GetAdditionalAttributes() map[string]string {
...
+	if keyLen == 32 {
+		// Check if all characters are hex digits (0-9, a-f)
+		for i := 0; i < keyLen; i++ {
+			c := key[i]
+			if !((c >= '0' && c <= '9') || (c >= 'a' && c <= 'f')) {
+				return false
+			}
+		}
+		return true
+	} else if keyLen == 64 {
```
ianwilkes: nit: idiomatically in go using a `switch` is preferred over `else if`

config/file_config_test.go
```diff
@@ -112,3 +113,207 @@ func TestAccessKeyConfig_IsAccepted(t *testing.T) {
...
+					General: GeneralConfig{
+						DatasetPrefix: tc.prefix,
+					},
+				},
+			}
+
+			result := config.CalculateSamplerKey(tc.apiKey, tc.dataset, tc.environment)
+			assert.Equal(t, tc.expected, result,
+				"CalculateSamplerKey(%q, %q, %q) = %q, want %q",
+				tc.apiKey, tc.dataset, tc.environment, result, tc.expected)
```
ianwilkes: nit: formatting. Since this doesn't fit on one line you should fully expand it with every param on a new line. (someday gofmt will do this automatically)

config/file_config_test.go
```diff
@@ -112,3 +113,207 @@ func TestAccessKeyConfig_IsAccepted(t *testing.T) {
...
+						"my-destination": {
+							DynamicSampler: &DynamicSamplerConfig{
+								FieldList: []string{"sampling_key", "service.name"},
+							},
+						},
+					},
+				},
+			}
+			result := config.GetSamplingKeyFieldsForDestName(tc.destName)
+			assert.Equal(t, tc.expected, result,
```
ianwilkes: formatting, as above

config/file_config_test.go
```diff
@@ -112,3 +113,207 @@ func TestAccessKeyConfig_IsAccepted(t *testing.T) {
...
+				"getSamplingKeyFieldsForDestName(%q) = %v, want %v",
+				tc.destName, result, tc.expected)
+		})
+	}
+}
+
+func TestIsClassicKey(t *testing.T) {
```
ianwilkes: TestIsLegacyAPIKey

types/payload.go
```diff
@@ -350,7 +349,31 @@ func (p *Payload) extractMetadataFromBytes(data []byte) (int, error) {
...
 		}
 
-		// If we didn't handle this field as metadata, skip it
+		// If not handled as metadata, check if this is a key field
+		// only check for key fields if we haven't found a match yet
+		if !handled && len(samplingKeyFields) > 0 && keysFound < len(samplingKeyFields) {
+			// Check if the key matches any of the key fields
+			var val any
+			for _, field := range samplingKeyFields {
+				if p.memoizedFields[field] != nil {
```
ianwilkes: arguably this should use the two-value map lookup to check if any value has been set, including nil. Unless we want to overwrite any nils that were previously set.

types/payload.go
```diff
@@ -350,7 +349,31 @@ func (p *Payload) extractMetadataFromBytes(data []byte) (int, error) {
 			return len(data) - len(remaining), fmt.Errorf("failed to read value for key %s: %w", string(keyBytes), err)
 		}
 
-		// If we didn't handle this field as metadata, skip it
+		// If not handled as metadata, check if this is a key field
+		// only check for key fields if we haven't found a match yet
+		if !handled && len(samplingKeyFields) > 0 && keysFound < len(samplingKeyFields) {
```
ianwilkes: You don't need to check `len(samplingKeyFields) > 0`. If `len(samplingKeyFields)` is zero and keysFound is zero we'll still skip this.

config/file_config.go
VinozzZ: Oh yes, good catch. I missed it

PR #1621, opened 2025-07-15
fix: use unique port number for integration test

VinozzZ:

Integration tests are running in parallel. We need to use unique port for each of them

- use unique port number for `TestAppIntegrationEmptyEvent`


robbkidd (commented):
port authority

PR #1626, opened 2025-07-17
perf: remove metricsnamer to avoid dynamic string concatenation

VinozzZ:

-

-


cmd/refinery/main.go
```diff
@@ -164,23 +165,19 @@ func main() {
 		ForceAttemptHTTP2:   true,
 	}
 
-	genericMetricsRecorder := metrics.NewMetricsPrefixer("")
-	upstreamMetricsRecorder := metrics.NewMetricsPrefixer("libhoney_upstream")
-	peerMetricsRecorder := metrics.NewMetricsPrefixer("libhoney_peer")
```
robbkidd, line 169: If I'm understanding all of these changes correctly, the `libhoney_` portion of the existing prefix goes away in this PR I _think_ regardless of whether the `transmit` (libhoney) or `direct_transmit` (all msgpack all the time) is used.

Will that name change for all metrics break current Refinery boards?

VinozzZ: Ugh, you are absolutely correct. I think it's appropriate to remove `libhoney_` prefix from the metrics produced by `direct_transmit` since it no longer uses `libhoney`. However, I also hear the concern about existing Refinery boards.

Maybe this is the reason we should consider doing a major release so that we can modify existing metrics. What do you think?

robbkidd: 🤔 

My first thought—which admittedly is often not the _best_ thought—is add `libhoney_` back in for the prefix for all metrics that had it before, even if the metric is generated by the direct transmission. Let the name be a lie so that the performance improvements included in the upcoming release can be seen in anyone's existing Refinery ops boards.

Maybe we can make the `libhoney_` inclusion in the names configurable so that we can default it to **included** until a major release where we change the default to **not included** and have new board templates ready for folks.

VinozzZ: I added back the old `libhoney_` prefix. I agree that if we can change it until a major release

transmit/direct_transmit.go
```diff
@@ -203,6 +247,39 @@ func (d *DirectTransmission) EnqueueSpan(sp *types.Span) {
...
+			d.metricKeys.updownQueuedItems = fullName
+		case histogramQueueTime:
+			d.metricKeys.histogramQueueTime = fullName
+		case counterEnqueueErrors:
+			d.metricKeys.counterEnqueueErrors = fullName
+		case counterResponse20x:
+			d.metricKeys.counterResponse20x = fullName
+		case counterResponseErrors:
+			d.metricKeys.counterResponseErrors = fullName
+		// below are existing internal libhoney metrics. Prefix it with libhoney to ensure compatibility
```
robbkidd: I recommend including what we're ensuring compatibility with in the comment for our Future Selves™ to consider the plan for deprecation and removal of this prefix.

```suggestion
		// Below are metrics previously associated with the libhoney transmission used to send data upstream or to peers.
		// Even though libhoney isn't used, include the prefix in these metric names to avoid breaking existing Refinery operations boards & queries.
```

robbkidd (approved):
Everything gets a name.

PR #1629, opened 2025-07-18
perf: avoid dynamic string creation for sampling fields with no data

VinozzZ:

When a field used for sampling has no value, Refinery formats that `<nil>` value dynamically. This results in unnecessary allocation when we know what the value is going to be. 

- treat `nil` value as a known data type when creating sampling keys

## Benchmark Results
```
**OLD**
goos: darwin
goarch: arm64
pkg: github.com/honeycombio/refinery/sample
cpu: Apple M2 Max
BenchmarkTraceKeyBuild/empty_fields         	 807175	    14129 ns/op	   1056 B/op	    118 allocs/op
PASS
ok  	github.com/honeycombio/refinery/sample	12.004s

**NEW**
goos: darwin
goarch: arm64
pkg: github.com/honeycombio/refinery/sample
cpu: Apple M2 Max
BenchmarkTraceKeyBuild/empty_fields         	1208139	     9945 ns/op	    522 B/op	     18 allocs/op
PASS
ok  	github.com/honeycombio/refinery/sample	12.621s
```

### DIFF
--- a/sample/trace_key_test.go
+++ b/sample/trace_key_test.go
@@ -415,7 +415,7 @@ func BenchmarkTraceKeyBuild(b *testing.B) {
 			fields:         []string{"empty-field", "http.status_code", "request.path"},
 			useTraceLength: true,
 			spanCount:      100,
-			uniqueValues:   10, // Will hit maxKeyLength
+			uniqueValues:   10,
 			rootFields:     false,
 		},
 	}


robbkidd (approved):
Yep. That's better faster.

Daft Punk approves

PR #1631, opened 2025-07-23
fix: implement missing libhoney metrics in DirectTransmit

VinozzZ:

When we first replaced `transmit` + `libhoney` with `directTransmit`, we didn't implement the existing libhoney internal metrics. One of the internal libhoney metrics is required by Stress Relief. 

This PR does _not_ implement the `libhoney_*_send_retries` metric since we currently don't have retry logic yet. That will be added in a follow up PR when retry logic is added.

- Implement below metrics
  - libhoney_upstream/peer_queue_length
  - libhoney_upstream/peer_batches_sent
  - libhoney_upstream/peer_messages_sent
  - libhoney_upstream/peer_response_decode_errors


robbkidd (approved):
meters

PR #1632, opened 2025-07-23
perf: use protobuf to msgp translator for otlp/http protobuf request

VinozzZ:

This PR uses the new request translator in husky that directly converts http/protobuf message into messagepack.
Since this is feature is still in development, I just used the commit hash as the tag for Husky's import path. Once we are ready for release, we can go through the proper release tag

- unmarshal http/protobuf payload using `TranslateTraceRequestFromReaderSizedWithMsgp`
- add a new metric, `router_otlp_http_proto`, to track number of request that are going through the new unmarshaler
- add tests for the new metric


route/route.go
```diff
@@ -606,16 +611,59 @@ func (router *Router) processOTLPRequest(
 				Environment: environment,
 				SampleRate:  uint(ev.SampleRate),
 				Timestamp:   ev.Timestamp,
-				Data:        types.NewPayload(router.Config, ev.Attributes),
+				Data:        payload,
```
codeboten: not a big deal but is this change necessary?

VinozzZ: It's _unnecessary_ :). I removed it

PR #1634, opened 2025-07-24
test: add OTLP to app test and benchmark

ianwilkes:
Our GRPC server setup code at the App level was never being tested, and our app-level integration test didn't include OTLP at all.

Adds OTLP in both GRPC and HTTP forms to our app integration test, to ensure we're actually setting up GRPC correctly. This will be important since I'm about to change that. Lightly refactors the non-distributed trace benchmarks to make them more efficient, and drops the single-value benchmark because it's not an optimization target.

Also attempts to fix the flaky TestOriginalSampleRateIsNotedInMetaField test.

### DIFF
--- a/app/app_test.go
+++ b/app/app_test.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"compress/gzip"
 	"context"
+	"encoding/binary"
 	"encoding/json"
 	"fmt"
 	"io"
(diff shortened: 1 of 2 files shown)


app/app_test.go
```diff
@@ -1016,86 +1308,13 @@ func BenchmarkTraces(b *testing.B) {
 			req.Header.Set("Content-Type", encoding.contentType)
 			req.Header.Set("Content-Encoding", encoding.encoding)
 
-			b.Run("single", func(b *testing.B) {
```
VinozzZ, line 1019: Is this so that it matches with the otlp benchmark test?

ianwilkes: Yeah, it was a hassle to keep it and I really see no value, it's distracting.

PR #1637, opened 2025-07-28
fix: sampling key fields extraction for root-prefixed sampling fields

VinozzZ:

This PR fixes issues in Refinery's sampling key fields extraction logic. The change makes sure to interpret `root.` prefix in sampling field definition during extraction.

This PR does not address the computed field during extraction. That will be addressed in a separate PR

- Fixed parameter order in CalculateSamplerKey(apiKey, env, dataset) across all implementations and call sites
- Enhanced field extraction in CoreFieldsUnmarshaler to properly handle root and non-root field combinations 
- Standardized root prefix handling by moving RootPrefix constant to config package 
- Fixed field processing in payload unmarshaling to use the `root.` prefix aware key field extraction logic


types/payload.go
```diff
@@ -528,7 +528,7 @@ func (p CoreFieldsUnmarshaler) UnmarshalPayloadComplete(bts []byte, payload *Pay
 // msgpMap. This is useful for cases where the Payload is part of a larger message and we want to avoid
 // unnecessary allocations.
 func (p CoreFieldsUnmarshaler) UnmarshalPayload(bts []byte, payload *Payload) ([]byte, error) {
```
robbkidd: If `p` is getting updated to `cu` for `UnmarshalPayloadComplete()`,maybe update it here, too?

PR #1638, opened 2025-07-28
fix: flaky TestDirectTransmissionQueueLengthGauge

VinozzZ:

This test was flaky because the test isn't waiting for the full batch timeout.

- make sure the test waits longer than the batch timeout before checking batch sent metric

### DIFF
--- a/transmit/direct_transmit_test.go
+++ b/transmit/direct_transmit_test.go
@@ -949,7 +949,15 @@ func TestDirectTransmissionQueueLengthGauge(t *testing.T) {
 	mockMetrics := &metrics.MockMetrics{}
 	mockMetrics.Start()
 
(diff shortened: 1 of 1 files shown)


transmit/direct_transmit_test.go
```diff
@@ -983,7 +991,7 @@ func TestDirectTransmissionQueueLengthGauge(t *testing.T) {
 	assert.Eventually(t, func() bool {
 		v, _ := mockMetrics.Get(dt.metricKeys.gaugeQueueLength)
 		return v == 0
-	}, 200*time.Millisecond, 10*time.Millisecond)
+	}, batchTimeout+100*time.Millisecond, 10*time.Millisecond)
```
robbkidd: What do you think of making these `waitFor` and `tick` durations relative to the value of `batchTimeout`?

```suggestion
	}, batchTimeout*2, batchTimeout/10)
```

PR #1639, opened 2025-07-28
ci: fix flaky tests

robbkidd:

- Flaky tests!

Expanding on what was found in #1623:

* use the `*assert.CollectT` instance instead of `testing.T` inside of the `EventuallyWithT()` assertion function, otherwise the first failed assertion in the Eventually loop fails the entire test
* base the `waitFor` and `tick` durations on the `SendTickerValue`
* replace a `time.Sleep()` with an `EventuallyWithT()`

### DIFF
--- a/collect/collect_test.go
+++ b/collect/collect_test.go
@@ -450,11 +450,16 @@ func TestAddSpan(t *testing.T) {
 	span.Event.Data.ExtractMetadata()
 	coll.AddSpanFromPeer(span)
 
-	assert.EventuallyWithT(t, func(collect *assert.CollectT) {
-		trace := coll.getFromCache(traceID)
-		require.NotNil(t, trace)
-		assert.Equal(t, traceID, trace.TraceID, "after adding the span, we should have a trace in the cache with the right trace ID")
-	}, time.Second*1, time.Millisecond)
+	assert.EventuallyWithT(t,
+		func(cT *assert.CollectT) {
+			trace := coll.getFromCache(traceID)
+			require.NotNil(cT, trace)
+			assert.Equal(cT, traceID, trace.TraceID)
+		},
+		conf.GetTracesConfig().GetSendTickerValue()*60, // waitFor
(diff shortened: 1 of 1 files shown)


collect/collect_test.go
```diff
@@ -473,11 +478,21 @@ func TestAddSpan(t *testing.T) {
...
+		},
+		conf.GetTracesConfig().GetSendTickerValue()*60, // waitFor
+		conf.GetTracesConfig().GetSendTickerValue()*2,  // tick
+		"within a reasonable time after adding a decision span, we should have 2 descendants in the trace",
+	)
 
 	trace := coll.getFromCache(traceID)
 	require.NotNil(t, trace)
-	assert.Equal(t, int(trace.DescendantCount()), 2, "after adding a decision span, we should have 2 descendant in the trace")
 	assert.Equal(t, trace.SendBy.Unix(), sendBy, "send_by should be the same as the one set from its peer")
```
VinozzZ, line 496: Should we put all these into the eventuallyWithT instead? We probably should since the trace _could_ be removed from the cache after we checked in the assert.Eventually and before the `coll.getFromCache` on line 494

robbkidd: 🤔 

Is there some—like—convergence to a state here? Do some of these assertions fail when some succeed, but we would want to wait until all succeed? 

Or is "we've got two descendants" enough to know that the rest of the assertions that follow ought to also succeed?

VinozzZ: I believe this is because we want to test it before the trace is taken out of the cache for a sampling decision.

robbkidd: Does the sampling decision get triggered by the root span getting added? That occurs after these assertions whether they are asserted inside or outside the `EventuallyWithT()`.

VinozzZ: Oh, right. I guess that's ok then. Another way the sampling decision get triggered is through TraceTimeout, which is set to 60s in this test. Hopefully we won't ever reach that

VinozzZ (approved):

PR #1640, opened 2025-07-29
Fix: Ignore computed fields during sampling key extraction

VinozzZ:

Refinery was attempting to extract computed fields (prefixed with `?.`) like `?.NUMBER_DESCENDANTS` as if they were regular trace fields during the unmarshaling process. This caused unnecessary performance overhead since computed fields are virtual fields calculated during rule evaluation and don't exist as actual fields in trace data.

- Ignore computed fields in `GetKeyFields` function

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -242,6 +242,7 @@ const RootPrefix = "root."
 // GetKeyFields returns the fields that should be used as keys for the sampler.
 // It returns two slices: the first contains all fields, including those with the root prefix,
 // and the second contains fields that do not have the root prefix.
+// Fields that start with "computed." are ignored since they should not exist as a field in a trace.
 func GetKeyFields(fields []string) (allFields []string, nonRootFields []string) {
 	if len(fields) == 0 {
 		return nil, nil
@@ -251,11 +252,26 @@ func GetKeyFields(fields []string) (allFields []string, nonRootFields []string) 
 	nonRootFields = make([]string, 0, len(fields))
 
 	for _, field := range fields {
-		if strings.HasPrefix(field, RootPrefix) {
-			rootFields = append(rootFields, field[len(RootPrefix):])
-		} else {
+		values := strings.SplitAfterN(field, ".", 2)
+		if len(values) < 2 {
+			nonRootFields = append(nonRootFields, field)
+			continue
+		}
+
+		switch values[0] {
+		case RootPrefix:
+			// If the field starts with "root.", add it to rootFields
+			rootFields = append(rootFields, values[1])
+		case ComputedFieldPrefix:
+			// If the field starts with "computed.", skip it
+		default:
+			// Otherwise, add it to nonRootFields
 			nonRootFields = append(nonRootFields, field)
 		}
+	}
+
+	if len(rootFields) == 0 && len(nonRootFields) == 0 {
+		return nil, nil
 	}
 
 	if len(rootFields) == 0 {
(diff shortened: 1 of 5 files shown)


config/config.go
```diff
@@ -242,6 +242,7 @@ const RootPrefix = "root."
 // GetKeyFields returns the fields that should be used as keys for the sampler.
 // It returns two slices: the first contains all fields, including those with the root prefix,
 // and the second contains fields that do not have the root prefix.
+// Fields that start with "computed." are ignored since they should not exist as a field in a trace.
```
kentquirk: Both here and below, it's not "computed" it's "?". Right?

kentquirk (approved):
Basic concept is sound, but I think it's worth a quick performance check.

Gah, my comment got lost! I don't think the `SplitAfterN` function is going to be anywhere near as fast as a simple HasPrefix in the switch statement. I think you should benchmark it to make sure.

```
			switch {
			case strings.HasPrefix(key, RootPrefix):
```

VinozzZ:
> Basic concept is sound, but I think it's worth a quick performance check.
> 
> Gah, my comment got lost! I don't think the `SplitAfterN` function is going to be anywhere near as fast as a simple HasPrefix in the switch statement. I think you should benchmark it to make sure.
> 
> ```
> 			switch {
> 			case strings.HasPrefix(key, RootPrefix):
> ```

I tried out the string.HasPrefix version. The benchmark is surprising that the `SplitAfterN` is faster
```
goos: darwin
goarch: arm64
pkg: github.com/honeycombio/refinery/types
cpu: Apple M2 Max
                                             │   split-after.txt   │              has-prefix.txt              │
                                             │   sec/op    │   sec/op     vs base                │
UnmarshalPayload/UnmarshalPayload-12           16.25µ ± 1%   17.82µ ± 1%   +9.67% (p=0.000 n=10)
UnmarshalPayload/UnmarshalPayloadComplete-12   15.60µ ± 1%   17.43µ ± 1%  +11.77% (p=0.000 n=10)
geomean                                        15.92µ        17.63µ       +10.72%

                                             │   new.txt    │               new-2.txt               │
                                             │     B/op     │     B/op      vs base                 │
UnmarshalPayload/UnmarshalPayload-12           6.974Ki ± 0%   6.974Ki ± 0%       ~ (p=1.000 n=10) ¹
UnmarshalPayload/UnmarshalPayloadComplete-12     997.0 ± 0%     997.0 ± 0%       ~ (p=1.000 n=10) ¹
geomean                                        2.606Ki        2.606Ki       +0.00%
¹ all samples are equal

                                             │  new.txt   │              new-2.txt              │
                                             │ allocs/op  │ allocs/op   vs base                 │
UnmarshalPayload/UnmarshalPayload-12           13.00 ± 0%   13.00 ± 0%       ~ (p=1.000 n=10) ¹
UnmarshalPayload/UnmarshalPayloadComplete-12   12.00 ± 0%   12.00 ± 0%       ~ (p=1.000 n=10) ¹
geomean                                        12.49        12.49       +0.00%
¹ all samples are equal
```

config/config.go
```diff
@@ -251,13 +252,22 @@ func GetKeyFields(fields []string) (allFields []string, nonRootFields []string)
 	nonRootFields = make([]string, 0, len(fields))
 
 	for _, field := range fields {
-		if strings.HasPrefix(field, RootPrefix) {
+		switch {
+		case field[0] == 'r' && strings.HasPrefix(field, RootPrefix):
```
kentquirk: Do you want to put these runes into constants next to the definitions of the full name constants, just so they're together?

PR #1643, opened 2025-07-29
feat: use husky's optimized direct translation for GRPC calls

ianwilkes:
GRPC wants to do all unmarshaling itself based on stock generated code, which doesn't play nicely with our custom, optimized data path.

Replaces the stock generated OTLP trace server with a new one which conforms to GRPC's, shall we say, idiosyncratic interface, but which uses husky's optimized translation instead of unmarshaling. This is a little awkward since (in some cases) we depend on the request info to extract the dataset, so request validation is now happening during Unmarshal, before any grpc interceptors get a look at the request. Fortunately we're not using any interceptors at the moment. The advantage of doing it this early is that we never need to make a copy of the input buffer.

Also moves some testing to work with the full GRPC stack, to remove dependencies on the details of our implementation. Also adds a GRPC concurrency test, which can help expose race conditions.

Benchmark results are promising, note this looks slower than HTTP mainly because of overhead in the benchmark itself, where the GRPC client is re-serializing the request on every iteration:
```
benchmark                         old ns/op     new ns/op     delta
BenchmarkTracesOTLP/grpc          564784        329547        -41.65%
benchmark                         old allocs    new allocs    delta
BenchmarkTracesOTLP/grpc          5070          909           -82.07%
benchmark                         old bytes     new bytes     delta
BenchmarkTracesOTLP/grpc          322256        119653        -62.87%
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -119,4 +119,4 @@ tool github.com/google/go-licenses/v2
 
 replace go.opentelemetry.io/proto/otlp => github.com/honeycombio/opentelemetry-proto-go/otlp v1.3.1-compat
 
-replace github.com/honeycombio/husky => github.com/honeycombio/husky v0.37.1-0.20250729164345-ed6a8be4045a
+replace github.com/honeycombio/husky => github.com/honeycombio/husky v0.37.1-0.20250729215257-eb95c950c0e9
--- a/go.mod
+++ b/go.mod
@@ -15,7 +15,7 @@ require (
 	github.com/hashicorp/golang-lru/v2 v2.0.7
 	github.com/honeycombio/dynsampler-go v0.6.3
 	github.com/honeycombio/hpsf v0.5.1-0.20250605195630-9b072f9dbd61
-	github.com/honeycombio/husky v0.36.0
+	github.com/honeycombio/husky v0.37.1-0.20250730154742-ddf9db5b5fe7
 	github.com/honeycombio/libhoney-go v1.25.0
 	github.com/jessevdk/go-flags v1.6.1
 	github.com/jonboulle/clockwork v0.5.0
@@ -118,5 +118,3 @@ require (
 tool github.com/google/go-licenses/v2
 
 replace go.opentelemetry.io/proto/otlp => github.com/honeycombio/opentelemetry-proto-go/otlp v1.3.1-compat
-
-replace github.com/honeycombio/husky => github.com/honeycombio/husky v0.37.1-0.20250729215257-eb95c950c0e9
--- a/route/otlp_trace.go
+++ b/route/otlp_trace.go
@@ -6,6 +6,7 @@ import (
 
 	huskyotlp "github.com/honeycombio/husky/otlp"
 	"github.com/honeycombio/refinery/internal/otelutil"
+	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 
@@ -77,9 +78,15 @@ func (r *Router) processOTLPRequestWithMsgp(ctx context.Context, w http.Response
 	}
 }
 
+// CustomTraceServiceServer defines our custom interface for the trace service.
+// It processes pre-translated trace data using msgpack optimization rather than
+// accepting raw OTLP protobuf requests.
+type CustomTraceServiceServer interface {
+	ExportTraceData(context.Context, *huskyotlp.TranslateOTLPRequestResultMsgp, huskyotlp.RequestInfo) (*collectortrace.ExportTraceServiceResponse, error)
+}
+
 type TraceServer struct {
 	router *Router
-	collectortrace.UnimplementedTraceServiceServer
 }
 
 func NewTraceServer(router *Router) *TraceServer {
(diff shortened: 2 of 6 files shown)


go.mod
```diff
@@ -118,5 +118,3 @@ require (
 tool github.com/google/go-licenses/v2
 
 replace go.opentelemetry.io/proto/otlp => github.com/honeycombio/opentelemetry-proto-go/otlp v1.3.1-compat
-
-replace github.com/honeycombio/husky => github.com/honeycombio/husky v0.37.1-0.20250729164345-ed6a8be4045a
```
VinozzZ, line 122: Why do we remove the replace here?

route/otlp_trace.go
```diff
@@ -77,42 +78,115 @@ func (r *Router) processOTLPRequestWithMsgp(ctx context.Context, w http.Response
...
+// CustomTraceServiceServer defines our custom interface for the trace service.
+// It processes pre-translated trace data using msgpack optimization rather than
+// accepting raw OTLP protobuf requests.
+type CustomTraceServiceServer interface {
+	ExportTraceData(context.Context, *huskyotlp.TranslateOTLPRequestResultMsgp, huskyotlp.RequestInfo) (*collectortrace.ExportTraceServiceResponse, error)
+}
+
 type TraceServer struct {
 	router *Router
-	collectortrace.UnimplementedTraceServiceServer
```
VinozzZ, line 82: Why do we remove this helper function?

route/otlp_trace.go
```diff
@@ -77,42 +78,115 @@ func (r *Router) processOTLPRequestWithMsgp(ctx context.Context, w http.Response
...
+	// Use optimized msgpack processing path with already translated data
+	if err := t.router.processOTLPRequestBatchMsgp(ctx, result.Batches, ri.ApiKey, ri.UserAgent); err != nil {
 		return nil, huskyotlp.AsGRPCError(err)
 	}
 
-	ri.ApiKey = keyToUse
+	return &collectortrace.ExportTraceServiceResponse{}, nil
+}
+
+// translatedTraceServiceRequest is a wrapper that implements proto.Message but
```
VinozzZ: can we update the comment to point to the [v1 interface](https://pkg.go.dev/google.golang.org/protobuf/runtime/protoiface#MessageV1) ?

route/otlp_trace.go
```diff
@@ -77,42 +78,115 @@ func (r *Router) processOTLPRequestWithMsgp(ctx context.Context, w http.Response
...
+// actually produces a translated TranslateOTLPRequestResultMsgp instead of a
+// real unserialized protobuf struct.
+type translatedTraceServiceRequest struct {
+	requestInfo huskyotlp.RequestInfo
+	result      *huskyotlp.TranslateOTLPRequestResultMsgp
+	ctx         context.Context
+}
+
+// Implement proto.Message interface.
```
VinozzZ: same as above

route/otlp_trace.go
```diff
@@ -77,42 +78,115 @@ func (r *Router) processOTLPRequestWithMsgp(ctx context.Context, w http.Response
...
+}
+
+func (r *translatedTraceServiceRequest) String() string {
+	return "translatedTraceServiceRequest"
+}
 
-	result, err := huskyotlp.TranslateTraceRequest(ctx, req, ri)
+func (r *translatedTraceServiceRequest) ProtoMessage() {}
+
+// Unmarshal implements a custom unmarshaling method that performs translation
```
VinozzZ: Can we also include the `proto.Method` doc and the [`Unmarshaler`](https://github.com/protocolbuffers/protobuf-go/blob/8e8926ef675d99b1c9612f5d008f4dc803839f7a/internal/impl/legacy_message.go#L381) here to imply that the Unmarshal method is implementing this interface?

route/otlp_trace_test.go
```diff
@@ -36,10 +39,62 @@ import (
...
+}
 
+// createGRPCContext creates a context with GRPC metadata from the provided headers.
+func createGRPCContext(headers map[string]string) context.Context {
+	md := metadata.New(headers)
+	return metadata.NewOutgoingContext(context.Background(), md)
+}
+
+// makeGRPCCall executes a trace export request via GRPC and returns any error.
+func makeGRPCCall(
```
VinozzZ: Do we really need this helper function?

go.mod
```diff
@@ -15,7 +15,7 @@ require (
 	github.com/hashicorp/golang-lru/v2 v2.0.7
 	github.com/honeycombio/dynsampler-go v0.6.3
 	github.com/honeycombio/hpsf v0.5.1-0.20250605195630-9b072f9dbd61
-	github.com/honeycombio/husky v0.36.0
+	github.com/honeycombio/husky v0.37.1-0.20250730154742-ddf9db5b5fe7
```
VinozzZ, line 18: We can update this to point to the latest commit from `main` now

go.mod
ianwilkes: Don't need it anymore since the changes we depend on are landed to husky's main branch.

route/otlp_trace.go
ianwilkes: We don't need it, and its definition conflicts with ours. I'm not sure why it was there to begin with.

VinozzZ: Based on the doc, it's to provide forward compatibility. I guess we don't support that anymore anyway

go.mod
ianwilkes: That's what this is.

route/otlp_trace.go
ianwilkes: "Forward compatibility" meaning, I guess, that they can introduce new methods to the interface and we automatically get an "unimplemented" implementation so the code still "works". I don't see that as a good thing at all.

VinozzZ (approved):

PR #1644, opened 2025-07-30
perf: optimized JSON batch handling

ianwilkes:
JSON is still a thing, so we need to handle it efficiently.

Short-circuits the batch JSON parsing pathway to directly pick out the fields we want using optimized code, and then turn the data blob directly into messagepack. That makes this slower than messagepack, but faster than OTLP, which isn't bad.

```
benchmark                              old ns/op     new ns/op     delta
BenchmarkRouterBatch/batch_json-12     346258        26287         -92.41%

benchmark                              old allocs     new allocs     delta
BenchmarkRouterBatch/batch_json-12     1935           30             -98.45%

benchmark                              old bytes     new bytes     delta
BenchmarkRouterBatch/batch_json-12     113436        17539         -84.54%
```

### DIFF
--- a/route/batched_event.go
+++ b/route/batched_event.go
@@ -4,12 +4,14 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"sync"
 	"time"
+
+	"github.com/tinylib/msgp/msgp"
+	"github.com/valyala/fastjson"
 
 	"github.com/honeycombio/refinery/config"
 	"github.com/honeycombio/refinery/types"
-	"github.com/tinylib/msgp/msgp"
-	"github.com/valyala/fastjson"
 )
 
 var fastJsonParserPool fastjson.ParserPool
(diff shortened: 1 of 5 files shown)


route/batched_event.go
```diff
@@ -163,22 +142,99 @@ func (b *batchedEvents) MarshalJSON() ([]byte, error) {
...
+	var dataValue *fastjson.Value
+	obj.Visit(func(key []byte, v *fastjson.Value) {
+		if err != nil {
+			return
+		}
+
+		switch string(key) {
+		case "time":
+			if v.Type() == fastjson.TypeString {
+				s, _ := v.StringBytes()
```
VinozzZ: It looks like we can use `GetStringBytes` here if we don't care about the returned error

route/batched_event.go
```diff
@@ -163,22 +142,99 @@ func (b *batchedEvents) MarshalJSON() ([]byte, error) {
...
+
+	obj, err := v.Object()
+	if err != nil {
+		return err
+	}
+
+	// Visit each field in the event object
+	var dataValue *fastjson.Value
+	obj.Visit(func(key []byte, v *fastjson.Value) {
+		if err != nil {
```
VinozzZ: Do we need this call here? I don't think the err is reset anywhere in the Visit function

ianwilkes: I have no idea how that got there.

VinozzZ (approved):

PR #1650, opened 2025-08-05
test: improve efficiency of BenchmarkCollectorWithSamplers

ianwilkes:
This benchmark was totally bogged down in creating test data points, making it slow to run and innaccurate.

Changes the structure of the benchmark so that N is the number of spans, not the number of traces. Gives up on the 1M-span trace which I think is not really necessary to get the idea of the performance curve. Re-uses spans.

At some point the input will again become the bottleneck here, but multiple input goroutines really slowed things down so I stuck with just the one - for now anyway.

### DIFF
--- a/collect/collect_test.go
+++ b/collect/collect_test.go
@@ -3,6 +3,7 @@ package collect
 import (
 	"context"
 	"fmt"
+	"math"
 	"math/rand"
 	"runtime"
 	"strconv"
@@ -2547,7 +2548,11 @@ func TestSpanLimitSendByPreservation(t *testing.T) {
 
 }
 
-// BenchmarkCollectorWithSamplers runs benchmarks for different sampler configurations
+// BenchmarkCollectorWithSamplers runs benchmarks for different sampler configurations.
+// This is a tricky benchmark to interpret because just setting up the input data
+// can easily be more expensive than the collector's routing code. The goal is to
+// demonstrate the collector's (in-)ability to scale across CPUs. It does NOT work
+// well with just one CPU.
 func BenchmarkCollectorWithSamplers(b *testing.B) {
 	// Common test scenarios to run for each sampler
 	scenarios := []struct {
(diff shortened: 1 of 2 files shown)


collect/collect_test.go
```diff
@@ -2618,57 +2622,35 @@ func BenchmarkCollectorWithSamplers(b *testing.B) {
...
 				}
-				collector := setupBenchmarkCollector(b, sampler.config, sender)
+				collector := setupBenchmarkCollector(sampler.config, sender)
 				collector.Start()
 
 				defer collector.Stop()
 
-				// use b.N as the number of traces to create
-				totalSpans := b.N * scenario.spansPerTrace
+				b.N = scenario.spansPerTrace * int(math.Ceil(float64(b.N)/float64(scenario.spansPerTrace)))
```
VinozzZ, line 2632: I didn't know you can overwrite `b.N`

collect/collect_test.go
```diff
@@ -2679,75 +2661,105 @@ func BenchmarkCollectorWithSamplers(b *testing.B) {
...
-
-	// Array field
-	data["tags"] = []string{
-		fmt.Sprintf("tag_%d", spanIdx),
-		fmt.Sprintf("tag_%d", spanIdx+1),
-		fmt.Sprintf("tag_%d", spanIdx+2),
+	span.Event.Data.Set("trace.trace_id", traceID)
+	span.Event.Data.Set("trace.span_id", fmt.Sprintf("span-%018d", spanIdx))
+	if isRoot {
+		span.Event.Data.UnsetForTest("trace.parent_id")
```
VinozzZ, line 2756: If it's root, I think we also need to unset the `IsRoot` field on the span as well as the `MetaIsRoot` field on the `Payload`

collect/collect_test.go
ianwilkes: You can, and the benchmark framework uses the new count. I don't know if this is strictly kosher since it doesn't seem to be documented, but it works.

VinozzZ (approved):

PR #1651, opened 2025-08-06
maint: rename incoming/peer_router_otlp to incoming/peer_router_otlp_other

VinozzZ:

To make the metric name more descriptive for users who doesn't know the internal implementation details, I would like to change the metric that is tracking the otlp request sent via http other than JSON or protobuf as `incoming/peer_router_otlp_other` 

- rename incoming/peer_router_otlp to incoming/peer_router_otlp_other


metrics.md
```diff
@@ -108,7 +108,7 @@ Metrics in this table don't contain their expected prefixes. This is because the
 | _router_peer | Counter | Dimensionless | the number of spans proxied to a peer |
 | _router_batch | Counter | Dimensionless | the number of batches of events received |
 | _router_batch_events | Counter | Dimensionless | the number of events received in batches |
-| _router_otlp | Counter | Dimensionless | the number of otlp requests received |
+| _router_otlp_other | Counter | Dimensionless | the number of otlp requests not using optimized http json/protobuf processing |
```
dustxd: Since all the other descriptions have some sort of verbs like "received", "dropped", etc. I could be totally missing what this is doing but I'm wondering if the following would make sense? 

```suggestion
| _router_otlp_other | Counter | Dimensionless | the number of otlp requests received without http json/protobuf processing optimization |
```

dustxd: Oh or something like this based on your PR description...?

`the number of OTLP requests received via HTTP using neither JSON nor Protobuf`

dustxd (approved):
the other one

robbkidd (approved):
look at those numbers

PR #1652, opened 2025-08-07
maint: update streamlined data handling branch to the latest changes in main

VinozzZ:

Before we can bring in all commits from the performance-work-branch into main, we need to update the perf branch to be update-to-date with main

- stuff from 2.9.7
- a few changes after 2.9.7
- merge conflict fix: include recording that a payload is empty during field memoization instead of checking data lengths later


robbkidd (approved):
merge is coming

PR #1653, opened 2025-08-07
perf: streamline data handling by reducing the number of data transforms

robbkidd:
Bring in all the improvements from the performance-work-branch. Intentionally merged with a merge commit and not squashed to preserve the history of discrete changes reviewed and merged into that feature branch.


VinozzZ (approved):
🚀

jiekun:
Hello! Hope it’s not too late to ask a question. I was impressed by this excellent blog post: https://www.honeycomb.io/blog/how-we-saved-70-cpu-60-memory-refinery

May I ask whether the memory reduction is mainly achieved by better allocation practices, while the raw data still remains in memory before being discarded?
That is to say, the memory usage should roughly represented by:
- the size of the raw data (which seems necessary)
- plus the extra overhead for mapping, sampling, and other processing.

And the memory reduction mainly comes from optimizing the latter overhead part?

MikeGoldsmith:
> Hello! Hope it’s not too late to ask a question. I was impressed by this excellent blog post: https://www.honeycomb.io/blog/how-we-saved-70-cpu-60-memory-refinery
> 
> May I ask whether the memory reduction is mainly achieved by better allocation practices, while the raw data still remains in memory before being discarded? That is to say, the memory usage should roughly represented by:
> 
> * the size of the raw data (which seems necessary)
> * plus the extra overhead for mapping, sampling, and other processing.
> 
> And the memory reduction mainly comes from optimizing the latter overhead part?

Yes, the main theory was not to be smart about what attributes we were deserialising from the raw bytes. Previously we deserialised everything, now we only grab what we need. The original byte array is still available and the whole event will be passed through refinery if the trace is selected to be sampled.

PR #1655, opened 2025-08-07
test: allow tests to pass when run multiple times

ianwilkes:
It should be possible for refinery tests to pass when run with `-count` > 1, but currently the don't.

Just skip additional runs of the tests that set global prometheus state since it panics on a duplicate metric. And, remove the assertion of the initial state of libhoney.UserAgentAddition because I don't think we care.


robbkidd (commented):
(This was supposed to be an approval, but mobile got weird.)

robbkidd (approved):
fair enough

PR #1656, opened 2025-08-07
fix: error handling in postOTLPTrace

VinozzZ:

Return an error response if an otlp http trace request can't be successfully processed

- Only `JSON` and `Protobuf` is supported for the otlp http trace endpoint. If the content type is not either of those, we should return an error
- add tests


route/otlp_trace_test.go
```diff
@@ -666,6 +666,124 @@ func TestOTLPHandler(t *testing.T) {
...
+			request, _ := http.NewRequest("POST", "/v1/traces", bytes.NewReader(invalidBody))
+			request.Header = http.Header{}
+			request.Header.Set("content-type", "application/protobuf")
+			request.Header.Set("x-honeycomb-team", legacyAPIKey)
+			request.Header.Set("x-honeycomb-dataset", "dataset")
+
+			w := httptest.NewRecorder()
+			router.postOTLPTrace(w, request)
+
+			assert.Equal(t, http.StatusInternalServerError, w.Code)
```
robbkidd: 🤔 Hmm. We ought to be returning HTTP errors in the 4xx range for these cases where the content is malformed or of a type we do not support. These are not 5xx (our fault) errors.

robbkidd (approved):
error

robbkidd (approved):
more simple

PR #1661, opened 2025-08-12
fix: make sure all transmit metrics have correct libhoney prefix

VinozzZ:

We would like to preserve all existing `libhoney` metric names so that users can compare the performance between version upgrades easily.

- add `libhoney_` prefix to all transmit metrics


TylerHelmuth (approved):

TylerHelmuth:
@VinozzZ are you sure this isn't a breaking change? Are the metrics prefixed like this in 2.9.7?

VinozzZ:
> @VinozzZ are you sure this isn't a breaking change? Are the metrics prefixed like this in 2.9.7?

This is to make sure the metric names is the same as 2.9.7 so that we can avoid a breaking change

robbkidd (approved):
put honey on everything

PR #1662, opened 2025-08-13
fix: preserve draining during shutdown

VinozzZ:

Draining in-flight data during shutdown should not be tied with redistribution during cluster peer membership change

- remove the `DisableRedistribution` config check from `Shutdown`


VinozzZ:
we need to talk more about the shutdown behavior

PR #1665, opened 2025-08-14
fix: race condition in DirectTransmit for metrics registration

VinozzZ:

We are referencing `Metrics` object in `Start` method in `DirectTransmit`. It has a race condition between writes in `RegisterMetrics` and accessing metrics in `Start` method.

- Make `metrics` as a dependency for `DirectTransmit` using the injection library.


robbkidd (approved):
renew the deal with the DI devil

Issue #1642, opened 2025-07-29
If lastversion hasn't been reached yet, log a deprecation warning at startup


kentquirk:
**Is your feature request related to a problem? Please describe.**

- Customer got confused by CacheCapacity, which no longer does what it used to do, but it is still used to establish default values for OTHER fields. 
- We can't just turn it off because it would be a breaking change since our validation would prevent its use.

**Describe the solution you'd like**

- Refinery "lastversion" flag should log a warning whenever it's been specified and Refinery is not at that version yet. 
- Set it on CacheCapacity and add a "deprecationText" field to the YAML format for config; the contents of that field would be used in the log message. This would allow us to have a planned deprecation where users get a warning a couple of versions in advance, without breaking things.

**Describe alternatives you've considered**

- Never obsoleting things

**Additional context**

PR #1667, opened 2025-08-15
feat: config option deprecation

VinozzZ:

fix: #1642 
In the issue, the proposed solution is to return error if the `latestversion` defined for a config option is older than the current version. In this PR, we will hold off on that change since this is the first time we introduce the concept of deprecation in Refinery. All deprecated configs will log warnings for now.

- Add a warning log for config fields that have `lastversion` value set and the current Refinery version is before `lastversion`
- Log a warning for config fields that have `latestversion` value set and the current Refinery version is after `latestversion`
- Suuport custom deprecation message through `deprectiontext` field in `configMeta.yaml`

## Example output:
```
ERROR: Validation failed for config [tools/convert/minimal_config.yaml]:
  field General.DatasetPrefix (%%) must be purely alphanumeric
WARNING: Configuration warnings for [tools/convert/minimal_config.yaml]:
  field BufferSizes.PeerBufferSize is deprecated since version v2.9.7
  CacheCapacity is deprecated since version v2.9.7. Set PeerQueueSize and IncomingQueueSize instead.
  field BufferSizes.UpstreamBufferSize is deprecated since version v2.9.7
  Please update your configuration following the latest documentation here: https://docs.honeycomb.io/manage-data-volume/sample/honeycomb-refinery/configure/
```


kentquirk (commented):
Did you see #1642? The proposal there was to allow setting LastVersion to a future version number, so that we can pre-warn people that something will be going away. 

If lastversion has already been reached, isn't it the case that the field simply won't pass validation?

config/configLoadHelpers.go
```diff
@@ -391,42 +391,42 @@ func loadConfigsIntoMap(dest map[string]any, configs []configData) error {
 // validateConfigs gets the configs from the given location and validates them.
 // It returns a list of failures; if the list is empty, the config is valid.
 // err is non-nil only for significant errors like a missing file.
-func validateConfigs(configs []configData, opts *CmdEnv) ([]string, error) {
+func validateConfigs(configs []configData, opts *CmdEnv, currentVersion ...string) (warnings []string, errors []string, err error) {
```
kentquirk: I'm really not fond of this change of signature. Instead of having two buckets of warnings and errors, I think it would be more useful to change the return type from []string. Maybe something like this?

```go
type ValidationResult {
  Message string
  Severity int    // an enum with Warning and Error at least
}
type ValidationResults []ValidationResult
func (v ValidationResults) isOnlyWarnings() {...}
```

config/validate.go
```diff
@@ -235,8 +235,10 @@ func maskString(s string) string {
 // If a field name is of type "object" then this function is called
 // recursively to validate the sub-object.
 // Note that the flatten function returns only 2 levels.
-func (m *Metadata) Validate(data map[string]any) []string {
-	errors := make([]string, 0)
+
+func (m *Metadata) Validate(data map[string]any, currentVersion ...string) (warnings []string, errors []string) {
```
kentquirk: See comment in configLoadHelpers.

config/version.go
```diff
@@ -0,0 +1,43 @@
...
+// compareVersions compares two version strings using semantic versioning
+// Returns:
+//
+//	-1 if current < target
+//	 0 if current == target
+//	 1 if current > target
+//	error if versions are malformed
+func compareVersions(current, target string) (int, error) {
+	// Ensure versions have 'v' prefix for semver package
+	if !semver.IsValid(current) {
```
kentquirk, line 17: Why do we need to do this? All our versions have a v anyway.

kentquirk (approved):
Not wanting to block you, so this is approved, but I also would prefer if it were changed a bit. I don't feel entirely comfortable with these function signatures.

PR #1669, opened 2025-08-15
fix: quick and dirty concurrency safety for samplers

ianwilkes:
Some of our samplers are concurrency-safe, some are not. They need to be for the new concurrent collect loop implementation.

Adds concurrent tests for each sampler, and adds locks where necessary. Also updates to dynsampler-go HEAD, which has a concurrency fix.

This was largely vibe-coded and is going to introduce some bottlenecks given the aggressive locking; I'd like to find a better solution for thread-local key-building and metrics. But this is a minimal fix to unblock my other work.

### DIFF
--- a/sample/totalthroughput.go
+++ b/sample/totalthroughput.go
@@ -28,7 +28,6 @@ type TotalThroughputSampler struct {
 	key                      *traceKey
 	keyMu                    sync.Mutex
 	clusterMu                sync.Mutex
-	dynsamplerMu             sync.Mutex
 	keyFields, nonRootFields []string
 
 	dynsampler      *dynsampler.TotalThroughput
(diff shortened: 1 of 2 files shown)


sample/ema_throughput_test.go
```diff
@@ -50,3 +53,167 @@ func TestEMAThroughputAddSampleRateKeyToTrace(t *testing.T) {
...
+			}
+
+			// Test concurrent SetClusterSize calls if UseClusterSize is enabled
+			if tc.config.UseClusterSize {
+				wg.Add(1)
+				go func() {
+					defer wg.Done()
+					for i := 0; i < 20; i++ {
+						clusterSize := (i%5) + 1 // Cluster sizes 1-5
+						sampler.SetClusterSize(clusterSize)
```
VinozzZ, line 210: Should we also test concurrent read and write combination?

ianwilkes: That's happening here - this is running concurrently with the GetSampleRate() calls above, which I think ultimately depend on this value.

VinozzZ (approved):

PR #1670, opened 2025-08-15
feat: initial concurrent collect loops

ianwilkes:
The single collect loop is a giant bottleneck.

First stab at concurrent collect loops, with mutexes all over the place to protect shared resources. This has a lot of bottlenecks; if most of the collect loop time is being spent in sampler decison-making, we'll probably want each loop to have its own set of sampler instances.

Includes a bunch of new testing and some light test cleanup.

### DIFF
--- a/collect/collect.go
+++ b/collect/collect.go
@@ -6,16 +6,17 @@ import (
 	"fmt"
 	"os"
 	"runtime"
-	"sort"
 	"sync"
 	"time"
 
-	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
+
+	"github.com/dgryski/go-wyhash"
+	"github.com/jonboulle/clockwork"
+	"github.com/sirupsen/logrus"
 
 	"github.com/honeycombio/refinery/collect/cache"
 	"github.com/honeycombio/refinery/config"
-	"github.com/honeycombio/refinery/generics"
 	"github.com/honeycombio/refinery/internal/health"
 	"github.com/honeycombio/refinery/internal/otelutil"
 	"github.com/honeycombio/refinery/internal/peer"
@@ -26,8 +27,6 @@ import (
 	"github.com/honeycombio/refinery/sharder"
 	"github.com/honeycombio/refinery/transmit"
 	"github.com/honeycombio/refinery/types"
-	"github.com/jonboulle/clockwork"
-	"github.com/sirupsen/logrus"
 )
 
 const (
@@ -76,7 +75,7 @@ type sendableTrace struct {
 	samplerSelector string
 }
 
-// InMemCollector is a single threaded collector.
+// InMemCollector is a collector that can use multiple concurrent collection loops.
 type InMemCollector struct {
 	Config  config.Config   `inject:""`
 	Logger  logger.Logger   `inject:""`
@@ -97,20 +96,20 @@ type InMemCollector struct {
 	TestMode       bool
 	BlockOnAddSpan bool
 
+	// Parallel collection support
+	collectLoops []*CollectLoop
+
 	// mutex must be held whenever non-channel internal fields are accessed.
-	// This exists to avoid data races in tests and startup/shutdown.
 	mutex sync.RWMutex
-	cache cache.Cache
 
 	datasetSamplers map[string]sample.Sampler
 
 	sampleTraceCache cache.TraceSentCache
 
 	shutdownWG     sync.WaitGroup
-	incoming       chan *types.Span
-	fromPeer       chan *types.Span
+	collectLoopsWG sync.WaitGroup // Separate WaitGroup for collect loops
+	reload         chan struct{}  // Channel for config reload signals
 	outgoingTraces chan sendableTrace
-	reload         chan struct{}
 	done           chan struct{}
 
 	hostname string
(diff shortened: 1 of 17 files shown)


collect/collect.go
```diff
@@ -175,12 +175,18 @@ func (i *InMemCollector) Start() error {
...
+
+	i.Logger.Info().WithField("num_loops", numLoops).Logf("Starting InMemCollector with %d collection loops", numLoops)
+
 	i.StressRelief.UpdateFromConfig()
 
 	// listen for config reloads
 	i.Config.RegisterReloadCallback(i.sendReloadSignal)
 
+	// Find or create a test, make sure we signal health based (somehow)
+	// on all the collect loops running.
```
VinozzZ, line 190: The purpose of this health check is to verify that the collector hasn’t panicked or stopped running unexpectedly. One obvious solution would be for each collector loop to register itself with the health check. The downside is that if any single loop stops, Refinery will report itself as unhealthy. That might be acceptable, since we currently don’t have a recovery mechanism if a loop dies anyway.

collect/collect_test.go
```diff
@@ -1342,12 +1268,18 @@ func TestAddAdditionalAttributes(t *testing.T) {
 	}
 	coll.AddSpan(rootSpan)
 
+	// Wait for events to be transmitted
+	assert.Eventually(t, func() bool {
```
VinozzZ: Why do we need to check for `len(transmission.Events)` explicitly? Doesn't `transmission.GetBlock()` wait for events to be sent anyway?

collect/multi_loop_test.go
```diff
@@ -0,0 +1,602 @@
...
+				foundCount := 0
+				for i := 0; i < numTraces; i++ {
+					traceID := fmt.Sprintf("trace-%d", i)
+					if _, exists := allTraces[traceID]; exists {
+						foundCount++
+					}
+				}
+
+				// We should find most traces (some may have expired)
+				return foundCount >= numTraces/2
```
VinozzZ: If we set `TraceTimeout` to something like 60s, then we should be able to assert all traces are found here

collect/multi_loop_test.go
```diff
@@ -0,0 +1,602 @@
...
+			}, 5*time.Second, 500*time.Millisecond) // Check less frequently since each check is now fast
+
+			met := collector.Metrics.(*metrics.MockMetrics)
+			count, ok := met.Get("span_received")
+			assert.True(t, ok)
+			assert.Equal(t, float64(1000), count)
```
VinozzZ:
```suggestion
			assert.Equal(t, float64(spanPerTrace*numTraces), count)
```

collect/multi_loop_test.go
```diff
@@ -0,0 +1,602 @@
...
+
+	// Recreational reload command, should be fine.
+	collector.sendReloadSignal("hash1", "hash2")
+
+	// Wait for all input goroutines to complete - note this will deadlock if
+	// we fill the input channels.
+	wg.Wait()
+
+	// Now that all the events have been enqueued, advance time to allow transmisison.
+	clock.Advance(time.Second)
```
VinozzZ, line 389: Do we need to call `BlockUntilContext` before we call `Advance` so that we know there's ticker or timer waiting for the signal?

collect/multi_loop_test.go
```diff
@@ -0,0 +1,602 @@
...
+
+	// Wait for spans to be processed and sent
+	var totalSent int
+	assert.Eventually(t, func() bool {
+		clock.Advance(time.Second)
+		events := transmission.GetBlock(0)
+		totalSent += len(events)
+
+		// Sample rate of 2 in the confix means we expect roughly half of our
```
VinozzZ:
```suggestion
		// Sample rate of 2 in the config means we expect roughly half of our
```

collect/multi_loop_test.go
```diff
@@ -0,0 +1,602 @@
...
+					clock.Advance(10 * time.Millisecond)
+				}
+			}
+		}(g)
+	}
+
+	wg.Wait()
+
+	// Verify the collector handled memory pressure
+	assert.Greater(t, atomic.LoadInt32(&spansAdded), int32(0), "Should have added spans")
```
VinozzZ, line 493: Maybe we can assert the `collector_cache_eviction` metric to at least check that the `checkAlloc` code has been triggered

config/metadata/configMeta.yaml
```diff
@@ -1454,6 +1454,21 @@ groups:
         description: >
           Sets the time interval for sending accumulated kept decisions in batches. This ensures that kept decisions are processed at regular intervals, improving efficiency by reducing the frequency of network transmissions. If the maximum batch size (MaxKeptDecisionBatchSize) is reached before this interval elapses, the batch will be sent immediately.
 
+      - name: NumCollectLoops
+        type: int
+        valuetype: nondefault
+        default: 8
+        reload: true
```
VinozzZ: I don't think this config is hot-reloadable

collect/collect.go
ianwilkes: Since there's no recover() here, if it panics the entire process will go down. But getting stuck is also a distinct possibility, I'll give them all their own health checks.

collect/collect_test.go
ianwilkes: Fixed this pattern in a couple of places.

collect/multi_loop_test.go
ianwilkes: I forgot about that function. Hasn't been necessary but agree it's a good idea.

collect/multi_loop_test.go
ianwilkes: Good idea, doing that caught that the collector's ticker loop was still using a real clock instead of the fake one.

config/metadata/configMeta.yaml
ianwilkes: doh, obvs

PR #1673, opened 2025-08-22
perf: CollectLoop-local samplers, with shared dynsamplers

ianwilkes:
Samplers include some expensive operations (mainly key building) which are currently happening inside a mutex, defeating many of the potential gains for concurrent collect loops.

Gives each collect loop a local set of samplers, as well as a reload signal to rebuild them on config change. The dynsamplers, which can contain global state related to total throughput, are still shared, but this is now managed by the sampler factory. Each sampler has a function defining how to create the dynsampler for that type, which is used by the sampler factory. Unfortunately the penchant for lots of concrete types and struct fields here makes this code a little more verbose than it would otherwise need to be. On the plus side, the samplers themselves get a little simpler, and the rules-based sampler now defers to the sampler factory to make its sub-samplers, which I think it probably the right long-term direction.

Includes a pretty ferocious new integration test for config reloads in the app package. This revealed race conditions in the config code which are also fixed here.


sample/sample.go
```diff
@@ -43,65 +54,162 @@ func (s *SamplerFactory) updatePeerCounts() {
...
 
 	return sampler
 }
 
+// GetSamplerImplementationForKey returns the sampler implementation for the given
+// samplerKey (dataset for legacy keys, environment otherwise), or nil if it is not defined
+func (s *SamplerFactory) GetSamplerImplementationForKey(samplerKey string) Sampler {
+	c, _ := s.Config.GetSamplerConfigForDestName(samplerKey)
+
+	return s.createSampler(c, "")
```
VinozzZ: I believe we need to pass in the samplerKey here. Otherwise, two datasets or environments could be sharing the same dynsampler when they probably shouldn't.

The use case for this is if we have a rules config that looks like 
```
"RulesVersion", 2,
"Samplers/__default__/DeterministicSampler/SampleRate", 1,
"Samplers/production/TotalThroughputSampler/GoalThroughputPerSec", 10,
"Samplers/production/TotalThroughputSampler/UseClusterSize", true,
"Samplers/dogfood/TotalThroughputSampler/GoalThroughputPerSec", 10,
"Samplers/dogfood/TotalThroughputSampler/UseClusterSize", true,
```

sample/sample.go
```diff
@@ -43,65 +54,162 @@ func (s *SamplerFactory) updatePeerCounts() {
...
+}
+
+// GetDownstreamSampler creates a downstream sampler for use in rules-based sampling,
+// ensuring it participates in the shared dynsampler system
+func (s *SamplerFactory) GetDownstreamSampler(
+	parentSamplerKey string,
+	downstreamConfig *config.RulesBasedDownstreamSampler,
+) Sampler {
+	keyPrefix := fmt.Sprintf("rules:%s:", parentSamplerKey)
```
VinozzZ, line 167: I wonder if it would be more clear if we always put environment or dataset name first in the dynsampler keys. Instead of `rules:production:totoalthroughput` and `production:totalthroughput`, we can have `production:rules:totalthroughput`

sample/sample.go
ianwilkes: Oof, what a testing gap. Fixed.

sample/sample.go
ianwilkes: I see where you're coming from but doing that complicates creating these keys, for what it's ultimately an aesthetic choice. I don't think it's worthwhile.

VinozzZ (approved):

PR #1677, opened 2025-08-28
perf: use single zstd decoder

VinozzZ:

We hit an issue where all zstd decoders in the router could be exhausted, causing new incoming requests to block indefinitely until a decoder was released. On affected instances this led to memory usage spikes of 5x or more.

- Single zstd.Decoder with DecodeAll
	•	Initially, benchmarks showed this approach was ~5x slower than maintaining multiple decoders.
	•	However, after adjusting the decoder’s concurrency settings(`WithDecoderConcurrency`), performance improved significantly and matched or exceeded the pooled approach.
	•	This option is much simpler to maintain: only one decoder is needed, and DecodeAll is concurrency-safe.
	•	It also eliminates the risk of decoder exhaustion entirely and reduces memory pressure, since creating/holding multiple decoders is expensive.

## Benchmark Result
```
goos: darwin
goarch: arm64
pkg: github.com/honeycombio/refinery/route
cpu: Apple M2 Max
BenchmarkDecoderOld/orig_100B_comp_113B-12         	1352589	      910.0 ns/op	    505 B/op       6 allocs/op
BenchmarkDecoderOld/orig_1024B_comp_401B-12        	 689023	     1789 ns/op	    541 B/op       6 allocs/op
BenchmarkDecoderOld/orig_10240B_comp_2178B-12      	 313251	     3857 ns/op	    585 B/op       6 allocs/op
BenchmarkDecoderOld/orig_102400B_comp_20079B-12    	  44568	    26935 ns/op	    623 B/op       6 allocs/op
BenchmarkDecoderNew/orig_100B_comp_113B-12         	1411065	      853.3 ns/op	    504 B/op       6 allocs/op
BenchmarkDecoderNew/orig_1024B_comp_401B-12        	 707530	     1722 ns/op	    513 B/op       6 allocs/op
BenchmarkDecoderNew/orig_10240B_comp_2178B-12      	 316370	     3760 ns/op	    515 B/op       6 allocs/op
BenchmarkDecoderNew/orig_102400B_comp_20079B-12    	  45318	    26344 ns/op	    533 B/op       6 allocs/op
PASS
ok  	github.com/honeycombio/refinery/route	12.657s
```


route/route.go
```diff
@@ -860,6 +845,21 @@ func (r *Router) readAndCloseMaybeCompressedBody(req *http.Request) (*bytes.Buff
 		return nil, err
 	}
 
+	if encodingScheme == "zstd" {
```
codeboten: please add a comment here to give more insight to future us as to why this is outside of the switch now

codeboten (commented):
This is looking pretty good! just a comment about a comment :)

codeboten (approved):

PR #1679, opened 2025-08-28
maint: make CacheCapacity deprecation text more actionable

VinozzZ:

Users may mistakenly set PeerQueueSize and IncomingQueueSize to be the same as CacheCapacity. That's not what we want.

- Clarify PeerQueueSize and IncomingQueueSize should be 3x CacheCapacity in warning message


robbkidd (approved):
Yey more actionable!

PR #1681, opened 2025-09-02
fix: dataset name should only be url encoded once

VinozzZ:

Customer has reported that their `service.name` with a space in the value is showing up as `foo%20bar`

Turns out there was a [fix](https://github.com/honeycombio/libhoney-go/commit/9f4f37c3759e781e74a3c901417e80b50796a63e) added for this issue in libhoney. Our new `DirectTransmit` didn't include this fix. This PR introduce the fix into `DirectTransmit`.
TLDR is `url.Path` is expected to be unescaped. `url.String()` does the escape by default. If you set url.Path to be foo%20bar, url.String() will return foo%2520bar

### DIFF
--- a/transmit/direct_transmit.go
+++ b/transmit/direct_transmit.go
@@ -723,11 +723,7 @@ type httpError interface {
 }
 
 func buildRequestURL(apiHost, dataset string) (string, error) {
-	// validate the apiHost is a valid URL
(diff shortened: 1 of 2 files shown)


MikeGoldsmith (approved):
❤️

PR #1687, opened 2025-09-04
fix: increase upstream tranmission timeout to 30s

VinozzZ:

If Honeycomb API is taking longer than expected when receiving a batch, we want to try our best to send that data instead of bailing fast

- Allow upstreamTransmission and peerTransmission to have different timeout value for their http client
- set client timeout to be 30s for upstreamTransmission

### DIFF
--- a/app/app_test.go
+++ b/app/app_test.go
@@ -313,7 +313,7 @@ func newStartedApp(
 	if mockTransmission != nil {
 		upstreamTransmission = mockTransmission
 	} else {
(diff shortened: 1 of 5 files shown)


VinozzZ:
> I'd still feel more comfortable with a shorter timeout. In the case of HC actually being down it could add up to a ton of memory pressure.
> 
> I guess if Honeycomb is down, Refinery's gonna have a bad time anyway.

That's fair. I mostly got the value from libhoney. I decreased it to 30s. We can see how that goes

robbkidd:
Is this a reasonable candidate for being tunable through configuration? 😅

PR #1691, opened 2025-09-11
maint: remove deprecated LegacyMetrics

VinozzZ:

LegacyMetrics has been deprecated since 2.0. This is to officially stop supporting it in upcoming 3.0

- remove all LegacyMetrics related code and configs


config/metadata/configMeta.yaml
```diff
@@ -729,81 +729,6 @@ groups:
           requests for `/metrics`. Must be different from the main Refinery
           listener. Only used if `Enabled` is `true` in `PrometheusMetrics`.
 
-  - name: LegacyMetrics
```
kentquirk, line 732: The general expectation is that we'd mark all the individual fields with `lastVersion`. That would mean someone could put LegacyMetrics as a heading with no fields in it (since we don't have a way to mark a heading as deprecated), but I think that's not harmful. And that way we wouldn't lose the history of this, and we'd still have the potential to generate docs for any version number.

VinozzZ: Oh right! That's a good point. I will change it to do that. I forgot that deleting it will remove the ability to use `lastVersion`

PR #1692, opened 2025-09-12
feat: allow deprecating an entire config group

VinozzZ:

We currently can only deprecate individual field. This PR allows deprecating an entire config group

- Group-level replacements. Then entire group can be renamed and each individual field will be copied over to the new group. If an individual field has its own replacement defined, it will be honored before the group rename.
- Updated validation behavior: If the current Refinery version is after the defined deprecation version for a config field/group, it will generate an error and prevent Refinery from starting up


config/version.go
```diff
@@ -14,23 +14,31 @@ import (
 //	 1 if current > target
 //	error if versions are malformed
 func compareVersions(current, target string) (int, error) {
-	if !semver.IsValid(current) {
-		return 0, fmt.Errorf("invalid version format: %s", current)
+	currentValue := current
+	targetValue := target
+	if !semver.IsValid(currentValue) {
+		currentValue = "v" + currentValue
```
VinozzZ, line 20: I had to add this back because in our build process. We actually remove the `v` prefix for docker image tags from our version number

kentquirk (approved):
This seems to be correct. 

The below is NOT a required change. Just some thoughts:

I don't love the creation of the Deprecated type and I especially don't really like the embedding of it. It does save the duplication of the implementation of the deprecator interface, but I'm not sure that it's worth it. Personally, I find embedding in Go to be  hard to follow. You don't really know what it brings in.

As I said, I'm not requiring you change it, just noting that I think it hurts readability for minimal benefit.

PR #1694, opened 2025-09-24
fix: set default sizes for queues

robbkidd:

- Starting up Refinery 3.0-beta without IncomingQueueSize or PeerQueueSize configured results in queue sizes set to `0`.

- The sizes of these queues [used to be computed from the CacheCapacity config option that has been removed](https://github.com/honeycombio/refinery/pull/1666). With that math gone, we need to declare defaults or these will be size 0 and then you're going to have a bad time.

### DIFF
--- a/config/file_config.go
+++ b/config/file_config.go
@@ -312,8 +312,8 @@ type RedisPeerManagementConfig struct {
 }
 
 type CollectionConfig struct {
-	PeerQueueSize       int        `yaml:"PeerQueueSize"`
-	IncomingQueueSize   int        `yaml:"IncomingQueueSize"`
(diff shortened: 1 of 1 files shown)


kentquirk (approved):
I believe you also want to change the default values listed in configMeta.yaml.

I thought I only commented, but it seems to have taken it as approval. I didn't intend to approve.

VinozzZ:
> I believe you also want to change the default values listed in configMeta.yaml.
> 
> I thought I only commented, but it seems to have taken it as approval. I didn't intend to approve.

I believe the default value is already set to 30_000 in configMeta.yaml

VinozzZ (approved):

PR #1699, opened 2025-09-26
fix: only try to track usage when enabled

TylerHelmuth:

- Refinery was writing logs when trying to record usage metrics when usage tracking was not enabled.

- add a check to `healthCheck` to only record usage if the config option is enabled.


agent/agent.go
```diff
@@ -216,17 +216,20 @@ func (agent *Agent) healthCheck() {
...
 
-			traceUsage, ok := agent.metrics.Get("bytes_received_traces")
-			if !ok {
-				agent.logger.Errorf(context.Background(), "unexpected missing trace usage metric")
-			}
-			logUsage, ok := agent.metrics.Get("bytes_received_logs")
-			if !ok {
-				agent.logger.Errorf(context.Background(), "unexpected missing log usage metric")
-			}
+			recordUsage := agent.effectiveConfig.GetOpAMPConfig().RecordUsage
```
VinozzZ: You can do `agent.effectiveConfig.GetOpAMPConfig().RecordUsage.Get()`. Then you don't need to do the manual nil check and de-referencing the pointer

TylerHelmuth: sweet

MikeGoldsmith (approved):

Issue #1708, opened 2025-10-03
Images published are no longer multi-architecture


robbkidd:
**Versions**

- Go: N/A
- Refinery: 3.0.0

**Additional context**

Starting with v3.0.0, images published to registries no longer multi-architecture and are `linux/amd64` only, the arch of the build host.

<img width="2538" height="1688" alt="Image" src="https://github.com/user-attachments/assets/98e0ffbd-1dd6-4262-9be9-686b0e3dafde" />

#1683 consolidated container publishing to a single image build and multiple publishes of the same image, but in [using docker push](https://github.com/honeycombio/refinery/pull/1683/files#diff-1f844d52612ddd88e4d7d89071524917bc96ed89f50961e89b4ba2b7dba160ffR100) to do so, it dereferences the build host's CPU architecture and only pushes the image for that arch.

PR #1709, opened 2025-10-03
fix(ci): publish multi-arch image to multiple registry

VinozzZ:

- #1708 

`ko` does produce a multi-arch image, but `docker push` doesn't push a multi-arch image. 

This PR uses `crane cp` which has multi-arch support to copy multi-arch image manifests between registries.

- wire up `crane` as a docker build dependency
- reorganize `build_docker.sh` to use a `PRIMARY_DOCKER_REPO` environment variable to designate a prime destination for a particular build
  - if left unset, `ko.local` is the default and the image will only be built and "published" to the local registry
  - if set, the image build will be published to the given registry target along with the appropriate tags as determined by the build script
- `COPY_DOCKER_REPOS` replaces `KO_DOCKER_REPOS`
  - a comma-separated list of **additional** registry repos to copy the image published to `PRIMARY_DOCKER_REPO`
  - same collection of tags will be applied to the copies

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -66,7 +66,7 @@ commands:
 jobs:
   test:
     docker:
-      - image: cimg/go:1.24
+      - image: cimg/go:1.25
       - image: redis:6.2
     resource_class: xlarge
     steps:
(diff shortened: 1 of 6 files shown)


robbkidd (approved):
publish the images!

PR #1717, opened 2025-10-17
maint: consolidate Peer mocks into one mock implementation

VinozzZ:

We have three Peer mock implementation, `MockPeers`, `MockPeersWithUpdate`, and `TestPeerMockWithCallback`, in our test suit. This PR is to consolidate them into one so that we don't need to maintain duplicated code.

- Remove `TestPeerMockWithCallback` and `MockPeersWithUpdate`
- Update `MockPeers` to be concurrent safe and callbacks are triggered during Peer list update

### DIFF
--- a/sample/sample_test.go
+++ b/sample/sample_test.go
@@ -419,38 +419,6 @@ func (m *MockSampler) SaveState() ([]byte, error) {
 	return nil, nil
 }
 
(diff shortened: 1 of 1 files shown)


robbkidd (approved):
duplicate Spider-men

PR #1718, opened 2025-10-17
maint: assert RulesBasedSampler shares the same underlying dysamplers

VinozzZ:

`TestRulesBasedSamplerWithSharedDynsamplers`  is meant to test that rule based sampler can correctly uses the sampler factory to maintain shared dynsampler implementation. The assertion previously didn't do what's promised.

- check dynsampler inside RulesBasedSampler is a shared instance


robbkidd (approved):
It would be rude not to share.

PR #1719, opened 2025-10-17
feat: set default number of workers to GOMAXPROCS

robbkidd:
With Go 1.25, GOMAXPROCS is now aware of the number of logical CPUs available in containerized compute environments. We use that for our default number of collect loop workers.

https://go.dev/blog/container-aware-gomaxprocs

This change includes:

* Update config getters for queue sizes to perform the per-collect-loop worker math only ever used during Collect.Start()
* A fix for copying the pub/sub subscriptions for the workers.
* Update a looot of Collect tests configuration to have a smaller number of collect loops, proportionate to the very small test load placed upon them.

### DIFF
--- a/app/app_test.go
+++ b/app/app_test.go
@@ -423,11 +423,16 @@ func TestAppIntegration(t *testing.T) {
 
 	require.EventuallyWithT(t, func(collect *assert.CollectT) {
 		events := testServer.getEvents()
-		require.Len(collect, events, 1)
-		assert.Equal(collect, "dataset", events[0].Dataset)
-		assert.Equal(collect, "bar", events[0].Data.Get("foo"))
-		assert.Equal(collect, "1", events[0].Data.Get("trace.trace_id"))
-		assert.Equal(collect, int64(1), events[0].Data.Get(types.MetaRefineryOriginalSampleRate))
+		// TODO: fix flaky behavior where sometimes two events are received in the testServer
+		// Loop through all received events so that we can see what's the failure in CI
+		require.Greater(collect, len(events), 0)
+		for _, e := range events {
+			assert.Equal(collect, "dataset", e.Dataset)
+			assert.Equal(collect, "bar", e.Data.Get("foo"))
+			assert.Equal(collect, "1", e.Data.Get("trace.trace_id"))
+			assert.Equal(collect, int64(1), e.Data.Get(types.MetaRefineryOriginalSampleRate))
+		}
+		assert.Len(collect, events, 1)
 	}, 2*time.Second, 10*time.Millisecond)
 
 	err = startstop.Stop(graph.Objects(), nil)


config/file_config.go
```diff
@@ -348,6 +349,9 @@ func (c CollectionConfig) GetIncomingQueueSize() int {
 // GetNumCollectLoops returns the number of parallel collection loops.
 // Ensures the value is at least 1.
 func (c CollectionConfig) GetNumCollectLoops() int {
+	if c.NumCollectLoops == 0 {
+		return runtime.GOMAXPROCS(0)
```
VinozzZ: I don't know if this is a valid case. What if runtime.GOMAXPROCS return 0? Maybe it's still worth to do the max operation when we are taking the GOMAXPROCS value

robbkidd: Whew. GOMAXPROCS returning 0 would be super weird, but you're right: we always want at least one worker and this is currently the best place to ensure that.

config/file_config.go
```diff
@@ -319,7 +320,7 @@ type CollectionConfig struct {
 	// NumCollectLoops controls the number of parallel collection loops.
 	// Each loop processes a subset of traces independently.
 	// Higher values can improve throughput on multi-core systems.
-	NumCollectLoops int `yaml:"NumCollectLoops" default:"8"`
+	NumCollectLoops int `yaml:"NumCollectLoops"` // default: GOXMAXPROCS
```
robbkidd: 🤭 

```suggestion
	NumCollectLoops int `yaml:"NumCollectLoops"` // default: GOMAXPROCS
```

config/file_config.go
```diff
@@ -331,24 +332,28 @@ func (c CollectionConfig) GetMaxAlloc() MemorySize {
...
-// GetPeerBufferCapacity returns the capacity of the in-memory channel for peer traces.
-// If PeerBufferCapacity is not set, it uses 3x the cache capacity.
-// The minimum value is 3x the cache capacity.
-func (c CollectionConfig) GetPeerQueueSize() int {
-	return c.PeerQueueSize
+// GetPeerQueueSizePerLoop returns the capacity per collect loop worker of the in-memory channel for peer traces.
+// It divides queue sizes among loops, rounding up to make sure total queue size across loops is equal to or greater than configured.
+func (c CollectionConfig) GetPeerQueueSizePerLoop() int {
+	numLoops := c.GetNumCollectLoops()
+	return (c.PeerQueueSize + numLoops - 1) / numLoops
```
robbkidd, line 339: 🤔  Is there a minimum for these queue sizes we ought to ensure like we do for number of collect loop workers?

collect/collect_test.go
```diff
@@ -160,6 +160,7 @@ func TestAddRootSpan(t *testing.T) {
 		TraceIdFieldNames:  []string{"trace.trace_id", "traceId"},
 		ParentIdFieldNames: []string{"trace.parent_id", "parentId"},
 		GetCollectionConfigVal: config.CollectionConfig{
+			NumCollectLoops:   2,
```
robbkidd, line 163: My palms itch to create a reusable `config.CollectionConfig{}` to declare configuration common to all these tests in one place and override as needed. But I suppose that might be too much duplication removal.

robbkidd: I will resist.

config/file_config.go
robbkidd: Added [a test case](https://github.com/honeycombio/refinery/pull/1719/commits/ef2ac3baf91f425be8e2a653e5ca3a47d612f5e5) to assert the current minimum queue size of 1-per-loop. I wouldn't want to go to production with that queue size, but at least the current behavior is tested so changing that minimum should be straightforward if we choose to.

PR #1720, opened 2025-10-20
fix: round up queue size per loop

VinozzZ:

In Go, integer division truncates towards zero. That means when NumCollectLoop is 3 and IncomingQueueSize is 7, per loop queue size would end up with 2. The total queue size after the division becomes less than the configured queue size.

- Round up the division result so that we can guarantee the total queue size is greater or equal to the configured queue size

### DIFF
--- a/pubsub/pubsub_local.go
+++ b/pubsub/pubsub_local.go
@@ -78,8 +78,8 @@ func (ps *LocalPubSub) Publish(ctx context.Context, topic, message string) error
 	ps.Metrics.Count("local_pubsub_published", 1)
 	ps.Metrics.Count("local_pubsub_received", int64(len(ps.topics[topic])))
 	// make a copy of our subs so we don't hold the lock while calling them
-	subs := make([]*LocalSubscription, 0, len(ps.topics[topic]))
-	subs = append(subs, ps.topics[topic]...)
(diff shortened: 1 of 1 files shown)


collect/collect.go
```diff
@@ -212,8 +212,9 @@ func (i *InMemCollector) Start() error {
 	i.collectLoops = make([]*CollectLoop, numLoops)
 
 	// Divide queue sizes among loops
-	incomingPerLoop := imcConfig.GetIncomingQueueSize() / numLoops
-	peerPerLoop := imcConfig.GetPeerQueueSize() / numLoops
+	// Round up the queue size per loop to make sure we have at least configured total queue size
+	incomingPerLoop := (imcConfig.GetIncomingQueueSize() + numLoops - 1) / numLoops
```
MikeGoldsmith, line 216: Could we use math.Ceil instead of adding config and numLoops then subtract 1?

MikeGoldsmith (approved):
Looks okay.

Left a non-blocking query whether we can use a math func instead of the inline calculations.

collect/collect.go
VinozzZ: math.Ceil requires the value to be an float instead of int type

PR #1722, opened 2025-10-22
maint: rename CollectLoop related code to improve readability

VinozzZ:

The current naming for the collector (manager) and collect_loop(worker) doesn't reflect the architecture design. This makes naming metrics and understanding code logic hard. We would like to rename the newly added code to reflect the new architecture so that it's easier to understand and work with

Type & Struct Renames:

- CollectLoop → CollectorWorker
- NewCollectLoop() → NewCollectorWorker()

Field & Variable Renames in InMemCollector:

- collectLoops []*CollectLoop → workers 
  []*CollectorWorker
- collectLoopsWG → workersWG
- houseKeepingWG → monitorWG
- outgoingTraces → tracesToSend

Method Renames:

- houseKeeping() → monitor()
- getLoopForTrace() → getWorkerIDForTrace()

Constant Renames:

- collectorLoopHealthPrefix → collectWorkerHealthPrefix

File Renames:
- collect/collect_loop.go → collect/collector_worker.go

### DIFF
--- a/metrics/otel_metrics.go
+++ b/metrics/otel_metrics.go
@@ -27,6 +27,9 @@ var _ MetricsBackend = (*OTelMetrics)(nil)
 // aggregates (e.g. avg, p50, p95, etc.). OTel, on the other hand, sends the raw
 // histogram values and lets Honeycomb do the aggregation on ingest. The columns
 // in the resulting datasets will not be the same.
+//
+// To minimize lock contention, this implementation uses sync.Map to reduce lock
+// contention for concurrent access to metric instruments.
 type OTelMetrics struct {
 	Config  config.Config `inject:""`
 	Logger  logger.Logger `inject:""`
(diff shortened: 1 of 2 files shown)


collect/collect.go
```diff
@@ -209,22 +209,22 @@ func (i *InMemCollector) Start() error {
 		}
 	}
 
-	i.collectLoops = make([]*CollectLoop, numLoops)
+	i.workers = make([]*CollectorWorker, numLoops)
```
kentquirk: There are still a lot of "loops" around. Do you want to change these to numWorkers and workerID?

collect/collect.go
```diff
@@ -333,21 +336,21 @@ func (i *InMemCollector) houseKeeping() {
...
+				totalReceived += worker.localSpanReceived.Swap(0)
 			}
 
 			i.Metrics.Histogram("collector_incoming_queue", float64(totalIncoming))
 			i.Metrics.Histogram("collector_peer_queue", float64(totalPeer))
 			i.Metrics.Gauge("collector_incoming_queue_length", float64(totalIncoming))
 			i.Metrics.Gauge("collector_peer_queue_length", float64(totalPeer))
 			i.Metrics.Gauge("collector_cache_size", float64(totalCacheSize))
-			i.Metrics.Gauge("collector_num_loops", float64(len(i.collectLoops)))
+			i.Metrics.Gauge("collector_num_loops", float64(len(i.workers)))
```
kentquirk: similarly here -- it's harder to change a metric, but should we?

collect/collector_worker.go
```diff
@@ -18,21 +18,23 @@ import (
 	"github.com/honeycombio/refinery/types"
 )
 
-const collectorLoopHealthPrefix = "collector-loop-"
+const collectLoopHealthPrefix = "collect-loop-"
```
kentquirk: should this be worker as well?

collect/collect.go
VinozzZ: This is a new metric that hasn't been released yet. It's probably safe to rename! Good Catch!

collect/collector_worker.go
robbkidd: _See how confused we got?_

PR #1724, opened 2025-10-23
perf: replace runtime.ReadMemStats with runtime/metrics

VinozzZ:

[runtime.ReadMemStats](https://pkg.go.dev/runtime#ReadMemStats) requires a `stopTheWorld` call when retrieving memory usage stats. Since Go 1.16, the new [runtime/metrics](https://pkg.go.dev/runtime/metrics#pkg-overview) doesn't require a `stopTheWorld` call.

The reason why I would like to avoid the `stopTheWorld` call here is because we are calling `ReadMemStats` every 100ms now. In Refinery 3.0, the `ReadMemStats` is only called when peer queue is empty. That's way less frequent and less disruptive. I'm hoping this PR will allow us to maintain accurate memory usage reporting while having less interruption to the main work of Refinery. 

- use `runtime/metrics` to read `"/memory/classes/heap/objects:bytes"` stats in `checkAlloc`

If this new package proves to work well, we can also use it report an estimated CPU usage. I will leave that work in a future PR

### DIFF
--- a/collect/collect.go
+++ b/collect/collect.go
@@ -6,7 +6,7 @@ import (
 	"fmt"
 	"os"
 	"runtime"
-	runtimemetrics "runtime/metrics"
+	rtmetrics "runtime/metrics"
 	"sync"
 	"time"
 
@@ -38,6 +38,7 @@ const (
 	defaultKeptDecisionTickerInterval = 1 * time.Second
 
 	collectorHealthKey = "collector"
+	memMetricName      = "/memory/classes/heap/objects:bytes"
 )
 
 var ErrWouldBlock = errors.New("Dropping span as channel buffer is full. Span will not be processed and will be lost.")
@@ -115,7 +116,7 @@ type InMemCollector struct {
 
 	hostname string
 
-	memMetricSample []runtimemetrics.Sample // Memory monitoring using runtime/metrics
+	memMetricSample []rtmetrics.Sample // Memory monitoring using runtime/metrics
 }
 
 // These are the names of the metrics we use to track the number of events sent to peers through the router.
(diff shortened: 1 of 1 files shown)


collect/collect.go
```diff
@@ -209,6 +212,11 @@ func (i *InMemCollector) Start() error {
 		}
 	}
 
+	// Initialize runtime/metrics sample for efficient memory monitoring
+	i.memMetricSample = []runtimemetrics.Sample{
+		runtimemetrics.Sample{Name: "/memory/classes/heap/objects:bytes"},
```
kentquirk: I know that in the past we've only had the heap available to us, but given that we sort of tie our behavior to total available memory, should we maybe convert to `/memory/classes/total:bytes`?

kentquirk: Also, I had to read this several times. Would this be clearer?

```go
	const memMetric = "/memory/classes/heap/objects:bytes"
	i.memMetricSample := make([]metrics.Sample, 1)
	i.memMetricSample[0].Name = memMetric
```

collect/collect.go
```diff
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"runtime"
+	runtimemetrics "runtime/metrics"
```
kentquirk: `runttimemetrics` is hard to read. Maybe `rtmetrics`?

collect/collect.go
VinozzZ: Will changing to total memory usage affect the sensitivity of stress level calculation using this metric?

kentquirk: Yes, but I'm not sure by how much. The main thing is that the sum total of memory usage could get closer to the actual number, so instead of setting limits at 75% it might become 80% or more. It's probably too risky, now that I think it through.

collect/collect.go
robbkidd:
```
	runtiMemeTrics "runtime/metrics"
```

kentquirk: That's what I was reading, yes.

robbkidd: 🤔 A reasonable alias for system performance numbers at a coarse collection interval.

PR #1727, opened 2025-10-27
perf: reduce lock contention for health check system

VinozzZ:

Refinery’s health check system relies on a global state protected by a mutex. 

### Before:

Each worker registers and reports its readiness by writing to the global Health state. When Refinery runs with many workers, these concurrent updates lead to high lock contention, negatively impacting performance.

### After:

The Collector registers with the Health system. The Collector's readiness check is whether all its workers are healthy. Workers keep track of their own health check-in times to reduce updates to global Health state.

This change introduces a collector-level health aggregation model:
- Each worker reports its health only to its collector (local state).
- The collector is solely responsible for reporting health to the global health system.
- If all workers are healthy, the collector reports healthy.
- If any worker is unhealthy, the collector reports unhealthy.

### DIFF
--- a/collect/collect.go
+++ b/collect/collect.go
@@ -332,20 +332,8 @@ func (i *InMemCollector) isReady() bool {
 	timeout := i.Config.GetHealthCheckTimeout()
 
 	for _, worker := range i.workers {
-		lastUpdate := worker.lastHealthUpdate.Load()
-		if lastUpdate == 0 {
-			// Worker hasn't initialized yet, consider it not healthy
-			ready = false
-		}
-
-		lastUpdateTime := time.Unix(0, lastUpdate)
-		if now.Sub(lastUpdateTime) > timeout {
-			// Worker hasn't reported in within timeout period
-			i.Logger.Warn().
-				WithField("worker_id", worker.ID).
-				WithField("last_update", lastUpdateTime).
-				WithField("timeout", timeout).
-				Logf("Worker is unhealthy - hasn't reported within timeout")
+		healthy := worker.IsHealthy(now, timeout)
+		if !healthy {
 			ready = false
 		}
 	}
--- a/collect/collect_test.go
+++ b/collect/collect_test.go
@@ -1839,7 +1839,7 @@ func TestWorkerHealthReporting(t *testing.T) {
 	}, 2*time.Second, 50*time.Millisecond, "InMemCollector should initially be healthy")
 
 	for i, worker := range coll.workers {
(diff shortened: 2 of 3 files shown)


collect/collect.go
```diff
@@ -337,7 +367,9 @@ func (i *InMemCollector) monitor() {
 	for {
 		select {
 		case <-ticker.Chan():
-			i.Health.Ready(collectorHealthKey, true)
+			// Check worker health and report aggregated status
+			allWorkersHealthy := i.checkWorkersHealth()
+			i.Health.Ready(collectorHealthKey, allWorkersHealthy)
```
robbkidd:
```suggestion
			i.Health.Ready(collectorHealthKey, i.checkWorkersHealth())
```

Can we remove the local variable and call the worker health check directly?

collect/collect.go
```diff
@@ -323,9 +323,39 @@ func (i *InMemCollector) checkAlloc(ctx context.Context) {
...
+// Returns true if all workers are healthy, false if any worker is unhealthy.
+func (i *InMemCollector) checkWorkersHealth() bool {
+	now := i.Clock.Now()
+	timeout := i.Config.GetHealthCheckTimeout()
+
+	for _, worker := range i.workers {
+		lastUpdate := worker.lastHealthUpdate.Load()
+		if lastUpdate == 0 {
+			// Worker hasn't initialized yet, consider it not healthy
+			return false
```
MikeGoldsmith: Should we wait a set amount of time for workers to report in before determining them as unhealthy? What are the repercussions if the process is marked unhealthly?

robbkidd: > Should we wait a set amount of time for workers to report in before determining them as unhealthy?

Our current thinking is to rely on the Health system's three-strikes rule. That system allows for three rounds of readiness checking to fail before concluding the system is unhealthy.

> What are the repercussions if the process is marked unhealthly?

The same repercussion as a single-threaded Collector not clocking in for a health check within three check timeouts, but instead based on any three workers failing to clock in within three check timeouts. Three consecutive failures of this function will mark the Collector unhealthy which will mark the Refinery host unhealthy.

We're keeping the implementation for detecting and reacting to worker health relatively naive at the moment because we haven't observed their health under load in this way yet.

The Before behavior for this PR is that each worker would register with the Health system. Each worker's health/readiness would be tracked individually and each would have three chances to get healthy again. After this PR, the Collector registers with the health system and any worker being unhealthy for three consecutive readiness checks would make the Collector unhealthy.

PR #1735, opened 2025-11-05
fix: use sendKey for rule matching if it's configured

### LINKED ISSUE #1732: SendKeyMode doesn't replace api key during sampling
**Versions**

- Refinery: 3.0.0

**Steps to reproduce**

1. Send libhoney events to the /batch or /event endpoint with an API key that belongs to environment A
2. Configure Refinery with `SendkeyMode: all` and a `SendKey` that belongs to environment B
3. Set up Refinery rules targeted for environment B and a `__default__` rule
4. Observing the events sent with apikey belongs to env A get sampled by `__default__` rule instead of the targeted one for env B

**Additional context**

VinozzZ:

* Fixes #1732 

Refinery uses the API key included in an incoming request to determine the target environment and dataset. However, when a SendKey is configured, it can override the original incoming API key in the request.

When SendKey is configured to override, Refinery should resolve the environment and dataset based on the SendKey, not the original incoming API key. Otherwise, sampling rules may be incorrectly matched using the wrong key.

This PR ensures that when a SendKey is applied, the environment/dataset resolution and sampling rule matching are based on the SendKey.

- Replaces the request’s APIKey header value with the configured SendKey, when applicable
- Adds integration tests to validate SendKey-based resolution behavior

### DIFF
--- a/app/app_test.go
+++ b/app/app_test.go
@@ -460,7 +460,7 @@ func TestAppIntegrationSendKey(t *testing.T) {
 			incomingAPIKey:      "",
 			expectedUpstreamKey: nonLegacyAPIKey,
 			shouldSucceed:       true,
-			expectedReason:      "rules/trace/test",
+			expectedReason:      "rules/trace/match.environment",
 			description:         "Blank key should be replaced with SendKey",
 		},
 		{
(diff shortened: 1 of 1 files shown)


robbkidd (approved):
key keeper

PR #1737, opened 2025-11-06
perf: reduce the amount of time the write lock is held during drain()

VinozzZ:

Reading from a channel is concurrency-safe in Go, so there’s no need to hold a mutex while pulling work from it. Currently, holding the lock during this phase causes unnecessary contention, limiting throughput and slowing down channel draining.

This change reduces lock contention and gives more room for concurrent readers to operate efficiently.

- Build the batch of trace decisions that need to be inserted into the cuckoo cache concurrently. 
- Only grab the mutex when actually doing the insertion.

### DIFF
--- a/collect/cache/cuckoo.go
+++ b/collect/cache/cuckoo.go
@@ -121,6 +121,7 @@ func (c *CuckooTraceChecker) drain() {
 		return
 	}
 	batch := batchPool.Get().([]string)
+	timeout := time.NewTimer(500 * time.Microsecond)
 queueLoop:
 	for i := 0; i < n; i++ {
 		select {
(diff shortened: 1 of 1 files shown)


collect/cache/cuckoo.go
```diff
@@ -114,33 +120,48 @@ func (c *CuckooTraceChecker) drain() {
...
+	// TODO: we could have a goroutine to do this work
+	c.mut.Lock()
+	// we don't start the timer until we have the lock, because we don't want to be counting
+	// the time we're waiting for the lock.
+	lockStart := time.Now()
+	timeout := time.NewTimer(1 * time.Millisecond)
+
+insertLoop:
+	for _, b := range batch {
+		select {
```
VinozzZ: We can't just bail out here because we have already grabbed the item off of the queue

collect/cache/cuckoo.go
```diff
@@ -114,34 +120,39 @@ func (c *CuckooTraceChecker) drain() {
...
-			s := []byte(t)
-			c.current.Insert(s)
-			// don't add anything to future if it doesn't exist yet
-			if c.future != nil {
-				c.future.Insert(s)
-			}
-		case <-timeout.C:
-			break outer
+			// building up the queue
+			batch = append(batch, t)
```
VinozzZ, line 134: we should also bail out if the `len(batch)` has reached 1000

PR #1743, opened 2025-11-10
perf: remove unecessary locks

VinozzZ:

This PR removes three mutex from the code base that no longer is needed. 

- remove `keptMux` in CuckooSentCache. The `keptMux` is used to make the `kept` cache concurrent safe. Since the underlying LRU cache is thread safe already, we don't need to have another mutex around cache access.
- remove `keyMut` and lock in `RecordMetrics since refinery samplers are not shared among CollectorWorkers
- Use `atomic.Uint32` for `KeptTraceCacheEntry` for concurrent safe access

### DIFF
--- a/collect/cache/cuckooSentCache.go
+++ b/collect/cache/cuckooSentCache.go
@@ -2,6 +2,7 @@ package cache
 
 import (
 	"sync"
+	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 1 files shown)


kentquirk (commented):
Why are you saying that GetSampleRate is concurrency-safe? It's called from the logger, and from stress relief, as well as the collect loop. I don't understand why this is now unneeded.

PR #1745, opened 2025-11-17
perf: separate trace decision cache per worker

VinozzZ:

The trace decision cache is where sampling decisions are remembered. There is a buffered channel acting as a queue between a collector worker and the decision cache.

We have implemented multiple collector workers who all enqueue decisions. Currently there is only a single trace decision cache and it has a single goroutine responsible for draining the channel to insert those decisions into the cache.

Under high load, especially when many workers attempt to record decisions at once, the channel fills faster than the single consumer can process it. Once the queue is full, new sampling decisions are dropped. This results in incomplete or missing entries in the decision cache, which can lead to inconsistent sampling behavior. 

This PR replaces the single shared trace decision cache with individual trace decision caches for each collector worker. Instead of sending all sampling decisions through one shared channel—creating a bottleneck where a single goroutine must drain and update a global cache—each worker now writes directly to its own local cache.

### DIFF
--- a/collect/collect_test.go
+++ b/collect/collect_test.go
@@ -179,7 +179,7 @@ func TestAddRootSpan(t *testing.T) {
 
 	span := &types.Span{
 		TraceID: traceID1,
-		Event: types.Event{
+		Event: &types.Event{
 			Dataset: "aoeu",
 			APIKey:  legacyAPIKey,
 		},
@@ -201,7 +201,7 @@ func TestAddRootSpan(t *testing.T) {
 
 	span = &types.Span{
 		TraceID: traceID2,
-		Event: types.Event{
+		Event: &types.Event{
 			Dataset: "aoeu",
 			APIKey:  legacyAPIKey,
 		},
@@ -260,7 +260,7 @@ func TestOriginalSampleRateIsNotedInMetaField(t *testing.T) {
 		sendAttemptCount++
 		span := &types.Span{
 			TraceID: fmt.Sprintf("trace-%v", sendAttemptCount),
-			Event: types.Event{
+			Event: &types.Event{
 				Dataset:    "aoeu",
 				APIKey:     legacyAPIKey,
 				SampleRate: originalSampleRate,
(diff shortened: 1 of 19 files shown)


config/file_config.go
```diff
@@ -386,6 +386,16 @@ type SampleCacheConfig struct {
 	KeptSize          uint     `yaml:"KeptSize" default:"10_000"`
 	DroppedSize       uint     `yaml:"DroppedSize" default:"1_000_000"`
 	SizeCheckInterval Duration `yaml:"SizeCheckInterval" default:"10s"`
+
+	WorkerCount uint `yaml:"-"`
```
VinozzZ, line 392: add a code comment about this field is only exported for test purpose.

collect/collector_worker.go
```diff
@@ -52,6 +52,9 @@ type CollectorWorker struct {
 	// there is no way to safely access this directly from the outside.
 	cache cache.Cache
 
+	// Our local sample trace cache for tracking sent traces
+	sampleTraceCache cache.TraceSentCache
```
VinozzZ: `sampleTraceCache` -> `sampleCache`

collect/collector_worker.go
```diff
@@ -52,6 +52,9 @@ type CollectorWorker struct {
 	// there is no way to safely access this directly from the outside.
 	cache cache.Cache
 
+	// Our local sample trace cache for tracking sent traces
```
VinozzZ: A cache for storing sampling decisions, kept or dropped

collect/collector_worker.go
```diff
@@ -52,6 +52,9 @@ type CollectorWorker struct {
 	// there is no way to safely access this directly from the outside.
```
VinozzZ, line 52: Our local cache of traces awaiting sampling decision

config/file_config_test.go
```diff
@@ -293,3 +293,34 @@ func TestGetSamplingKeyFieldsForDestName(t *testing.T) {
 		})
 	}
 }
+
+func Test_GetKeptSizePerWorker(t *testing.T) {
```
VinozzZ: Test_GetSampleCacheSizePerWorker

PR #1746, opened 2025-11-18
fix: generate sampling key fields for OTLP

VinozzZ:

Passing arguments to NewCoreFieldsUnmarshaler is error-prone when the function takes a long list of positional string arguments. It’s easy to misorder or mismatch parameters, and mistakes can be difficult to detect during code reviews. Linters also provide limited help when everything is just string.

This PR replaces the positional argument list with a struct-based configuration. By using a struct, each field is explicitly named, making the call site clearer and less error-prone

### DIFF
--- a/route/batched_event_test.go
+++ b/route/batched_event_test.go
@@ -49,14 +49,24 @@ func TestBatchedEventRoundTrip(t *testing.T) {
 	}
 
 	// Create batchedEvents struct and set events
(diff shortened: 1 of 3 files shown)


robbkidd (approved):
I love it.

robbkidd:
I don't love the failure to publish, tho.

PR #1748, opened 2025-11-18
perf: do not extract sampling fields for OTLP endpoint

VinozzZ:

In large Refinery deployments, the majority of incoming spans are forwarded to peer nodes for their final sampling decision. This means the ingesting node does not need to extract or memoize the sampling key fields during initial unmarshaling. However, the current implementation still performs this work for every span, even when the local node will not use the sampling key. This results in unnecessary allocations and higher memory usage. By skipping sampling key memoization on ingest, we can significantly reduce allocation overhead and improve memory efficiency.

This optimization currently applies only to OTLP data, because OTLP is accepted exclusively from external sources and not from peers. While we ultimately want to extend this optimization to all incoming data, achieving that will require substantial additional work. We plan to address this more comprehensive implementation in the future.

- create a `UnmarshalMsgpEventMetadataOnly` method
- use the new method in OTLP path

### DIFF
--- a/route/route.go
+++ b/route/route.go
@@ -672,7 +672,15 @@ func (router *Router) processOTLPRequestBatchMsgp(
 		totalEvents += len(batch.Events)
 		for _, ev := range batch.Events {
 			payload := types.NewPayload(router.Config, nil)
-			err := coreFieldsUnmarshaler.UnmarshalPayloadCompleteMetadataOnly(ev.Attributes, &payload)
+			// In large Refinery deployments, most incoming spans are forwarded to peers for sampling,
+			// so the ingesting node doesn't need to extract or memoize sampling key fields. Skipping
+			// this work reduces allocations and improves memory efficiency.
+			//
+			// This optimization currently applies only to OTLP data because OTLP is accepted only from
+			// external sources, not from peers.
+			//
+			// TODO: Apply this optimization to all incoming data instead of relying on OTLP-specific behavior.
+			err := coreFieldsUnmarshaler.UnmarshalMsgpEventMetadataOnly(ev.Attributes, &payload)
 			if err != nil {
 				router.Logger.Error().Logf("Error unmarshaling payload: " + err.Error())
 				continue
--- a/types/payload.go
+++ b/types/payload.go
@@ -500,26 +500,26 @@ func NewCoreFieldsUnmarshaler(opt CoreFieldsUnmarshalerOptions) CoreFieldsUnmars
 	}
 }
 
-// UnmarshalPayloadCompleteMetadataOnly is similar to UnmarshalPayload, but it does not return
+// UnmarshalMsgpEventMetadataOnly is similar to UnmarshalPayload, but it does not return
 // the remaining bytes. It's expected to be used on a single message
 // where the entire byte slice is consumed.
 // It memoizes metadata fields only. If you need sampling key fields to be memoized, please use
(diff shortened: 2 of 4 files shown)


robbkidd (commented):
> … could we switch between the Complete and MetadatOnly versions based on whether the current node will be the decision maker? 

We use a Sharder in the router to determine the decision maker for an event with the event's trace ID. 

We don't know the trace ID until we unmarshal the payload.

Mmmmaybe the CoreFieldsUnmarshaller could be given a reference to the router's Sharder so that the determination could be made during extractCriticalFieldsFromBytes() as it loops over msgp fields. Once it finds the trace ID, check shard ownership and skip the sampling-key field memoization if the current host is not the owning shard.

types/payload.go
```diff
@@ -500,13 +500,28 @@ func NewCoreFieldsUnmarshaler(opt CoreFieldsUnmarshalerOptions) CoreFieldsUnmars
 	}
 }
 
+// UnmarshalPayloadCompleteMetadatOnly is similar to UnmarshalPayload, but it does not return
+// the remaining bytes. It's expected to be used on a single message
+// where the entire byte slice is consumed.
+// It memoizes metadata fields only. If you need sampling key fields to be memoized, please use
+// UnmarshalPayloadComplete.
+// CAUTION: This should only be used when the entire byte slice is safe for the payload to keep.
+func (cu CoreFieldsUnmarshaler) UnmarshalPayloadCompleteMetadatOnly(bts []byte, payload *Payload) error {
```
robbkidd:
```diff
- UnmarshalPayloadCompleteMetadatOnly
+ UnmarshalPayloadCompleteMetadataOnly
```

route/route.go
```diff
@@ -672,7 +672,7 @@ func (router *Router) processOTLPRequestBatchMsgp(
 		totalEvents += len(batch.Events)
 		for _, ev := range batch.Events {
 			payload := types.NewPayload(router.Config, nil)
-			err := coreFieldsUnmarshaler.UnmarshalPayloadComplete(ev.Attributes, &payload)
+			err := coreFieldsUnmarshaler.UnmarshalPayloadCompleteMetadataOnly(ev.Attributes, &payload)
```
VinozzZ: add comments about why we are only doing this for OTLP endpoint

types/payload.go
```diff
@@ -500,13 +500,28 @@ func NewCoreFieldsUnmarshaler(opt CoreFieldsUnmarshalerOptions) CoreFieldsUnmars
...
 // UnmarshalPayloadComplete is similar to UnmarshalPayload, but it does not return
 // the remaining bytes. It's expected to be used on a single message
 // where the entire byte slice is consumed.
 // CAUTION: This should only be used when the entire byte slice is safe for the payload to keep.
 func (cu CoreFieldsUnmarshaler) UnmarshalPayloadComplete(bts []byte, payload *Payload) error {
+	return cu.unmarshalPayloadComplete(bts, payload, cu.samplingKeyFields)
+}
+
+// unmarshalPayloadComplete is the common implementation for unmarshaling a complete payload.
+func (cu CoreFieldsUnmarshaler) unmarshalPayloadComplete(bts []byte, payload *Payload, samplingKeyFields []string) error {
```
VinozzZ: unmarshalPayloadComplete -> UnmarshalMsgpEvent
UnmarshalPayload -> UnmarshalMsgpFirstEvent

PR #1755, opened 2025-11-25
ci: update dependencies license compliance check process

robbkidd:

- Dependency license verification for some reason was not notifying us of changes as soon as we'd expect. Also, not in a very informative manner.

- **Upgrade go-licenses to latest release.** Note: it's currently unmaintained, but latest release is better than even older alpha.
  - Install go-licenses with `go install` according to the current Idioms™ and rely on the version declared in go.mod for which one to install. `go mod tidy` ought to have retrieved it and its dependencies already, so `go install` puts the correct executable for the current OS/arch on PATH. 
- **Use git for diff check.** Instead of having go-licenses download to a temporary directory, run our "update-licenses" make target to see if it makes changes to the git-tracked LICENSES directory. Use git status to determine if there are changes and, if so, exit failure.
- **Only check Linux build deps.** Scope the execution of go-licenses to GOOS=linux, our only supported targeted OS for distribution.

I throw myself on the mercy of the court for the Bash and Make crimes committed here in the service of license compliance.


codeboten (approved):

codeboten:
```
#: We ignore the standard library (go list std) as a workaround for \
"https://github.com/google/go-licenses/issues/244." The awk script converts the output \
  of `go list std` (line separated modules) to the input that `--ignore` expects (comma separated modules).
/bin/sh: 4: Syntax error: "(" unexpected
```

robbkidd:
Yeaaa. I'm working on the failure and I've fallen down a yak hole.

robbkidd:
I throw myself on the mercy of the court for the Bash and Make crimes committed here in the service of license compliance.

Issue #1704, opened 2025-10-01
Compute available memory automatically when "AvailableMemory" parameter is not set


seh:
_Refinery_ accepts two different means to govern how much memory it can use before it takes steps to conserve its use and allow reclaiming garbage that it has abandoned, but both methods require specifying an absolute quantity of bytes: `AvailableMemory` for the total from which to claim a proportioned amount, or `MaxAlloc` for the absolute amount it may use. In both cases, those values fall out of step easily with parallel configurations for a container's available memory, such as within a Kubernetes pod's container spec, in which we specify a request for or limit on the amount of memory.

It would help to eliminate the need for specifying the `AvailableMemory` parameter and allow sole use of the `MaxMemoryPercentage` parameter with a positive value, emulating [the OpenTelemetry Collector's _memory\_limiter_ processor](https://github.com/open-telemetry/opentelemetry-collector/tree/main/processor/memorylimiterprocessor). That processor [uses its `iruntime.TotalMemory` function](https://github.com/open-telemetry/opentelemetry-collector/tree/main/processor/memorylimiterprocessor) to deduce the amount of memory available to the process, using a variety of techniques such as [reading the container's _cgroup_ configuration](https://github.com/open-telemetry/opentelemetry-collector/blob/14a78328c6abbbced62ebd1e273dd200e46cc7fb/internal/memorylimiter/iruntime/total_memory_linux.go#L14-L54).

This request comes by way of [discussion in the "discuss-refinery" channel](https://honeycombpollinators.slack.com/archives/C01BZ3J7RMJ/p1759345662394549?thread_ts=1757608215.812999&cid=C01BZ3J7RMJ) of the _Honeycomb Pollinators_ Slack workspace.

PR #1756, opened 2025-11-30
Auto-detect available memory when AvailableMemory is unset and apply MaxMemoryPercentage

kushalShukla-web:
### Which problem is this PR solving?

When AvailableMemory and MaxAlloc are not set, Refinery has no way to determine its memory limit. This makes configs drift from actual container memory. This PR implements the behavior requested in #1704.

### Short description of the changes

- Auto-detect container memory (via cgroup) when both values are unset
- Apply MaxMemoryPercentage to compute an effective MaxAlloc
- Do not require AvailableMemory anymore; keep backwards compatibility
- Add basic tests for the new fallback logic

### DIFF
--- a/internal/system/memory.go
+++ b/internal/system/memory.go
@@ -2,6 +2,7 @@ package system
 
 import (
 	"bufio"
+	"bytes"
 	"errors"
 	"math"
 	"os"
@@ -13,18 +14,16 @@ import (
 // It tries to read from cgroup limits first, then falls back to system memory.
 func GetTotalMemory() (uint64, error) {
 	// Try cgroup v2
-	mem, err := getCgroupV2Memory()
-	if err == nil {
+	if mem, err := getCgroupV2Memory(); err == nil {
 		return mem, nil
 	}
 
 	// Try cgroup v1
-	mem, err = getCgroupV1Memory()
-	if err == nil {
+	if mem, err := getCgroupV1Memory(); err == nil {
 		return mem, nil
 	}
 
-	// Fallback to system memory
+	// Fall back to system memory
 	return getSystemMemory()
 }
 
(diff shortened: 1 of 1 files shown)


internal/system/memory.go
```diff
@@ -0,0 +1,83 @@
...
+		return mem, nil
+	}
+
+	// Try cgroup v1
+	mem, err = getCgroupV1Memory()
+	if err == nil {
+		return mem, nil
+	}
+
+	// Fallback to system memory
```
seh:
```suggestion
	// Fall back to system memory
```

internal/system/memory.go
```diff
@@ -0,0 +1,83 @@
...
+	"strconv"
+	"strings"
+)
+
+// GetTotalMemory returns the total available memory in bytes.
+// It tries to read from cgroup limits first, then falls back to system memory.
+func GetTotalMemory() (uint64, error) {
+	// Try cgroup v2
+	mem, err := getCgroupV2Memory()
+	if err == nil {
```
seh:
```suggestion
	if mem, err := getCgroupV2Memory(); err == nil {
```

internal/system/memory.go
```diff
@@ -0,0 +1,83 @@
...
+		return mem, nil
+	}
+
+	// Try cgroup v1
+	mem, err = getCgroupV1Memory()
+	if err == nil {
```
seh:
```suggestion
	if mem, err := getCgroupV1Memory(); err == nil {
```

internal/system/memory.go
```diff
@@ -0,0 +1,83 @@
...
+	// Fallback to system memory
+	return getSystemMemory()
+}
+
+func getCgroupV2Memory() (uint64, error) {
+	data, err := os.ReadFile("/sys/fs/cgroup/memory.max")
+	if err != nil {
+		return 0, err
+	}
+	s := strings.TrimSpace(string(data))
```
seh: Consider using [the `bytes.TrimSpace` function](https://pkg.go.dev/bytes#TrimSpace) instead to defer converting the byte vector to a string until you're sure that you need to do so.

kushalShukla-web: Hi @seh i have made an update !

seh: Thank you. I'm not an official maintainer here, so I must defer to the maintainers for the remainder of the review. I was reading this one out of interest since I had written the motivating issue (#1704). My comments thus far are my initial reactions from reading the patch for the first time.

PR #1799, opened 2026-03-24
feat: add capacity/limit companion metrics for queues and memory

mterhar:
Since we don't necessarily know the configuration that is running in a refinery node, this information will allow us to sense when a measurement is getting close to the limit for peer queue, incoming queue, and memory allocation.

### DIFF
--- a/collect/collect.go
+++ b/collect/collect.go
@@ -348,7 +348,8 @@ func (i *InMemCollector) monitor() {
 			monitorConfig := i.Config.GetCollectionConfig()
 			i.Metrics.Gauge("collector_incoming_queue_capacity", float64(monitorConfig.IncomingQueueSize))
 			i.Metrics.Gauge("collector_peer_queue_capacity", float64(monitorConfig.PeerQueueSize))
-			i.Metrics.Gauge("memory_limit", float64(monitorConfig.GetMaxAlloc()))
+			maxAlloc := monitorConfig.GetMaxAlloc()
+			i.Metrics.Gauge("memory_limit", float64(maxAlloc))
 
 			// Aggregate metrics
 			totalIncoming := 0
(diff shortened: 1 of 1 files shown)


robbkidd (approved):
I agree.

We've talked in the past about the oddities of sending the same number from config over-and-over again, but I agree that this is useful in a world where the people looking at the telemetry don't necessarily have access to the config to look it up.

(I submit: _also_ useful to people who have access to the config because they don't have to leave looking at the data to look an expectation up.)

collect/collect.go
```diff
@@ -258,6 +261,7 @@ func (i *InMemCollector) checkAlloc(ctx context.Context) {
 	inMemConfig := i.Config.GetCollectionConfig()
 	maxAlloc := inMemConfig.GetMaxAlloc()
 	i.Metrics.Store(DENOMINATOR_MEMORY_MAX_ALLOC, float64(maxAlloc))
+	i.Metrics.Gauge("memory_limit", float64(maxAlloc))
```
VinozzZ: Since this is not a hot reloadable config, I don't think it should be reported every time the `checkAlloc` function runs. If we want these in metrics, I think it should be reported once during startup

collect/collect.go
```diff
@@ -341,6 +345,11 @@ func (i *InMemCollector) monitor() {
 			// Check worker health and report aggregated status
 			i.Health.Ready(collectorHealthKey, i.isReady())
 
+			// Emit queue capacity limits so consumers can compute utilization
+			monitorConfig := i.Config.GetCollectionConfig()
+			i.Metrics.Gauge("collector_incoming_queue_capacity", float64(monitorConfig.IncomingQueueSize))
```
VinozzZ, line 349: same as above

mterhar:
And don't forget LLMs that need to see stuff becase we have empathy for robots.

tdarwin:
I'd rather send them continuously even if they don't change, because of query math.

VinozzZ:
> I'd rather send them continuously even if they don't change, because of query math.

That's fair. Then I think we should do it in the `monitor()` for both of the metrics

collect/collect.go
```diff
@@ -341,6 +345,11 @@ func (i *InMemCollector) monitor() {
...
+			i.Metrics.Gauge("collector_peer_queue_capacity", float64(monitorConfig.PeerQueueSize))
+
```
VinozzZ, line 353:
```suggestion
maxAlloc := monitorConfig.GetMaxAlloc()
i.Metrics.Gauge("memory_limit", float64(maxAlloc))
```

collect/collect.go
VinozzZ:
```suggestion
```

VinozzZ (approved):

PR #1798, opened 2026-03-23
feat: add RaaS support — ReceiveKeyIDs, AdditionalAttributes, and granular event metrics

tdarwin:
## Which problem is this PR solving?

Refinery needs new configuration capabilities to support **RaaS (Refinery as a Service)** — Honeycomb's managed Refinery sampling feature. Specifically, Refinery must be able to: (1) authorize ingest traffic by key ID rather than full key value, (2) attach deployment-identifying resource attributes to its OTLP metrics, and (3) report granular event metrics (received/dropped × traces/logs) for usage tracking.

Relates to [FRE-19](https://linear.app/honeycomb/issue/FRE-19)

### Architecture Context

```mermaid
graph TD
    Admin[Honeycomb Admin] -->|Admin UI: CRUD deployments| Hound
    User[Team User] -->|User UI: manage rules| Hound
    Beekeeper -->|"GET /2/teams/{slug}/raas/{id}/configuration"| Hound
    Beekeeper -->|"POST /2/teams/{slug}/raas/{id}/health"| Hound
    Beekeeper -->|"OpAMP: push config_yaml + rules_yaml"| Refinery
    Refinery -->|"OpAMP: health + usage reports"| Beekeeper
```

This PR adds the Refinery-side changes for RaaS. Hound generates the Refinery config YAML and distributes it via Beekeeper/OpAMP. These changes enable Refinery to consume the new config fields.

#### 1. `AccessKeys.ReceiveKeyIDs` — Key ID-based authorization

New config field `AccessKeys.ReceiveKeyIDs` (`[]string`) allows authorizing API keys by their Honeycomb key ID (not the full key value). This is more secure for RaaS because key IDs are not secret.

- **`IsAccepted(key, keyID string) error`** now checks three sources: `SendKey`, `ReceiveKeys`, and `ReceiveKeyIDs`
- **`HasKeyIDs() bool`** helper controls whether middleware performs the key ID lookup (avoids unnecessary auth API calls when not using RaaS)
- **Middleware integration** (`route/middleware.go`): Only calls `getKeyID()` when `ReceiveKeyIDs` is configured

#### 2. `OTelMetrics.AdditionalAttributes` — Deployment-identifying resource attributes

New config field `OTelMetrics.AdditionalAttributes` (`map[string]string`) adds OTLP resource attributes on every metric exported by Refinery. RaaS injects `raas_deployment_name` and `raas_deployment_id` so metrics can be attributed to specific managed deployments.

[...]

PR #1803, opened 2026-04-07
feat: add ReceiveKeyIDs config option for key ID-based authorization

tdarwin:
## Summary

- Adds `AccessKeys.ReceiveKeyIDs` config option allowing Refinery to authorize traffic by Honeycomb ingest key IDs (from `/1/auth` endpoint) instead of requiring full API keys
- Updates `IsAccepted()` to check both `ReceiveKeys` (full keys) and `ReceiveKeyIDs` (key IDs) simultaneously
- Enables live reload for both `ReceiveKeys` and `ReceiveKeyIDs` config fields
- Adds benchmarks for the `IsAccepted` hot path with various list sizes

This is part of the RaaS (Refinery as a Service) feature set, split from PR #1798.

Closes FRE-6

## Test plan

- [x] Unit tests pass for `IsAccepted()` with key IDs, full keys, and mixed scenarios
- [x] Benchmarks verify performance of key lookup with 10/100 keys
- [x] Config validation passes (`make validate` in `tools/convert/`)
- [x] `go vet ./...` passes
- [x] All existing tests continue to pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,5 @@
 # Refinery Changelog
 
-## Unreleased
-
-### Features
-
-- feat: add `ReceiveKeyIDs` config option for API key ID-based authorization
-
 ## 3.1.2 2026-03-25
 
 This release addresses security vulnerabilities CVE-2026-27139, CVE-2026-27142, and CVE-2026-25679.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Refinery Changelog
 
+## Unreleased
+
+### Features
+
+- feat: add `ReceiveKeyIDs` config option for API key ID-based authorization
+
 ## 3.1.2 2026-03-25
 
 This release addresses security vulnerabilities CVE-2026-27139, CVE-2026-27142, and CVE-2026-25679.
--- a/config/metadata/configMeta.yaml
+++ b/config/metadata/configMeta.yaml
@@ -223,20 +223,36 @@ groups:
           will be proxied through to the upstream API directly without modifying
           keys.
 
+      - name: ReceiveKeyIDs
+        type: stringarray
+        valuetype: stringarray
+        example: "your-key-id-goes-here"
+        reload: false
+        validations:
+          - type: elementType
+            arg: string
+        summary: is a set of Honeycomb Ingest Key IDs that the proxy will treat specially.
+        description: >
+          When `AcceptOnlyListedKeys` is `true`, traffic using an API key whose
+          Honeycomb ingest key ID matches an entry in this list will be accepted.
+          The key ID is the `id` field returned by the Honeycomb `/1/auth`
+          endpoint; it is distinct from the full API key value.
+
+          This allows authorization based on key IDs rather than full key values,
+          which avoids storing secret key material in the configuration file.
+          Both `ReceiveKeys` and `ReceiveKeyIDs` may be used simultaneously.
+
       - name: AcceptOnlyListedKeys
         type: bool
         valuetype: conditional
         extra: nostar APIKeys
         default: false
         reload: true
-        validation:
-          - type: requiredWith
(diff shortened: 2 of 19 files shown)


tdarwin:
One thing to note is that I've updated the configMeta.yaml to say that ReceiveKeys and ReceiveKeyIDs are eligible for LiveReload. I found in digging into these that they _do,  in fact_ reload live, and the way things are written they will continue to do so, even if it wasn't originally intended for them to do so.  I did not have to actually change any components of how the configuration is read or processed to make it so that these were live-reload eligible.  I think they were just labeled incorrectly, or some other work we did at some point made it work and we didn't update the configMeta when that happened.

config/metadata/configMeta.yaml
```diff
@@ -223,20 +223,36 @@ groups:
...
+          which avoids storing secret key material in the configuration file.
+          Both `ReceiveKeys` and `ReceiveKeyIDs` may be used simultaneously.
+
       - name: AcceptOnlyListedKeys
         type: bool
         valuetype: conditional
         extra: nostar APIKeys
         default: false
         reload: true
-        validation:
```
VinozzZ, line 232: I understand that the validation logic currently does not support the semantic of requiredWith A or B, and that's why we removed this validation logic here. However, I do think it's a good validation to have.

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Refinery Changelog
 
+## Unreleased
```
VinozzZ: The changelog entry is created as a part of the release process. Can we remove it from this PR?

config/metadata/configMeta.yaml
tdarwin: Do you think the validation should be just what it was and ignore ReceiveKeyIDs, or should it have both config options, or just move it to ReceiveKeyIDs only?

CHANGELOG.md
tdarwin: Will do!

config/metadata/configMeta.yaml
```diff
@@ -223,20 +223,36 @@ groups:
...
+        type: stringarray
+        valuetype: stringarray
+        example: "your-key-id-goes-here"
+        reload: true
+        validations:
+          - type: elementType
+            arg: string
+        summary: is a set of Honeycomb Ingest Key IDs that the proxy will treat specially.
+        description: >
+          When `AcceptOnlyListedKeys` is `true`, traffic using an API key whose
```
VinozzZ, line 237: We probably should call out that this feature does not support legacy keys.
I also start to think that the name for this config should be something like `AcceptedKeyIDs`. This is because `ReceiveKeys` supports doing replacement of an incoming key. This new feature does not support anything like that other than permitting incoming traffic. I would like to avoid users thinking that ReceiveKeys and ReceiveKeyIDs work the same.

config/metadata/configMeta.yaml
VinozzZ: There's a bug with the `requiredWith` validation code. I will file the bug. For this PR, we can do the validations rule like below to express require A or B
```
        validations:
          - type: requiredWith
            arg: ReceiveKeys
          - type: requiredWith
            arg: ReceiveKeyIDs

```

config/metadata/configMeta.yaml
```diff
@@ -223,20 +223,36 @@ groups:
           will be proxied through to the upstream API directly without modifying
           keys.
 
+      - name: ReceiveKeyIDs
+        type: stringarray
+        valuetype: stringarray
+        example: "your-key-id-goes-here"
+        reload: true
+        validations:
```
VinozzZ, line 232: This is missing a firstVersion field

tdarwin: yup, firstVersion 3.2

VinozzZ (approved):
other than the minor missing firstVersion in the configMeta.yaml file. The change looks good to me

PR #1804, opened 2026-04-07
feat: add OTelMetrics.AdditionalAttributes config option

tdarwin:
## Summary

- Adds `OTelMetrics.AdditionalAttributes` config option to inject custom resource attributes (e.g., cluster ID, environment name) into all OTLP metrics emitted by Refinery
- Both keys and values must be strings, supplied as comma-separated `key:value` pairs
- Fixes metadata for HoneycombLogger `AdditionalAttributes` (removed invalid validations, fixed example quoting, fixed indentation)

This is part of the RaaS (Refinery as a Service) feature set, split from PR #1798.

Closes FRE-75

## Test plan

- [ ] `TestOTelMetricsAdditionalAttributes` verifies resource attributes are present
- [ ] `Test_OTelMetrics_AdditionalAttributes` verifies attributes appear in exported metrics
- [ ] Config validation passes (`make validate` in `tools/convert/`)
- [ ] `go vet ./...` passes
- [ ] All existing tests continue to pass


config/metadata/configMeta.yaml
```diff
@@ -916,6 +913,22 @@ groups:
           compression costs may outweigh the benefits, in which case `none`
           may be used.
 
+      - name: AdditionalAttributes
+        type: map
+        valuetype: map
+        example: "pipeline.id:'12345',rollout.id:'67890'"
```
VinozzZ, line 944: We removed the quote at for the log attribute, but preserved it here. Is that intentional?

config/metadata/configMeta.yaml
```diff
@@ -916,6 +913,22 @@ groups:
...
+        reload: false
+        firstversion: v2.10
```
VinozzZ: The first version would be v3.1.4

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Refinery Changelog
 
+## Unreleased
```
VinozzZ: Not needed for this PR

VinozzZ (commented):
I think we should also add envvar support for this config option to match with the logger one

config/metadata/configMeta.yaml
```diff
@@ -671,19 +671,16 @@ groups:
       - name: AdditionalAttributes
         type: map
         valuetype: map
-        example: "pipeline.id:'12345',rollout.id:'67890'"
+        example: "pipeline.id:12345,rollout.id:67890"
```
tdarwin: Fix this shit, we should not have removed these quotes

config/metadata/configMeta.yaml
tdarwin: We definitely want it preserved in both places

config/metadata/configMeta.yaml
tdarwin: ☹️ 3.2 I think

VinozzZ (approved):

PR #1805, opened 2026-04-07
feat: add event metrics for opamp usage tracking

tdarwin:
## Summary

- Tracks received and dropped event counts separately from byte counts
- Reports granular metrics via OTLP usage reporting (alongside existing `bytes_received`) for usage billing
- feature set, split from PR #1798.


agent/agent.go
```diff
@@ -228,6 +228,14 @@ func (agent *Agent) healthCheck() {
 
 				agent.usageTracker.Add(signal_traces, traceUsage)
 				agent.usageTracker.Add(signal_logs, logUsage)
+
+				incomingSpan, _ := agent.metrics.Get("incoming_router_span")
```
VinozzZ: we probably should still check for existence of the metric in case the race condition between metrics registration and agent startup

collect/collect.go
```diff
@@ -470,6 +471,7 @@ func (i *InMemCollector) ProcessSpanImmediately(sp *types.Span) (processed bool,
 
 	if !keep {
 		i.Metrics.Increment("dropped_from_stress")
+		i.Metrics.Increment("events_dropped")
```
VinozzZ, line 474: we should also increment dropped in `dealWithSendTrace()`

tdarwin: ack

agent/agent.go
tdarwin: ack

agent/agent.go
```diff
@@ -229,13 +229,15 @@ func (agent *Agent) healthCheck() {
 				agent.usageTracker.Add(signal_traces, traceUsage)
 				agent.usageTracker.Add(signal_logs, logUsage)
 
-				incomingSpan, _ := agent.metrics.Get("incoming_router_span")
-				incomingNonspan, _ := agent.metrics.Get("incoming_router_nonspan_event")
-				incomingEvent, _ := agent.metrics.Get("incoming_router_event")
-				agent.usageTracker.Add(signal_events_received, incomingSpan+incomingNonspan+incomingEvent)
+				if incomingSpan, ok := agent.metrics.Get("incoming_router_span"); ok {
+					incomingNonspan, _ := agent.metrics.Get("incoming_router_nonspan_event")
```
VinozzZ: Claude, it's ok if you write a few more lines of code and just check existence for each of the metrics 😆

tdarwin: 🙄

VinozzZ (approved):

Issue #1802, opened 2026-04-07
AdditionalErrorFields Not Reported in Refinery 3 errors


tdarwin:
It seems like in version 3 of Refinery, we're no longer adding additional error fields to events that are dropped because of the 1MB event limit.  This lack of context makes it really hard to track down what service is generating these overlarge events.

**Versions**

- Go: 1.25?
- Refinery: 3.X

**Steps to reproduce**

1. Configure `Debugging.AdditionalErrorFields` to add service.name or some other common field.
2. Send events over 1MB in size to Refinery
3. Look at error log messages which won't your error fields

**Additional context**

There are probably other places where Additional Error Fields aren't being unpacked when things are dropped early in the ingest process, though I don't know which cases those might be.

PR #1807, opened 2026-04-07
fix: include AdditionalErrorFields in logs for transmission code

VinozzZ:

#1802 

- Make sure that all error logs for sending events are populated with the default fields `dataset`, `apihost`, and `environment`.
- Make sure all error logs for sending events has the AdditionalErrorFields added.

### DIFF
--- a/config_complete.yaml
+++ b/config_complete.yaml
@@ -125,7 +125,7 @@ Network:
     ##
     ## default: https://api.honeycomb.io
     ## Eligible for live reload.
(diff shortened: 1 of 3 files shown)


config_complete.yaml
TylerHelmuth, line 1: Do we need these changes?

VinozzZ: I committed too fast ..... I have deleted my key

TylerHelmuth (approved):

PR #1820, opened 2026-05-13
fix: validator exits non-zero on YAML parse errors in rules

### LINKED ISSUE #1819: Validator doesn't exit non-zero when rules YAML has syntax error
**Versions**

- Go:
- Refinery: 3.2.1

**Steps to reproduce**

Have this in your rules.yaml:
```yaml
        - Name: Frontend Document Load Instrumentation
          Conditions:
            - Field: library.name
              Operator: =
              Value: @opentelemetry/instrumentation-document-load # <-- this is invalid as per yamllint due to the leading '@'
          Sampler:
            EMADynamicSampler:
              GoalSampleRate: 20 # 1 in 20
              AdjustmentInterval: 15m
              FieldList:
                - browser.name
```

1. `refinery --validate --config config.yaml --rules_config rules.yaml`
2. `echo $?`

exit code shouldn't be `0` given the output is:

```
% ./refinery --validate --config config.yaml --rules_config rules.yaml 
loadConfigsInto unable to load config rules.yaml: yaml: line 213: found character that cannot start any token
Config and Rules validated successfully.
```

**Additional context**
We use refinery validate in CI to guard against invalid config; this config shipped and failed to apply in production.

VinozzZ:
## Summary

Fixes #1819. `refinery --validate` was exiting 0 even when `rules.yaml` had a YAML syntax error, causing broken configs to pass CI and ship to production.

**Fix:** Replace the type assertion with a `c == nil` check. `NewConfig` already encodes severity in its return: nil config = fatal error, non-nil config = warnings only. Checking `c == nil` catches all fatal errors regardless of their type.

### DIFF
--- a/cmd/refinery/main.go
+++ b/cmd/refinery/main.go
@@ -99,11 +99,9 @@ func main() {
 
 	c, err := config.NewConfig(opts, version)
 	if err != nil {
-		if configErr, isConfigErr := err.(*config.FileConfigError); isConfigErr && configErr.HasErrors() {
-			fmt.Printf("%+v\n", err)
+		fmt.Printf("%+v\n", err)
+		if c == nil {
 			os.Exit(1)
-		} else {
-			fmt.Printf("%+v\n", err)
 		}
 	}
 	if opts.Validate {
--- a/config/file_config.go
+++ b/config/file_config.go
@@ -608,6 +608,13 @@ func writeYAMLToFile(data any, filename string) error {
 // nil, it uses the command line arguments.
 // It also dumps the config and rules to the given files, if specified, which
 // will cause the program to exit.
+//
+// Return values follow an intentional two-level contract:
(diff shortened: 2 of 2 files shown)


TylerHelmuth (approved):

cmd/refinery/main.go
```diff
@@ -99,11 +99,9 @@ func main() {
 
 	c, err := config.NewConfig(opts, version)
```
TylerHelmuth, line 100: Can you update the godoc string for NewConfig to say that non-nil config with non-nil error is a warning. It is pretty un-go-like to use a returned value if error is non-nil, but its fine for this case

VinozzZ: Done — added a three-case contract block to the godoc: (nil, err) = fatal, (cfg, err) = warning/proceed, (cfg, nil) = success.

PR #1824, opened 2026-05-19
feat: override dataset name with service.name

maplebed:
To cause the same behavior we have for OTLP traffic to be applied to Honeycomb protocol traffic, override the dataset name found in the URL with the value found in the event's `service.name` field (if one exists). Only use the URL's dataset name if no field exists within the event.

### DIFF
--- a/route/route.go
+++ b/route/route.go
@@ -531,10 +531,11 @@ func (r *Router) batch(w http.ResponseWriter, req *http.Request) {
 	}
 	defer recycleHTTPBodyBuffer(bodyBuffer)
 
-	dataset, err := getDatasetFromRequest(req)
+	datasetFromRequest, err := getDatasetFromRequest(req)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


route/route.go
```diff
@@ -566,6 +567,11 @@ func (r *Router) batch(w http.ResponseWriter, req *http.Request) {
 	userAgent := getUserAgentFromRequest(req)
 	batchedResponses := make([]*BatchResponse, 0, len(batchedEvents.events))
 	for _, bev := range batchedEvents.events {
+		datasetAny := bev.Data.Get("service.name")
+		dataset, ok := datasetAny.(string)
+		if !ok {
```
kentquirk: You probably also want to check for empty here

VinozzZ:
The failed test is giving me confidence that this hack will work as expected. We just need to update the test for the build to pass

VinozzZ:
Closing as we no longer need this feature

PR #1831, opened 2026-06-24
maint: PGO-optimize Honeycomb-internal refinery images

lizthegrey:

We already collect Refinery CPU profiles from production via `pgo-collector`, but never feed them back into the build. Refinery's hot path is dominated by zstd, msgpack (un)marshal, and GC — all of which [PGO](https://go.dev/doc/pgo) helps, typically a few percent across the board, for free and with no behavior change.

We don't want this on the **public** image: the profile reflects Honeycomb's own libhoney-heavy prod→dogfood traffic and may not suit OTLP-dominant deployments. So this is scoped to the images we build and run ourselves.

- **`build-docker.sh`**: opt-in `-pgo` — when `PGO_PROFILE` points at a pprof CPU profile, it's folded into `GOFLAGS`. Unset (the default) builds exactly as before.
- **`.circleci/config.yml`**: only the `publish_docker_to_private_ecr` job (our internal ECR) sets `PGO_PROFILE`. The release job (`publish_docker_to_all_public_registries`) leaves it unset, so the published customer image stays profile-neutral.
- **`cmd/refinery/honeycomb.pgo`**: the checked-in profile, merged from 24h of production Refinery. Deliberately **not** named `default.pgo`, so `go build` never applies it implicitly — only explicit opt-in via `-pgo` uses it.
- **`CONTRIBUTING.md`**: documents the opt-in and the caveat that the profile reflects our traffic mix.

Verified `go build -pgo=cmd/refinery/honeycomb.pgo ./cmd/refinery` builds clean. Pushing this branch produces a PGO'd image in the private ECR (the branch-filtered `publish_docker_to_private_ecr` job), so it's directly testable before any infra change. The infra-side image override to consume it is a separate, SRE-owned change.


lizthegrey:
minimal impact. closing to avoid complexity

PR #1836, opened 2026-07-21
maint: upgrade to go1.26 (bonus: better CPU performance)

lizthegrey:

- Refinery was still pinned to go1.25; go1.26 brings performance improvements we want to pick up.

- Bump `go.mod` to go1.26.0
- Bump CircleCI `cimg/go` build images from 1.25 to 1.26

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -66,7 +66,7 @@ commands:
 jobs:
   test:
     docker:
-      - image: cimg/go:1.25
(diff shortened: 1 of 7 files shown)


VinozzZ (approved):

robbkidd:
I'm going to categorize this change as an enhancement (better performance) even though the mechanism is maintenance (update build tooling).

PR #1840, opened 2026-07-25
feat: Detect and log Honeycomb API key region mismatches

kushalShukla-web:

- This PR addresses an issue where Refinery silently drops telemetry when the configured Honeycomb API key does not match the configured ingest region (for example, using an `HCAIK`-prefixed API key with `api.eu.honeycomb.io`). Since no logs are emitted, the misconfiguration is difficult to identify and debug.

### DIFF
--- a/rules.yaml
+++ b/rules.yaml
@@ -1,6 +1,13 @@
-# Automatically generated on 2023-06-18T14:12:45-04:00
 RulesVersion: 2
 Samplers:
     __default__:
         DeterministicSampler:
             SampleRate: 1
+    production:
+        DynamicSampler:
+            SampleRate: 2
+            ClearFrequency: 30s
+            FieldList:
+                - request.method
+                - http.target
+                - response.status_code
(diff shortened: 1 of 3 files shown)


kushalShukla-web:
Hi @mterhar , could you take a look at this PR when you have some time?

lizthegrey:
Your pr description does not accurately reflect the content of your change. You are, it appears, trying to record when API keys are rejected. Please document what your actual intention is with this feature.

kushalShukla-web:
My bad, @lizthegrey . I didn't catch that. I've updated the content to align with the issue.

VinozzZ:
Hi @kushalShukla-web, thanks for submitting the PR. Based on your PR description, the issue described is Refinery doesn't produce an error log if a API key is misconfigured against the wrong destination region. However, Refinery do emit an error log when a 401 Unauthorized error is returned. The code change in this PR is actually trying to rate-limit the error log from 401 responses.
Can you open a github issue describe the problem you are trying to solve? It will help the team to have a better understanding of your solution

PR #1841, opened 2026-07-25
fix: pin klauspost/compress to arm64 link-register fix

lizthegrey:

- klauspost/compress's generated arm64 zstd decoder assembly places function locals 8 bytes too low, overlapping the saved link register. This doesn't corrupt decoded data, but it breaks stack unwinding on arm64 during GC stack scans, panic handling, and CPU/heap profiling, which can crash the process. Present since compress v1.19.0. Upstream fix: [klauspost/compress#1176](https://github.com/klauspost/compress/pull/1176) (not yet merged).

- Pin `github.com/klauspost/compress` to the fix commit via a `replace` directive pointing at the `honeycombio/compress` fork, until the upstream PR merges and cuts a release.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Pah4t3q1rYVNDZJqqrzWWs)_

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -120,3 +120,7 @@ require (
 )
 
 tool github.com/google/go-licenses/v2
+
+// fix for arm64 zstd decoder asm overwriting the saved link register, which breaks
+// stack unwinding during GC scans and profiling: klauspost/compress#1176 (not yet merged upstream)
+replace github.com/klauspost/compress => github.com/honeycombio/compress v1.19.1-0.20260725070242-26c7cb34104b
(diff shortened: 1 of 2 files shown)


robbkidd (changes requested):
Out of band of this PR, a few of us decided to give upstream a week or so to review, merge, and release the proposed fix before we opt to release with a forked dependency. We'll reconsider on 2026-07-31.

quinnleong (commented):
I think Yingrong is out of office – I'll defer to Robb on this one, because I don't have any context on Refinery yet.

lizthegrey:
Ok to cut release without this fix. determined it was impossible to reach the situation of the link register being read by the go runtime, while the link register saved value was corrupt.

PR #1850, opened 2026-08-18
feat: support Classic environments that have been migrated in-place to E&S

robbkidd:

In Honeycomb, a Classic environment can be migrated in place so that it behaves like Environments & Services. The migration process gives the Classic environment a name but the keys for the environment remain Classic keys.

Currently in Refinery, OTLP requests with a Classic key:

- are being rejected if they lack an `x-honeycomb-dataset` header
  - E&S ignores a dataset header, so it shouldn't be required.
- have their destination dataset determined by that header when present
  - E&S places telemetry into datasets by `service.name`.
- select the sampler by dataset name
  - E&S samplers are keyed by environment name.

Refinery needs to detect a migrated environment (Classic key, non-blank environment name) and change behavior accordingly.

- Add a new config option `General.EnableMigratedClassicAsEnvironment` (bool, default `false`).
  - While it is off, every path behaves as it does today.
  - ⚠️ It is dangerous to enable this. The option is being [withheld from the generated configuration docs and sample config files](https://github.com/honeycombio/refinery/commit/570c95d720577da098ceb6d3bb3ced5602b7f135) because what an operator prepares before enabling it is not documented yet. Documentation on the in-place migration of Classic environments is written separately, shaped by the experiences of the bold few who go looking for this option now that it exists.
- Look up the environment for a classic key when the option is on.
  - `getEnvironmentName` returned blank immediately for any classic key; now it does the same `/1/auth` lookup an E&S key does.
- Set `ri.EnvironmentName` in the OTLP ingest handlers.
  - This meets the conditions for `ValidateDatasetHeader` to skip requiring the header.
  - Respects `SendKeyMode` so that the lookup uses the key the telemetry will be forwarded with rather than the incoming one.
  - The log handlers set the field for consistency, even though logs have never required the dataset header and have always preferred `service.name` over dataset header.
- Clear `ri.Dataset` in the trace handlers once a request has an environment name, so the destination dataset comes from the `service.name` resource attribute instead of the header.
- For a migrated Classic environment, return the environment name from `DetermineSamplerKey`, so the sampler is selected by environment.
- Name the sampler selector for what it is in the `Sending trace` debug log, [for every kind of key](https://github.com/honeycombio/refinery/commit/c8eb1c1dd84964c0953dfefdb3545e648a2154d8).
  - The field was named `dataset` or `environment` by re-deriving the kind from the API key. A classic key sending to a migrated environment breaks that assumption and reports the environment name as a dataset that does not exist.
  - The field is now `samplerSelector` for all traffic, so anything reading `dataset` or `environment` out of this line will no longer find them.

## New Tests

- This feature's behavior table was [written as a test suite before the implementation](https://github.com/honeycombio/refinery/pull/1850/commits/ee8fd350e90c20083aac5e6c9290df1079b148cc), and its failing cases drove each change. [Those cases were moved](https://github.com/honeycombio/refinery/pull/1850/commits/197db4de7a89b102fa43d6bff5137887f52005c8) into `TestOTLPHandler` and `TestLogsOTLPHandler`, beside the other cases for the same handler.
- `TestSamplerKeySelection` in `collect/` covers which key a trace's sampler is selected by. `DetermineSamplerKey` had its own tests, but we lacked an integration test to exercise it through a caller that supplies a trace.
- `TestSendTraceLogging` in `collect/` covers what the `Sending trace` debug log reports, for a classic key into an unmigrated environment, a classic key into a migrated one, and an E&S key.

## TODO

- [x] disperse the BDD test cases into signal-specific tests after review
- [x] local end-to-end test with a Refinery build sending to Honeycomb
  - This is how the misnamed log field was found. The commits since that run change a log field name and config metadata, neither of which affects routing or sampler selection.


route/otlp_logs.go
```diff
@@ -92,6 +101,15 @@ func (l *LogsServer) Export(ctx context.Context, req *collectorlogs.ExportLogsSe
 	}
 	keyToUse, _ := apicfg.GetReplaceKey(ri.ApiKey, keyID)
 
+	// Look up the environment with the key that will actually be sent, so a
+	// replaced key resolves its own environment. A failed lookup leaves the
+	// name blank, which validates as an unmigrated Classic environment.
+	envName, envErr := l.router.getEnvironmentName(keyToUse)
```
MikeGoldsmith: This looks looks to be duplicated twice each for logs and traces, can we make it generic and move to a func that is shared intsead?

MikeGoldsmith (approved):
Looks good to me. Left one suggestion.

route/otlp_logs.go
robbkidd: Yep! [Refactored into a shared function on the router](https://github.com/honeycombio/refinery/pull/1850/commits/4be313a3e8c2a217f47db11e9f9b950c72624107). Also [added a set of semi-integration tests through the refactored code path for E&S traces](https://github.com/honeycombio/refinery/pull/1850/commits/6f067a60d73f7366ff7e165d620510c71580b081). The interactions between what husky knows and what Refinery knows are tricky, so I decided to cover them to catch future regressions.

MikeGoldsmith (approved):

config/metadata/configMeta.yaml
```diff
@@ -78,6 +78,36 @@ groups:
...
+          `x-honeycomb-dataset` header. The destination dataset comes
+          from the `service.name` resource attribute, and the sampler is
+          selected by environment name instead of dataset name.
+
+          This only applies to Classic environments that have been
+          migrated to have an environment name. An unmigrated Classic
+          environment, and any Environments & Services key, behaves the
+          same as when this setting is disabled.
+
+          ⚠️ Before enabling this setting, make sure every migrated Classic
```
tdarwin: > every migrated Classic

This is weird wording in this paragraph.  The only way one Refinery would be handling multiple migrated classic environments would be if there were multiple teams' traffic running through one Refinery. I suppose it's _possible_, but it's exceedingly rare. I think it's fine to phrase this section around the idea that there is only one classic environment per team, and the expectation is that generally one Refinery cluster is associated with a single team.  Not a blocker, but it feels weird to me.

tdarwin (approved):
The code looks fine overall to me, the wording on the extra footgun notice feels like it could be better.  Claude really likes to assume that teams have multiple Classic environments (fought this a lot) and it's just not the case...

config/metadata/configMeta.yaml
```diff
@@ -78,6 +78,36 @@ groups:
           Kubernetes ConfigMaps) may cause delays in propagating configuration
           changes.
 
+      - name: EnableMigratedClassicAsEnvironment
+        type: bool
+        valuetype: nondefault
+        default: false
+        reload: false
```
VinozzZ, line 85: Can we add a firstversion here to document the refinery version this config is introduced in?

config/metadata/configMeta.yaml
```diff
@@ -78,6 +78,36 @@ groups:
           Kubernetes ConfigMaps) may cause delays in propagating configuration
           changes.
 
+      - name: EnableMigratedClassicAsEnvironment
+        type: bool
+        valuetype: nondefault
+        default: false
```
VinozzZ, line 84: We may want to add a validation rule to say that DatasetPrefix and EnableClassicAsEnvironment can't be set at the same time. If someone is relying on DatasetPrefix in their rules config, they need to know it no longer applies once they enable this new config

config/metadata/configMeta.yaml
```diff
@@ -78,6 +78,36 @@ groups:
...
+          from the `service.name` resource attribute, and the sampler is
+          selected by environment name instead of dataset name.
+
+          This only applies to Classic environments that have been
+          migrated to have an environment name. An unmigrated Classic
+          environment, and any Environments & Services key, behaves the
+          same as when this setting is disabled.
+
+          ⚠️ Before enabling this setting, make sure every migrated Classic
+          environment sampled by Refinery is ready. This setting applies to the
```
VinozzZ: The term "ready" here feels a bit ambiguous to me. I wonder if there's a public doc page that we can link to about when is appropriate to turn on this feature when migrating from classic to E&S. 
For example, should this be turned on after the classic environment has been updated to E&S? Should the code change (setting service.name) happen first or should the this setting be turned on first ? It may not be Refinery's job to give these instructions, but I do think a coherent step and step doc is helpful here

robbkidd:
Found during an end-to-end test: debug log output from a collect worker is currently misleading when sending telemetry to a migrate environment.

```
msg="Sending trace" dataset=header-ds    reason=rules/trace/matched-ruleset-header-ds
msg="Sending trace" ❗️dataset=fancy-env  reason=rules/trace/matched-ruleset-fancy-env ❗️ this is an environment not a dataset
msg="Sending trace" environment=ens-env  reason=rules/trace/matched-ruleset-ens-env
```

[The branch logic for labeling this field](https://github.com/honeycombio/refinery/blob/bc345102cf1ec8cfffe0b3dd027777c0113b88f1/collect/collect.go#L637-L641) `dataset` or `environment` has the now-out-of-date assumption that telemetry sent with a classic key is only destined for a named dataset.

What is being logged is the string used to select a sampler from samplers defined in rules.yaml. I propose the field name be changed to `samplerSelector` and the branching logic removed.

config/metadata/configMeta.yaml
robbkidd: Yup, good idea. I propose I add that with `v3.4.0` as the first version under the assumption that the next release will be a minor (because at a minimum the next release will include this PR which is an enhancement).

robbkidd: Added in 578569c9c38a623c9a7635c056e2f491b8946ed8.

robbkidd:
Field changed to `samplerSelector` for all the "Sending trace" logs in c8eb1c1d.

config/metadata/configMeta.yaml
robbkidd: Noting, @tdarwin and I chatted out-of-band about the docs for this option. We've opted to [mark it "unpublished" with a vague and foreboding warning](https://github.com/honeycombio/refinery/pull/1850/commits/570c95d720577da098ceb6d3bb3ced5602b7f135). We want the change in to start using it in special cases. It is from the results of those cases that further words will be written and lead to marking this option publishable with guidance.

config/metadata/configMeta.yaml
robbkidd: @tdarwin and I decided to mark this option as "unpublished" for now. I think that lets us delay a decision on adding a WILL-NOT-START `conflictsWith` validation to this option. I propose we decide that after this PR merges and before/during marking the option as "published".

config/metadata/configMeta.yaml
robbkidd: Agreed! Punting on getting the phrasing right here until later by marking this option as "unpublished".

tdarwin (approved):
Image

PR #1858, opened 2026-09-15
ci: test against Valkey as well as Redis

robbkidd:

Refinery uses very little Redis: the peer pub/sub code only PUBLISHes, SUBSCRIBEs, and optionally AUTHs.

There's been interest in using servers other than official Redis that speak the same protocol. They *ought* to work, but we didn't have anything to confirm behavior other than with *Redis* Redis. This is the smallest next step towards being able to more confidently assert that Refinery works with Valkey, too.

- Parameterize the `test` job on the server image and run it as a matrix over `redis:8` and `valkey/valkey:9`
  - Moved to latest major versions since our go-redis dependency only claims support for Redis 8+.
  - The job's matrix is given an alias so that downstream jobs can require one prerequisite and not have to keep up with any future variations.
- Update version of Redis in the smoke test, too.


VinozzZ (approved):

.circleci/config.yml
```diff
@@ -297,6 +301,12 @@ workflows:
   build:
     jobs:
       - test:
+          matrix:
+            alias: test
+            parameters:
+              redis_image:
+                - redis:6.2
```
robbkidd: More investigation: we're currently depending on go-redis v9.22.0. [That version's README claims official support only for 8.0+](https://github.com/redis/go-redis/tree/v9.22.0#supported-versions) and says 7.0+ is unsupported but expected to work.

Should we move this up to 7? To 8? Now that it's matrixed, both? [Redis 6.2 was EOL'd 4 years ago](https://endoflife.date/redis).

robbkidd: @VinozzZ and I chatted out of band and decided we'll matrix the latest versions of Redis and Valkey backends because we don't really _support_ support stuff that our dependencies don't support.

PR #1859, opened 2026-09-16
fix: include full sampler config in shared-dynsampler key

sam-spinner:
## Description

Refinery v3 shares dynsampler engines across rules based on a key built from `(environment, sampler type, goal rate, field list)`. This key omits other sampler configuration fields that affect sampling behavior, such as `MaxKeys`, `AdjustmentInterval`, `Weight`, and `BurstMultiple`.

As a result, two rules with the same sampler type, goal rate, and field list but different values for these behavioral fields can silently share a single engine. Whichever rule initializes first determines the effective configuration for both. For example, if one rule sets `MaxKeys: 1000` and another rule with the same type/rate/fields leaves `MaxKeys` unset, the second rule may cause the shared engine to use the code default of `500`, and the first rule's explicit `MaxKeys` is ignored.

This change includes every sampler configuration field that affects decisions in the shared-dynsampler key. The key is now derived from the full config struct, so two differently configured samplers can no longer merge. Samplers with identical configs continue to share the same engine, preserving the existing cross-worker consistency guarantee.

## How was this change tested?

- [x] Automated test (unit, integration, etc.)
- [ ] Manual test (provide reproducible testing steps below)

Added unit tests in `sample/sample_test.go`:
- `TestDifferentConfigsDoNotShareDynsampler/EMADynamicDifferentMaxKeys` verifies that two EMADynamic samplers with the same type/rate/fields but different `MaxKeys` do not share an engine.
- `TestDifferentConfigsDoNotShareDynsampler/EMADynamicDifferentAdjustmentInterval` verifies the same for `AdjustmentInterval`.
- `TestDifferentConfigsDoNotShareDynsampler/IdenticalConfigsStillShare` verifies that identical configs still share one engine.

Also ran the full `go test ./sample/` suite locally.

## Notes

This is a behavior change for deployments that currently rely on the under-defined sharing. After this fix, any two samplers with different config values will receive distinct engines. This matches user expectations: a config field that changes sampling behavior should not be silently dropped when another rule happens to initialize first.


robbkidd:
What use case do you have where MaxKeys, AdjustmentInterval, Weight, and BurstMultiple are set differently for the same (environment, sampler type, goal rate, field list)?

PR #1861, opened 2026-09-21
ci: add TLS and AUTH coverage to smoke tests

robbkidd:

- Some test coverage for working with Redis configured for TLS and AUTH.

Update the smoke-test to make assertions about things that it would be difficult to do in unit tests: two peers find each other through a redis that uses TLS.

- combined the existing docker-compose files into a single TLS'd config and replaced an old simple SSL image with a newer base openssl image (that requires more scripting)
- added PrometheusMetrics so that num_peers could be read directly
- mimicked our smokey BATS assertion suite from other projects to perform the setup and assertions instead of relying only on Docker's service ready check at startup
- reused assert_eventually_* vocabulary from our Go tests here in BATS for the things we have to wait to test
- ensure the BATS test output is readable by CircleCI job runs with JUnit output that is always saved


MikeGoldsmith (approved):
Great work Robb - love to see BATS back 🚀

